PR #5, opened 2019-06-15
Fix segfault when using rmw_get_node_names()

joxoby:
Sometimes dds_qget_userdata() returns true with a nullptr. It might be a good idea to revise its implementation.


eboasson:
Oh dear. Caught in my own trap as it were ...

I just added the ability to change the QoS settings in Cyclone and in the process made sure that a QoS object returned by the Cyclone will always return true for all the QoS that are applicable to the entity. The argument was mostly the unreasonableness of expecting the application to know what the default value is. (And all that started also with ``dds_qget_userdata`` returning false.)

You have a good point that perhaps the interface should be changed. Returning false when the {user,topic,group} sequence is empty is an option, but alternatively, it could return a non-null pointer to a zero byte and a size of 0. That way the claim that it always returns something that can safely be treated as a nul-terminated string is true.

Issue #11, opened 2019-06-17
Windows 10 Compilation Fails


allenh1:
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - source
- Version or commit hash:
  - master
- DDS implementation:
  - rmw_cyclonedds
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

Try to build on Windows 10 with Cyclone DDS installed.

#### Expected behavior

Successful compilation and the ability to be used in runtime.

#### Actual behavior

Compilation fails

---

Note: I'm working on some patches for you, I'll file them tonight.

One issue is with `__attribute__ ((unused))` which isn't supported on Windows 10. Since the `CMakeLists.txt` file specifies C++14, I've changed these instances to `static_cast<void>(var)`. In C++17, one could use `[[maybe unused]]`, but I'm only noting this for completeness.

I'll see if I can figure out what else is blocking the windows compilation as well, but that might take me a bit longer ;).

Glad to see a new middleware appear! Pretty exciting stuff!

eboasson:
Hi @allenh1 , I am a bit shocked to discover that I forgot to mention I had never even tried building it on Windows on account of not having a Windows box available. Surely I should have warned everyone ... (It was developed on macOS and checked on Ubuntu 18.04.)

Anyway, I am grateful that you're looking into it and have started fixing things. I browsed the ROS forums a bit, and it appears that the target really should be C++14. Casting to void would then indeed be the appropriate solution.

I'm curious to see what else you ran/run into ... and if there's anything I can do to help fix it ...

dennis-adlink:
This issue will be solved by PR [https://github.com/eclipse-cyclonedds/cyclonedds/pull/248](url) on CycloneDDS. With these fixes included I can successfully run the ROS2 examples using rmw_cyclonedds.

allenh1:
> This issue will be solved by PR https://github.com/eclipse-cyclonedds/cyclonedds/pull/248 on CycloneDDS

That's great! There seemed to be some more core adjustments to be made, glad this is all worked out.

Looking forward to trying out the new middleware.

joespeed:
@allenh1 @dennis-adlink @eboasson so this is fixed? thanks!

eboasson:
Based on @dennis-adlink's comment and my merging his PR on Cyclone, I'm presuming this is fixed (I don't have Windows to try it out myself).

allenh1:
[...]

PR #12, opened 2019-06-18
Improve Windows 10 Support

allenh1:
Replace `__attribute__ ((unused))` with `static_cast<void>`.

Signed-off-by: Hunter L. Allen <hunterlallen@protonmail.com>

---

Connects to #11 -- though this isn't a full fix.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -149,7 +149,10 @@ static Cdds gcdds;
 
 static void clean_waitset_caches ();
 
+#ifndef WIN32
+/* TODO(allenh1): check for Clang */
 #pragma GCC visibility push (default)
+#endif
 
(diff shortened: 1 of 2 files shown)


eboasson:
Thanks, @allenh1!

Regarding the "/* TODO(allenh1): check for Clang */" comment on
```C
#pragma GCC visibility push (default)
```
Clang is generally pretty compatible with GCC and it indeed builds fine using Clang on macOS. While I do think that that #pragma is fine, I haven't checked the manuals and so my observation should be considered no more than anecdata at this point. Therefore, I'll let the TODO stand.

allenh1:
> Clang is generally pretty compatible with GCC and it indeed builds fine using Clang on macOS. While I do think that that #pragma is fine, I haven't checked the manuals and so my observation should be considered no more than anecdata at this point. Therefore, I'll let the TODO stand.

My thoughts exactly ;) I wasn't sure, so I just made a note is all. We can definitely remove it. If there's a more Clang-y way, we might consider that in place, though.

PR #30, opened 2019-09-07
changed path as suggested

joespeed:
changed path per https://github.com/ros2/ros2_documentation/pull/316#discussion_r315392636

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,8 @@ implementation.
 All it takes to get Cyclone DDS support into ROS2 is to clone this repository and the Cyclone DDS
 one in the ROS2 workspace source directory, and then run colcon build in the usual manner:
 
-    cd ros2_ws/src
+    mkdir ros2_ws/src/atolab/rmw_cyclonedds
+    cd ros2_ws/src/atolab/rmw_cyclonedds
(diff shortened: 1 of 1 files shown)


eboasson:
Thanks @joespeed for looking after the README. I have no problem with recommending that this repository be cloned in ros2_ws/src/atolab, but to have another layer of rmw_cyclonedds underneath that seems a bit superfluous.

Secondly, it seems to me the point of adding the "atolab" part is to keep the packages organised by (GitHub) organisation. With that in mind, one would have to suggest cloning Cyclone into ros2_ws/src/eclipse. (You could argue that it should be "eclipse-cyclonedds" — it doesn't matter much to me.)

I.e., I think it would make more sense to do:

    mkdir -p ros2_ws/src/{atolab,eclipse}
    cd ros2_ws/src
    git clone https://github.com/atolab/rmw_cyclonedds atolab/rmw_cyclonedds
    git clone https://github.com/eclipse-cyclonedds/cyclonedds eclipse/cyclonedds
    cd ..

eboasson:
Thanks @joespeed but not that it has been decided to move the repository to ros2.org this pull request has become irrelevant.

PR #31, opened 2019-09-11
Implemented byte-swapping in deserializer

dennis-adlink:
Set the correct endianness flag when serializing ROS messages to CDR
and added support for byte-swapping in CDR deserializer when in a
mixed-endian network.

Signed-off-by: Dennis Potman <dennis.potman@adlinktech.com>

### DIFF
--- a/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdes.hpp
+++ b/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdes.hpp
@@ -24,6 +24,8 @@
 #include <vector>
 #include <type_traits>
 
+#include "dds/ddsi/q_bswap.h"
+
 class cycser
 {
 public:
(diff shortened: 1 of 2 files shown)


eboasson:
Thank you @dennis-adlink, but if I may suggest a small change:
```C++
  inline void deserialize(float & x) {
    deserialize(*reinterpret_cast<uint32_t *>(&x));
  }
```
works just as well as the much more involved loops you used for floating-point data (there are analogous variants for doubles and arrays of the same). You're really just swapping bytes and the interpretation of those bytes really is immaterial in this. This should compile down to a single instruction, at least on x86 (which has a ``bswap`` instruction) and provided the compiler manages to recognise the bit-fiddling in the ``bswap4u`` functions and the like.

Secondly, but that's more something that requires an apology than something that I think you could've done better: I just merged some pull requests this morning and now there is some code to deserialise the data into a string: pretty-printing of the data for the trace.

Obviously your changes do not cover that addition as they antedate the merging of that pull request. Would you mind adding the exact same byte-swapping to the print routines (in the same file)?

Thirdly, I'm not happy with relying on the header files in Cyclone that still have this old, weird, ``q_`` prefix. If you look in the history of changes, you can see that I have sorted out the prefixing and renaming of functions before using them from the RMW layer. Naturally, all of those in Cyclone should have been renamed a long time ago, but, well, that didn't happen yet.

With that in mind, it is probably more attractive to just define the three trivial byteswap functions in this file rather than import them now and ending up with this dependency on a file that should be renamed. A pull request on Cyclone that cleans up the naming and using the updated names in this on, is of course also welcome, but definitely more work.

PR #32, opened 2019-09-12
Use rosdep

rotu:
Recommend resolving and installing dependencies with rosdep instead of apt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ source directory, and then run colcon build in the usual manner:
     cd ros2_ws/src
     git clone https://github.com/atolab/rmw_cyclonedds
     cd ..
-    colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCYCLONEDDS_FROM_SOURCE=ON
+    colcon build --cmake-args -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCYCLONEDDS_FROM_SOURCE=ON
     export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp
 
 If you prefer to install CycloneDDS yourself, set `-DCYCLONEDDS_FROM_SOURCE=OFF`. CMake caches this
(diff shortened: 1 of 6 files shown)


dirk-thomas:
I don't think there is anything in this repo which needs / uses either Maven or Java. It is the `CycloneDDS` package which needs these depenencies and since that package doesn't have a manifest the information is added in the release repo: see https://github.com/ros2-gbp/cyclonedds-release/blob/12ca7b0b56f1188d9cb71194e1c95f1438dfe99a/dashing/package.xml

rotu:
You’re right, there isn’t. This package is just the best place that I could see to put these dependencies so rosdep will get them when building cyclonedds and rmw_cyclonedds from source.

Where else would you recommend putting these dependencies instead if the user is building cyclonedds?

rotu:
I *think* the right thing to do is create a cyclonedds_vendor package, which wraps cyclone_dds and adds the package.xml file but not sure.

eboasson:
> I don't think there is anything in this repo which needs / uses either Maven or Java. It is the [CycloneDDS](https://github.com/eclipse-cyclonedds/cyclonedds.git) package which needs these depenencies and since that package doesn't have a manifest the information is added in the release repo: see https://github.com/ros2-gbp/cyclonedds-release/blob/12ca7b0b56f1188d9cb71194e1c95f1438dfe99a/dashing/package.xml

Cyclone can be built without the Java parts by setting CMAKE variables to skip the IDL preprocessor and a configuration tool: ``BUILD_IDLC=NO`` and ``BUILD_CONFTOOL=NO``. Some examples and tools that depend on IDL preprocessing also don't get build, but you do get the core library, and that's enough for the RMW layer.

I don't know if that would be feasible, though.

eboasson:
Thank you @rotu, this is great!

rotu:
Shoot. I think there’s a bug with how this interacts with Colcon since the default is to build all packages in a workspace. I probably need to turn off building CycloneDDS *within the vendor package itself*. Follow up PR coming

dirk-thomas:
You have to choose which of the two options you want to do:
* clone `cyclonedds` into the workspace and build it as any other package or
* use a vendor package which by default builds CycloneDDS as an ExternalProject within

dirk-thomas:
@rotu FYI this change broke CI since the package doesn't declare a dependency on `git`: http://build.ros2.org/view/Dci/job/Dci__nightly-cyclonedds_ubuntu_bionic_amd64/32/consoleFull#console-section-209

Issue #7, opened 2019-06-16
Deserializer doesn't validate its input


eboasson:
The custom serializer/deserializer that converts between ROS message format and CDR (a derivative of the dynamic FastRTPS RMW layer) doesn't do any form of checking or byteswapping of its input in a mixed-endian network.

PR #36, opened 2019-09-17
Validation in deserializer

dennis-adlink:
This PR solves issue #7 by adding validation in CDR deserialization: max buffer length is checked
when deserializing fields and strings are checked for null-terminator (except for wstrings, which are serialized without null-terminator).


eboasson:
Thank you @dennis-adlink!

Still, there are a few things I'd like you to give a bit more thought to:

* While there is nothing wrong with using "Exception" and "DeserializationException" as file/class names, there are a bit out of keeping with most of the names that have been recently introduced in the Cyclone DDS-specific stuff (there's been a gradual change from one format to another, mostly because of inheriting some very useful things from the FastRTPS RMW layer).
* Checking whether enough input data is available should always be done in a way that protects against overflow. So instead of:
  ```C
  inline void validate_size(size_t sz)
  {
    if (pos + sz > lim) {
      throw DeserializationException("invalid data size");
    }
  }
  ```
  one should use something like ``sz > lim - pos`` instead. Of course that in turn requires guaranteeing that ``pos <= lim`` at all times. The reason is that otherwise, if one expects, e.g., a string and I craft some data with a very large number for the sequence length, the check will pass even though the contents are invalid. (Using ``size_t`` on a 64-bit machine, you're actually pretty safe considering that all lengths in CDR are 32-bits ... but there are also 32-bit machines out there.)
* An analogous problem exists with ``validate_size(sz * sizeof(T))``.
* It furthermore makes sense to verify that any sequence size in the input is no larger than the size of the input itself: that protects against doing huge memory allocations.
* If an exception is thrown by the serialiser (the case got added to the code, even though today there won't be such an exception), then I think (never entirely sure about C++ rules ...) it will leak the ``struct serdata_rmw`` that it allocates. I don't know what to do with memory allocated because of a partial deserialization, but I suspect that'd be okay-ish, because I think the message will either be reused or freed higher up the call stack.

Testing it thoroughly is a bit of a problem, but a quick experiment shows that a bogus message is returned when the input is invalid. That is not a flaw in the changes you did, but rather one in Cyclone's ``take`` operation, not bothering to check whether conversion to a sample succeeded is pretty bad if you allow it to fail ...

The default implementation doesn't suffer from this problem because it validates the input prior to inserting it in the reader history cache. The type information to do so is available here, too, but I'd say that the Cyclone should deal more gracefully with this situation, for example by marking the sample "invalid", or simply by deleting it altogether on first failed deserialisation.

So that detail is not something that blocks this PR.

PR #38, opened 2019-09-19
update repo URL with changed org unit

dirk-thomas:


dirk-thomas:
> I presume you have commit rights, and therefore I only need to approve?

Correct. Usually we let the author of the PR click the merge button since that person knows best when a good time is (sometimes in coordination with other PRs).

Btw you have admin rights to the moved repo. Please feel free to grant other collaborators access to the repo as you see fit.

PR #39, opened 2019-09-23
Code improvements in ser/deser code wrt passing data size

dennis-adlink:
Some code improvements in ser/deser code wrt passing data size to cycdeserbase class.

Signed-off-by: Dennis Potman <dennis.potman@adlinktech.com>


eboasson:
Thanks!

PR #41, opened 2019-09-23
Remove dead string serialization code

### LINKED ISSUE #40: memory leak? Does this malloc ever get freed?
https://github.com/ros2/rmw_cyclonedds/blob/0e6fd30a8c4ccb1743f93ca1beee7cefe3b25ce9/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdes.hpp#L295

There’s a lot of other questionable code in here, including much use of macros where templates would be more appropriate and repeated code.

rotu:
fix #40

### DIFF
--- a/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdes.hpp
+++ b/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdes.hpp
@@ -242,7 +242,6 @@ public:
   inline cycdeser & operator>>(uint64_t & x) {deserialize(x); return *this;}
   inline cycdeser & operator>>(float & x) {deserialize(x); return *this;}
   inline cycdeser & operator>>(double & x) {deserialize(x); return *this;}
-  inline cycdeser & operator>>(char * & x) {deserialize(x); return *this;}
   inline cycdeser & operator>>(std::string & x) {deserialize(x); return *this;}
   inline cycdeser & operator>>(std::wstring & x) {deserialize(x); return *this;}
   template<class T>
@@ -287,14 +286,6 @@ public:
     deserialize(sz);
     validate_size(sz, el_sz);
     return sz;
-  }
-  inline void deserialize(char * & x)
-  {
-    const uint32_t sz = deserialize_len(sizeof(char));
-    validate_str(sz);
-    x = reinterpret_cast<char *>(malloc(sz));
-    memcpy(x, data + pos, sz);
-    pos += sz;
   }
   inline void deserialize(std::string & x)
   {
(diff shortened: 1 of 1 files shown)


eboasson:
@rotu looks good to me, the only thing is that ``cycser`` and ``cycprint`` have the same unused function. I'm not sure how to give you a patch to add to this PR other than by just copy-pasting it here.

```patch
diff --git a/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdes.hpp b/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdes.hpp
index 77f7d29..4cdb678 100755
--- a/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdes.hpp
+++ b/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdes.hpp
@@ -47,7 +47,6 @@ public:
   inline cycser & operator<<(uint64_t x) {serialize(x); return *this;}
   inline cycser & operator<<(float x) {serialize(x); return *this;}
   inline cycser & operator<<(double x) {serialize(x); return *this;}
-  inline cycser & operator<<(const char * x) {serialize(x); return *this;}
   inline cycser & operator<<(const std::string & x) {serialize(x); return *this;}
   inline cycser & operator<<(const std::wstring & x) {serialize(x); return *this;}
   template<class T>
@@ -80,14 +79,6 @@ public:
   {
     serialize(static_cast<unsigned char>(x));
   }
-  inline void serialize(const char * x)
-  {
-    size_t sz = strlen(x) + 1;
-    serialize(static_cast<uint32_t>(sz));
-    resize(off + sz);
-    memcpy(data() + off, x, sz);
-    off += sz;
-  }
   inline void serialize(const std::string & x)
   {
     size_t sz = x.size() + 1;
@@ -395,7 +386,6 @@ public:
   inline cycprint & operator>>(uint64_t & x) {print(x); return *this;}
   inline cycprint & operator>>(float & x) {print(x); return *this;}
   inline cycprint & operator>>(double & x) {print(x); return *this;}
-  inline cycprint & operator>>(char * & x) {print(x); return *this;}
   inline cycprint & operator>>(std::string & x) {print(x); return *this;}
   inline cycprint & operator>>(std::wstring & x) {print(x); return *this;}
   template<class T>
@@ -461,15 +451,6 @@ public:
     validate_size(sz, el_sz);
     return sz;
   }
-  inline void print(char * & x)
-  {
-    const uint32_t sz = get_len(sizeof(char));
-    validate_str(sz);
-    const int len = (sz == 0) ? 0 : (sz > INT32_MAX) ? INT32_MAX : static_cast<int>(sz - 1);
-    static_cast<void>(x);
-    prtf(&buf, &bufsize, "\"%*.*s\"", len, len, static_cast<const char *>(data + pos));
-    pos += sz;
-  }
   inline void print(std::string & x)
   {
     const uint32_t sz = get_len(sizeof(char));
```

rotu:
Thanks! good call

eboasson:
Thanks @rotu!

PR #42, opened 2019-09-23
Rework serialization

rotu:
Make serialization better!
1. Better separation between the typesupport and serialization layers
2. All the non-compliant CDR1 serialization is now clearly delimited
3. No resizing buffers/memory allocation in the serialization layer


rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdes.hpp
```diff
@@ -35,122 +35,62 @@ class cycser
...
   }
+
   inline void serialize(const std::wstring & x)
   {
-    size_t sz = x.size();
-    serialize(static_cast<uint32_t>(sz));
-    resize(off + sz * sizeof(wchar_t));
-    memcpy(data() + off, reinterpret_cast<const char *>(x.c_str()), sz * sizeof(wchar_t));
-    off += sz * sizeof(wchar_t);
+      serialize_list(x.c_str(), x.size()); // todo: is the size right or do we need to include the null terminator?
```
rotu: @eboasson What's the intent here?
(1) should the string be both prefixed by a count *and* null-terminated?
(2) should the serialized wstring count include the null character?

eboasson: The CDR spec is very clear about ordinary strings: it must be null-terminated and the count must include the terminator (so the count for an empty string must be 1). For wide strings I simply matched what Fast-RTPS did. (Neither OpenSplice nor Cyclone supports wide strings natively yet, thankfully! But it also means I only have the specs and Fast-RTPS ...)

The specs are a bit of a mess with wide strings ([CORBA interoperability](https://www.omg.org/spec/CORBA/ISO/19500-2/PDF), section 9.4.2.7 is one; [XTypes](https://www.omg.org/spec/DDS-XTypes/1.3/Beta1/PDF) is another) and with the two together I think there is only a single interpretation:

Section 9.4.2.7 on wide strings:

> For GIOP version 1.1, a wide string is encoded as an unsigned long indicating the length of the string in octets or unsigned integers (determined by the transfer syntax for wchar) followed by the individual wide characters. The string contents include a single terminating null character. The string length includes the null character. The terminating null character for a wstring is also a wide character.
> For GIOP version 1.2 and 1.3, when encoding a wstring, always encode the length as the total number of octets used by the encoded value, regardless of whether the encoding is byte-oriented or not. For GIOP version 1.2 and 1.3 a wstring is not terminated by a null character. In particular, in GIOP version 1.2 and 1.3 a length of 0 is legal for wstring.
> NOTE: For GIOP versions prior to 1.2 and 1.3, interoperability for wstring is limited to the use of two-octet fixed-length encoding.
> Wstring values in encapsulations are assumed to be encoded using GIOP version 1.2 and 1.3 CDR.

GIOP versions go up to at least 1.4, which I guess means one should follow GIOP 1.3? And "encapsulations" are things in themselves. The DDS XTypes spec 1.3 (which is not even adopted yet) cleans things up a bit, and it says:

> 7.3.1.7 String Types
> The string container defined by this specification has two element types for which the behavior is defined: Char8 and Char16. Strings of Char8 shall be represented by the IDL type string. Strings of Char16 shall be represented by the IDL type wstring. In either case, any bound shall be retained.

So using UTF-32 (which is what wchar_t typically is) is not a smart choice. Methinks the length should be in octets and a terminating 0 is not supposed to be present.

Incidentally, wide characters suffer from similar problems. I'm not even sure whether an IDL wchar should be a code point or a grapheme ... If a code point, UTF-32 for a character (with endianness determined by the context) makes, if a grapheme, the special-cased sequence-of-octets encoding that section 9.4.1.6 uses makes more sense.

My apologies for also being confused by the situation!

rotu: Oh wow. That is confusing! According to the footnote on this page, CORBA is not a relevant standard for DDS. Also, GIOP is not mentioned on this page at all: https://www.dds-foundation.org/omg-dds-standard/

> DDS™ and CORBA® are two independent connectivity standards ... They both provide application portability (using standard language interfaces) and network interoperability. Both DDS and CORBA were developed by the Object Management Group® (OMG). However, they were designed to support different use cases and integration patterns. ... While the DDS and CORBA standards are independent, both leverage the OMG Interface Definition Language™ (IDL™) to specify data and service types.

It very clearly says XTypes is the relevant serialization format:

> DDS-XTypes v1.2 – defines the DDS Type System and the serialized representation of DDS data.

Here's what XTypes says about strings, which I think we do a poor job of following:

> Objects of `String<Char8>` type shall be represented using the UTF-8 character encoding. The serialized length of an object of type String<Char8> shall be the number of bytes in the CDR buffer taken by the `String<Char8>` characters, including the terminating NUL character. The serialized length may not be the same as the number of Unicode characters because a single Unicode character encoded using the UTF-8 encoding may take one to four bytes.

> Objects of `String<Char16>` type shall be represented using the UTF-16 character encoding. The serialized length of an object of type `String<Char16>` shall be the number of bytes in the CDR buffer taken by the `String<Char16>` characters. This is twice the number of characters in the string because a single character (in the Basic Multilingual Plane) encoded using UTF-16 takes 2 bytes to serialize.

> The UTF-16 representation of object of type `String<Char16>` shall not include a Byte Order Mark (BOM). The representation shall also not include any terminating NUL character(s).

rotu: That was quite a rabbit hole...

eboasson: Quite the rabbit hole indeed! I'm grateful you went down it, because I had completely forgotten that XTypes nowadays overrides the older ones, and so I would probably have spent some time looking in the wrong places ... (In the olden days, before XTypes, it just referred to CDR, and that GIOP spec is the only definition I have ever been able to find for CDR. Whether everything's properly compatible with that is an interesting but mostly academic question that I'm not inclined to study now.)

My sense of things is that the XTypes simply standardized RTI's practice. Which is fine, because at least there is a specification now. And for sure, we do a poor job of following it.

I think the only strategy that is both reasonable and viable long-term is to follow the XTypes spec.

eboasson (commented):
Thanks for making me laugh: all those "operator<<" &c. disappearing ... so obvious! 😄 There are also plenty of other sensible improvements.

The only problem I see with the changes are that they fails the "test_communication" tests for me, and consistently on complex types. What I suspect is the cause is that the template expansion now always ends up ``memcpy``ing the sequence/array elements, instead of serialising them one-by-one for complex types. I think the only case it ever handled differently was ``std::string``, so specializing that one or changing the way strings are handled in ``TypeSupport_impl.h`` should take care of that.

rotu:
Hmm. Where are the "test_communication" tests and how do I run them locally?

eboasson:
@rotu the sources are in ``src/ros2/system_tests/test_communication``, the built/configured ones show up in ``build/test_communication``. There are a great many different cases, specialised for different RMW implementations:

* ``test_serialize__rmw_cyclonedds_cpp`` seems to just be a local serialise/deserialise test that you can run easily, and is quite useful for a quick check;
* ``test_messages_c__rmw_cyclonedds_cpp`` is another one that (I think) exercises the C representation;
* the ``test_publisher_subscriber_...`` ones are tests where two different processes are used, and they can even come from different RMW implementations (as the names suggest).

All of these can be run using ``ctest``. What I find to work well as a quick sanity check is ``ctest --output-on-failure -R cyclonedds`` in this ``test_communication`` directory: that simply runs all known tests that have ``cyclonedds`` in the name, so including interoperability tests with other implementations.

I am not sure if these tests get built by default, they certainly do get built with ``-DBUILD_TESTING=ON`` among the CMake arguments on the colcon command line (i.e., ``colcon build --symlink-install --cmake-args -DBUILD_TESTING=ON``). One look in that ``build/test_communication`` directory should tell you whether that setting is needed.

rotu:
Thanks @eboasson! I'll get this working eventually :-p. In the meantime, could you review #41? That's supposed to be merged in before this PR.

rotu:
Oops. I just realized this actually harms performances for messages with large, homogeneous arrays/sequences. Will fix that.

rotu:
:-1: It's still slower and I definitely need to simplify the code.

rotu:
@eboasson, @joespeed

After lots of false starts, this is finally somewhere I'm happy with. The code is cleaner and more maintainable, and it is FAST. I used the Apex.ai benchmarks and here are the latencies (min/mean/max; lower is better). Keep in mind this is only changing the serialization side. I have not yet done deserialization and significant gains are to be made there too.

`perf_test -c ROS2 -t Array60k --max_runtime 10 --rate 3000 -s 3`
Before: 0.05/0.38/17.5 ms
After: 0.06/0.29/6.2 ms

`perf_test -c ROS2 -t Struct32k --max_runtime 10 --rate 100 -s 3`
Before: 1.3/2.3/14.8
After: 0.4/0.8/4.7

eboasson (approved):

rmw_cyclonedds_cpp/src/TypeSupport2.hpp
```diff
@@ -0,0 +1,510 @@
...
+      case ROSIDL_TypeKind::INT64:
+        return sizeof(int_least64_t);
+      case ROSIDL_TypeKind::STRING:
+      case ROSIDL_TypeKind::WSTRING:
+      case ROSIDL_TypeKind::MESSAGE:
+        throw std::runtime_error(
+                "not a primitive value type: " +
+                std::to_string(std::underlying_type_t<ROSIDL_TypeKind>(m_type_kind)));
+    }
+  }
```
dirk-thomas: Compiler warning: `warning: control reaches end of non-void function [-Wreturn-type]`

dirk-thomas:
I assume this change is also the reason that a lot of the tests with CycloneDDS started to fail: see https://ci.ros2.org/job/ci_linux/8821/

Locally I see the folowing assertion:
> `../rmw_cyclonedds_cpp/src/serdata.cpp:145: ddsi_serdata* serdata_rmw_from_ser(const ddsi_sertopic*, ddsi_serdata_kind, const nn_rdata*, size_t): Assertion `off < size' failed.`

dirk-thomas:
@rotu Just to clarify: did you run any of the ROS 2 tests before merging this? Please do so for future changes. We can always trigger a set of CI builds across all platforms on demand (its easier when the changes are on a branch in the rep rather than a fork).

rmw_cyclonedds_cpp/src/TypeSupport2.hpp
rotu: Thank you. Please take a look at https://github.com/ros2/rmw_cyclonedds/pull/77 which should address this issue

rotu:
> @rotu Just to clarify: did you run any of the ROS 2 tests before merging this? Please do so for future changes. We can always trigger a set of CI builds across all platforms on demand (its easier when the changes are on a branch in the rep rather than a fork).

I did, but I only ran the tests in system_tests.

Do you have any tips for setting up automated testing? (A new CI that I can trigger builds on or how to trigger local clean builds?)

dirk-thomas:
> Do you have any tips for setting up automated testing? (A new CI that I can trigger builds on or how to trigger local clean builds?)

Not at the moment. The CI builds building everything from source are triggered manually since they rely on sparse resources like macOS / Windows machines. The existing PRs jobs only build and test the packages in the repository containing the PR which isn't sufficient in this case.

I would recommend a local workspace with some set of ROS 2 packages to test with (e.g. `demos` and `system_tests`). We can also add you to the ROS 2 Jenkins so you would be able to trigger the CI builds yourself (rather then asking people to do it for you).

ros-discourse:

https://discourse.ros.org/t/ros-2-for-consumer-robotics/11860/1

PR #44, opened 2019-09-24
Return NODE_NAME_NON_EXISTENT instead of ERROR.

evshary:
There are some return value errors in CI test of rcl and rclaction, like [this](http://build.ros2.org/job/Eci__nightly-cyclonedds_ubuntu_bionic_amd64/41/testReport/rcl/TestGraphFixture__rmw_cyclonedds_cpp/test_rcl_get_publisher_names_and_types_by_node/).

I think we should have the return value `RMW_RET_NODE_NAME_NON_EXISTENT` if we can't get the node.

Signed-off-by: evshary <evshary@gmail.com>

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -2442,9 +2442,9 @@ static rmw_ret_t get_node_guids(
   if (ret != RMW_RET_OK) {
     return ret;
   } else if (guids.size() == 0) {
-    /* It appears the get_..._by_node operations are supposed to return ERROR if no such node
-       exists */
-    return RMW_RET_ERROR;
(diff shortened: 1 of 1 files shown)


eboasson:
I agree — see also https://github.com/ros2/rmw/pull/182, https://github.com/ros2/rmw_opensplice/pull/281 and https://github.com/ros2/rmw_fastrtps/pull/311.

dirk-thomas:
I would expect you to need separate branches for this. The `master` branch will work with the master branches of ROS 2 but e.g. for Dashin this won't work since this new error code isn't available in Dashing (and I would expect it not being backported).

eboasson:
Oops. Thanks for warning me, @dirk-thomas. Clearly I'm not quite used to tracking moving targets yet — which is quite strange for someone who once worked for a radar company.

Methinks #45 should solve it.

PR #46, opened 2019-10-09
update signature for added pub/sub options

eboasson:
This relates to https://github.com/ros2/rmw/pull/187.

Before merging, it'd probably be best to make a "dashing" branch.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1088,10 +1088,11 @@ extern "C" rmw_ret_t rmw_fini_publisher_allocation(rmw_publisher_allocation_t * 
 extern "C" rmw_publisher_t * rmw_create_publisher(
   const rmw_node_t * node,
   const rosidl_message_type_support_t * type_supports, const char * topic_name,
-  const rmw_qos_profile_t * qos_policies)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I would suggest holding off on branch and get https://github.com/ros2/rmw/issues/188#issuecomment-540058781 resolved first. Than you should be easily able to keep supporting both ROS distros from the same branch.

rotu:
I wasn't able to find a solution with SFINAE to make it cross-compatible, but I think I found a decent enough alternative:

https://github.com/eboasson/rmw_cyclonedds/pull/1

rotu:
I figured out how to do this cleanly with SFINAE! I prefer this approach, but I'll leave you guys to choose.

https://github.com/eboasson/rmw_cyclonedds/pull/2

A few tricks:
1. The `extern "C"` linkage specifier only needs to be on the function *declaration* in the `rmw.h` header and will link the matching overload.
2. I had to declare the new types (e.g. `rmw_subscription_options_t`) to be template parameters, so that they don't trigger an unknown type name.
3. Where I used new properties of existing classes, I had to replace the class name (e.g. `rmw_subscription_t`) with a template argument so it views it as a *substitution failure* instead of a *missing member*.

rotu:
Good to know. What’s the failure on MacOS?

eboasson:
Closing because #51 sorted it out

Issue #47, opened 2019-10-09
Action-server fails (with an action-client to self)


orduno:
## Bug report

**Required Info:**

- Operating System: Ubuntu 18.04
- Installation type: source
- Version or commit hash: 1205ab5ea223a72253db9c36b19707978b122a9e
- DDS implementation: CycloneDDS :smile:
- Client library (if applicable): rclcpp

#### Steps to reproduce issue
 - Build [this](https://github.com/orduno/examples/blob/issues/action_self_client_within_server/rclcpp/minimal_action_server/member_functions.cpp) modified example of the rclcpp minimal action server example.
- Run it:
  `ros2 run examples_rclcpp_minimal_action_server action_server_member_functions`
- Send an action request through the terminal:
  `ros2 action send_goal /fibonacci example_interfaces/action/Fibonacci '{order: 10}'`

#### Expected behavior
Action server receives the request and starts sending feedback:
```bash
[INFO] [minimal_action_server]: Received goal request with order 10
[INFO] [minimal_action_server]: Executing goal
[INFO] [minimal_action_server]: Publish Feedback
[INFO] [minimal_action_server]: Publish Feedback
...
```

#### Actual behavior
Node crashes:
```bash
[INFO] [minimal_action_server]: Received goal request with order 10
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  error taking goal response: error not set
[INFO] [minimal_action_server]: Executing goal
```

#### Additional information

As you inspect the code you'll notice something a bit unorthodox: we have an action-server created on `MinimalActionServer` and an action-client on an internal node, both actions have the same type and name. Also, nodes are spinning on separate executors.

The idea behind this is that the user has the option to request the execution of an action by publishing a `/trigger` message. This is a simplified version of something we're doing in `ros2/nav2`.

As mentioned above, if the user sends the request through the action interface the node crashes. However if the user sends the request as a message:

```bash
ros2 topic pub /trigger std_msgs/Empty
```

it is successfully processed :thinking:

rotu:
@orduno Thanks for distilling this down to a digestible issue!

orduno:
Also, I wonder if we might run into the same issue if we replace the actions with services. I haven't tried that though.

eboasson:
The issue is — or appears to be — that I incorrectly assumed all client code would be able to handle spurious events.

[...]

PR #48, opened 2019-10-11
Avoid spurious client wakeups (#47)

eboasson:
This PR adds code for only returning those clients from ``rmw_wait`` that really have a response waiting. This prevents ``rmw_take_response`` from ever returning "no errors, but no data either", which the rclcpp_action client executor doesn't like.

There are arguably more elegant ways of doing this, with two very obvious ones:
* Content filtering: then the responses to other clients' requests never enter the reader history, and so never cause trouble; the downside is the current, rather limited and inelegant, content filter support in Cyclone requires having to allocate, deserialise and free each incoming message just for the filtering.
* Only "taking" those responses that are meant for other clients in the (newly added) ``check_client_trigger``, leaving the relevant ones in there. This would avoid all the complications inherent in caching data and the interaction of those caches with the implementation of ``rmw_wait``. The only problem is that ``dds_readcdr`` doesn't currently exist in Cyclone, requiring more carefully scheduling of PRs for a fix to become available. It made more sense to me to get something testable first.

I'll happily wait with merging this PR for a bit. Perhaps none of this will be needed if rclcpp_action changes (see https://github.com/ros2/rclcpp/issues/886) although I don't know if dashing would then also be updated. Or perhaps I can get ``dds_readcdr`` into Cyclone quickly enough.

### DIFF
--- a/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdata.hpp
+++ b/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/serdata.hpp
@@ -74,4 +74,6 @@ struct ddsi_serdata * serdata_rmw_from_serialized_message(
   const struct ddsi_sertopic * topiccmn,
   const void * raw, size_t size);
 
+uint64_t serdata_response_guid(const struct ddsi_serdata * dcmn);
+
 #endif  // RMW_CYCLONEDDS_CPP__SERDATA_HPP_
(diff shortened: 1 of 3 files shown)


jacobperron:
@eboasson Let me know if https://github.com/ros2/rclcpp/pull/888 works as an alternative.

eboasson:
@jacobperron I finally had a chance to try it (on Eloquent on macOS 10.15) and it solves the problem. I am sure this doesn't come as a surprise to you, but I did want to confirm it. 🙂

eboasson:
Closing because of https://github.com/ros2/rclcpp/pull/888.

PR #49, opened 2019-10-18
Use right event info for RMW_EVENT_LIVELINESS_LOST

rotu:
Looks like a copy-paste mistake


eboasson:
> Looks like a copy-paste mistake

Definitely. Thanks for catching that one!

PR #50, opened 2019-10-18
Add localhost boolean parameter to create node function

BMarchi:
As the title says, add the localhost boolean parameter for the create node function. Connected to [this](https://github.com/ros2/ros2/issues/798)

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -491,7 +491,8 @@ static std::string get_node_user_data(const char * node_name, const char * node_
 extern "C" rmw_node_t * rmw_create_node(
   rmw_context_t * context, const char * name,
   const char * namespace_, size_t domain_id,
-  const rmw_node_security_options_t * security_options)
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -491,7 +491,8 @@ static std::string get_node_user_data(const char * node_name, const char * node_
 extern "C" rmw_node_t * rmw_create_node(
   rmw_context_t * context, const char * name,
   const char * namespace_, size_t domain_id,
-  const rmw_node_security_options_t * security_options)
+  const rmw_node_security_options_t * security_options,
+  bool localhost_only)
```
dirk-thomas: The implementation block should contain `(void)localhost_only` to avoid a compiler warning.

dirk-thomas:
@eboasson Can you please follow up on this and implement the necessary logic to limit the communication to localhost when the flag is set? Thank you.

eboasson:
> @eboasson Can you please follow up on this and implement the necessary logic to limit the communication to localhost when the flag is set? Thank you.

@dirk-thomas certainly.

PR #51, opened 2019-10-18
Compatibility before and after RMW pub/sub option structures

rotu:
Includes changes in #46, in a way that makes it work both before and after https://github.com/ros2/rmw/pull/187

### DIFF
--- a/rmw_cyclonedds_cpp/CMakeLists.txt
+++ b/rmw_cyclonedds_cpp/CMakeLists.txt
@@ -72,6 +72,13 @@ ament_target_dependencies(rmw_cyclonedds_cpp
 
 configure_rmw_library(rmw_cyclonedds_cpp)
 
+target_compile_definitions(${PROJECT_NAME}
+  PRIVATE
+    RMW_VERSION_MAJOR=${rmw_VERSION_MAJOR}
+    RMW_VERSION_MINOR=${rmw_VERSION_MINOR}
+    RMW_VERSION_PATCH=${rmw_VERSION_PATCH}
+)
+
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
 target_compile_definitions(${PROJECT_NAME}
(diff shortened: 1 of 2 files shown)


rmw_cyclonedds_cpp/CMakeLists.txt
```diff
@@ -72,6 +72,18 @@ ament_target_dependencies(rmw_cyclonedds_cpp
 
 configure_rmw_library(rmw_cyclonedds_cpp)
 
+string(REGEX REPLACE "([0-9]*)\\.([0-9]*)\\.([0-9]*)((|-).*)" "\\1;\\2;\\3;\\5" parsed_rmw_version "${rmw_VERSION}")
+list(GET parsed_rmw_version 0 rmw_version_major)
+list(GET parsed_rmw_version 1 rmw_version_minor)
+list(GET parsed_rmw_version 2 rmw_version_patch)
```
dirk-thomas: No need to parse those manually. CMake already sets `MAJOR` / `MINOR` / `PATCH` variables based on the `rmw_VERSION` variable.

rotu: Very good call!

dirk-thomas:
The functional changes look good to me. Leaving it up to someone else to merge.

Why are the non-functional changes (line wrapping etc.) part of this PR?

eboasson:
Looks good to me, too. Just verified it on current Eloquent on macOS as well as building against an old binary install of dashing. I'll merge if @dirk-thomas is happy with it too.

Many thanks @rotu for helping out with the cyclone RMW and to @dirk-thomas for getting the RMW versioning in so quickly!

rotu:
> Why are the non-functional changes (line wrapping etc.) part of this PR?

I reformatted the code local to my changes to comply with the C++ developer guide. This should not be controversial; please don't bikeshed.

dirk-thomas:
> This should not be controversial; please don't bikeshed.

The change itself is not controversial. I would just suggest to use a separate PR for unrelated changes - which I don't think is bikeshedding but a commonly agreed upon rational for making pull requests. Anyway - I am not maintaining this repo - so it is just an outsider comment. No need to do anything about it.

PR #52, opened 2019-10-18
unbreak Dashing build after #50

rotu:
PR #50 broke CycloneDDS on Dashing. This fixes it.


rotu:
FYI, @eboasson, I think it's time to cut a dedicated Dashing branch.

eboasson:
Thank you very much, @rotu!

> FYI, @eboasson, I think it's time to cut a dedicated Dashing branch.

From what I gather the RMW interface of Eloquent should not change anymore now, so let's wait for a little bit longer. The pain of maintaining multiple branches is also significant, at least in my experience ...

PR #53, opened 2019-10-21
zero copy api for cyclonedds

Karsten1987:
addresses the missing functions for zero-copy API as noted in here: https://github.com/ros2/ros2/issues/785#issuecomment-544337245

Connects to https://github.com/ros2/rclcpp/pull/896

Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1237,6 +1237,16 @@ extern "C" rmw_ret_t rmw_return_loaned_message_from_publisher(
   return RMW_RET_UNSUPPORTED;
 }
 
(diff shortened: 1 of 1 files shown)


eboasson:
@Karsten1987 @dirk-thomas This PR doesn't work for me locally because it provides the interface as amended by https://github.com/ros2/rmw/pull/192, rather than the one currently implemented by ``rcl`` on master.

Adding the "old" interface (``rmw_return_loaned_message`` and ``rmw_release_loaned_message``) solves the problem. It is somewhat silly to add them only to remove them again in the very near future, but as @dirk-thomas mentioned on https://github.com/ros2/ros2/issues/785, we really need to get this thing building again. Pushing this now should fix tonight's CI run.

Hopefully the two of you don't mind this temporary thing. If you do, too bad for me ...

PR #60, opened 2019-10-27
Support localhost-only communications

eboasson:
This PR implements the "localhost only" feature of https://github.com/ros2/ros2/issues/798 for which the interface change was introduced into rmw_cyclonedds by #50. I am not entirely happy with the interface in Cyclone, so there'll likely be a follow-up when Cyclone's interface is fixed, but that'll be trivial.

It should avoid the problem of hard-coded IP address (as discussed in https://github.com/ros2/ros2/issues/798#issuecomment-544596424 and the comments that follow) by relying on the name "localhost", but that assumes hostname lookup for "localhost" always succeeds. That seems like a pretty reasonable assumption. In any case, configuration overrides can still be done the normal way, so on machines where "localhost" doesn't resolve correctly, there is a fallback.


dirk-thomas:
I don't know enough about this repo but if it works :+1: from me to merge this.

PR #62, opened 2019-11-05
WIP: Only use cpp typesupport

rotu:


dirk-thomas:
With C typesupport removed do C messages (e.g. coming from rclpy) still work?

rotu:
> With C typesupport removed do C messages (e.g. coming from rclpy) still work?

Why shouldn't they? The messages are the same on the wire, and the cpp introspection is just a thin layer on top of the c introspection. I thought the default generators *always* generate the cpp introspection?

dirk-thomas:
But if Python publishes a message it hands an instance of a C message to the rmw API. In order to read / interpret that you need the corresponding C introspection typesupport.

rotu:
Gotcha. Is this only local? Or does typesupport travel on the wire?

dirk-thomas:
Typesupport provides information / functionality for a data type in a specific programming language (it could be introspection information, it could be a conversion function between the ROS type and an external type - usually the one from a DDS vendor). It never affects the wire representation of a message.

rotu:
Gotcha. I figured the ROS2 runtime interpreted between different typesupports and the introspection typesupport was just a language-specific layer. I guess I have some reworking to do!

Issue #65, opened 2019-11-11
failed to create domain - Precondition Not Met


dirk-thomas:
The aarch64 build shows a lot of failing tests, e.g. https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_extra_rmw_release/531/testReport/(root)/projectroot/test_publisher__rmw_cyclonedds_cpp/:

```
[ERROR] [rmw_cyclonedds_cpp]: rmw_create_node: failed to create domain, error Precondition Not Met
```

dirk-thomas:
The same error seems to be happening with x64 builds on [Ubuntu](https://ci.ros2.org/view/nightly/job/nightly_linux_extra_rmw_release/524/testReport/(root)/projectroot/test_publisher__rmw_cyclonedds_cpp/) and [CentOS](https://ci.ros2.org/view/nightly/job/nightly_linux-centos_debug/225/testReport/(root)/projectroot/test_publisher__rmw_cyclonedds_cpp/).

eboasson:
And on macOS ...

The failures appear to all be related to cases where a single process creates some nodes, deletes them all, and then attempts to create new nodes. I really was asleep at the wheel when I reviewed and merged the original implementation of ``dds_create_domain`` ...

#66 fixes it for me locally (for the test cases that checked)

dirk-thomas:
Closed by #66.

PR #66, opened 2019-11-11
Address "Precondition not met" on rmw_create_node (#65)

eboasson:
Cyclone's original code for dds_create_domain() appears to create a domain entity that is automatically deleted when the last attached participant disappears.  In reality, it leaks a reference while returning DDS_RETCODE_OK, making it appear as-if it is a regular entity that remains into existence until explicitly deleted.

The RMW code assumed that it would be automatically deleted when the last node was destroyed and that a subsequent call to rmw_create_node could create the domain anew.  This then fails with "precondition not met".

In an upcoming fixed version of dds_create_domain() the domain entities will behave normally (return a handle, require an explicit delete).

This commit provides a workaround for the bug in the original implementation: by recovering the handle from the first participant the preceding commits that were intended to future-proof the code will ensure that the entity now gets deleted explicitly.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -548,9 +548,6 @@ static void unref_ppant()
 #if SUPPORT_LOCALHOST
 static void node_gone_from_domain_locked(dds_domainid_t did)
 {
-  /* The initial support for dds_create_domain in Cyclone results in domains that get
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I will go ahead and merge this and kick of a new full build to see where we stand.

dirk-thomas:
[Build Status](http://build.ros2.org/view/Eci/job/Eci__nightly-cyclonedds_ubuntu_bionic_amd64/90/)

Issue #71, opened 2019-11-21
usage of getenv results in warning on Windows


dirk-thomas:
See https://ci.ros2.org/view/packaging/job/packaging_windows/1566/warningsResult/

Introduced in #60.

Use the function from [rcutils](https://github.com/ros2/rcutils/blob/fe215b4a3db580358f09bda6269e1356165fa920/include/rcutils/get_env.h#L62) instead.

nuclearsandwich:
Resolved by https://github.com/ros2/rmw_cyclonedds/pull/73

PR #73, opened 2019-11-25
Use rcutils_get_env() instead of getenv() (#71)

eboasson:
@dirk-thomas there's this:

https://github.com/ros2/rcutils/blob/540d9472b0c7de4b029f2e26f3388e3b9a5bb1eb/include/rcutils/get_env.h#L52

There is no way of knowing who will be calling ``rcutils_get_env`` or when, so I think this limitation is asking for trouble. I presume this has been considered in the past and deemed acceptable, and if so (and assuming my change gets through review and test build) feel free to simply merge it.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -24,8 +24,8 @@
 #include <utility>
 #include <regex>
 
-#include "rcutils/logging_macros.h"
 #include "rcutils/get_env.h"
+#include "rcutils/logging_macros.h"
 #include "rcutils/strdup.h"
 
 #include "rmw/allocators.h"
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -25,6 +25,7 @@
 #include <regex>
 
 #include "rcutils/logging_macros.h"
+#include "rcutils/get_env.h"
 #include "rcutils/strdup.h"
 
 #include "rmw/allocators.h"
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> There is no way of knowing who will be calling `rcutils_get_env` or when, so I think this limitation is asking for trouble.

While that is correct the same limitation is already the case with your current code using `getenv()`: https://en.cppreference.com/w/c/program/getenv

The thread safety is orthogonal and since the patch resolves #71 merging makes sense imo.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -25,6 +25,7 @@
 #include <regex>
 
 #include "rcutils/logging_macros.h"
+#include "rcutils/get_env.h"
```
dirk-thomas: Nitpick: please keep the includes in alphabetical order.

eboasson:
> While that is correct the same limitation is already the case with your current code using getenv(): https://en.cppreference.com/w/c/program/getenv

Funny that, one would have thought "they" would have learnt by now that such a weak specification is rather unhelpful ... Anyway, I totally agree with your conclusion (and suspect that for the platforms supported by ROS2 it is actually safe (or at least safe to call ``getenv()`` in the absence of calls to ``putenv()`` or equivalent).

dirk-thomas:
> I ... suspect that for the platforms supported by ROS2 it is actually safe (or at least safe to call getenv() in the absence of calls to putenv() or equivalent

I wouldn't say that :wink: Even without `putenv()` being called if two threads concurrently call `getenv()` the first call returns a pointer too "foo" and the second one a pointer to "bar" but since the function might only be backed by one storage the second call might also affect the content of pointer returned by the first call. So the least unsafe usage I would imagine (if `getenv` doesn't use a thread specific storage) is to assign the result immediately to a `std::string`.

(Nothing to be done here though I would say.)

PR #77, opened 2019-12-06
Mark code that should be unreachable

rotu:
Introduce a new `[[noreturn]] unreachable()` function that marks code as unreachable and throws a logic error if it is executed.
Fix build error due to Windows min/max macros.
Fix linker errors from referring to a non-constexpr extern from a constexpr.
Fix warnings about narrowing conversions.

### DIFF
--- a/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/exception.hpp
+++ b/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/exception.hpp
@@ -15,8 +15,8 @@
 #ifndef RMW_CYCLONEDDS_CPP__EXCEPTION_HPP_
 #define RMW_CYCLONEDDS_CPP__EXCEPTION_HPP_
 
+#include <stdexcept>
 #include <string>
-#include <exception>
 
 namespace rmw_cyclonedds_cpp
 {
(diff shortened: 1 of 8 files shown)


dirk-thomas:
Fails to build with gcc on Linux: [Build Status](https://ci.ros2.org/job/ci_linux/8822/)

dirk-thomas:
* Linux: [Build Status](https://ci.ros2.org/job/ci_linux/8824/)
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/8743/)

dirk-thomas:
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/8745/)

rotu:
That's weird. Why wasn't the missing #include caught by ament_lint_auto's cpplint IWYU checker?

dirk-thomas:
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/8746/)

rotu:
That one was my fault

dirk-thomas:
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/8747/)

dirk-thomas:
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/8760/)

rotu:
Thank you, Dirk. Could you run CI one more time? I think I addressed all the warnings.

dirk-thomas:
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/8761/)

rotu:
@eboasson Could you take a quick look before this gets merged in?

rotu:
@eboasson I'm on the same page as you.

NOMINMAX actually *is* needed in this code right now for the line `assert(value <= std::numeric_limits<uint32_t>::max());` I have yet to track down all the other places where the min/max macro might sneak in.

I don't love having the "default: unreachable()" either. I tend to use Clang to do my static analysis while coding, and it provides great feedback for non-exhaustive cases. I think you just gave me a great idea, though, if it passes the linters. PR incoming :-)

PR #88, opened 2020-01-21
Support for deadline, lifespan and liveliness qos

dennis-adlink:
This PR adds support for events in `rmw_wait`, which is required for the lifespan and deadline qos to work properly. I've also added code for setting the deadline, lifespan and liveliness qos policies in `create_readwrite_qos`. And when getting qos properties in `get_readwrite_qos`, an infinite duration is no longer returned as 0, as `test_get_actual_qos__rmw_cyclonedds_cpp` expects the value to be >= MAX_INT32

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <mutex>
+#include <unordered_map>
 #include <unordered_set>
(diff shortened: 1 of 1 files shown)


eboasson:
This fixes makes all the rcl tests go green for me locally on macOS in combination with https://github.com/eclipse-cyclonedds/cyclonedds/pull/378 and has no adverse affects that I can see (either from code inspection or building/running locally) without that PR on Cyclone applied. Merging.

PR #91, opened 2020-01-23
rmw_get_topic_endpoint_info doesn't exist on Dashing

eboasson:
... and this PR avoids trying to compile them on Dashing, similarly to how other RMW interface changes have been dealt with in the Cyclone DDS RMW layer so far.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_get_topic_endpoint_info.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_get_topic_endpoint_info.cpp
@@ -14,7 +14,7 @@
 
 #include "rmw_cyclonedds_cpp/rmw_version_test.hpp"
 
-#if RMW_VERSION_GTE(0, 8, 1)
+#if RMW_VERSION_GTE(0, 8, 2)
 
 #include "rmw/error_handling.h"
 #include "rmw/get_topic_endpoint_info.h"
--- a/rmw_cyclonedds_cpp/src/rmw_get_topic_endpoint_info.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_get_topic_endpoint_info.cpp
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "rmw_cyclonedds_cpp/rmw_version_test.hpp"
+
+#if RMW_VERSION_GTE(0, 8, 1)
+
 #include "rmw/error_handling.h"
 #include "rmw/get_topic_endpoint_info.h"
 #include "rmw/topic_endpoint_info_array.h"
@@ -40,3 +44,5 @@ rmw_get_subscriptions_info_by_topic(
   return RMW_RET_UNSUPPORTED;
 }
 }  // extern "C"
+
+#endif
--- a/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/rmw_version_test.hpp
+++ b/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/rmw_version_test.hpp
@@ -1,0 +1,24 @@
+// Copyright 2019 ADLINK Technology
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#ifndef RMW_CYCLONEDDS_CPP__RMW_VERSION_TEST_HPP_
+#define RMW_CYCLONEDDS_CPP__RMW_VERSION_TEST_HPP_
+
+/* True if the version of RMW is at least major.minor.patch */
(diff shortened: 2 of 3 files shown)


charvi-077:
I was also facing the same issue while building . Now its resolved, thanks @eboasson .

rmw_cyclonedds_cpp/src/rmw_get_topic_endpoint_info.cpp
```diff
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "rmw_cyclonedds_cpp/rmw_version_test.hpp"
+
+#if RMW_VERSION_GTE(0, 8, 1)
```
ivanpauno: The missing functions weren't part of release `0.8.1`, so `0.8.1` shouldn't be included. The problem is that we usually bump the version when releasing and not when developing.

I see that `Dashing` is using `master` branch ([1](https://github.com/ros2/ros2/blob/07c28ab41435154f6c1cee2dac33a990e97e9873/ros2.repos#L245)), while eloquent is using it's own branch ([2](https://github.com/ros2/ros2/blob/eloquent/ros2.repos)). Why the incosistence?
If there weren't such inconsistency, including the new functions in `0.8.1` would be a problem.

ivanpauno: Ok, I saw this comment about development versions https://github.com/ros2/rmw/issues/188#issuecomment-540233493, and also this:
https://github.com/ros2/rmw/blob/a3774531e9b7289574501275aafbc3f93120360e/rmw/CMakeLists.txt#L66.

I'm not pretty sure about how it should be used, but I think that should be considered in this PR.

eboasson: @ivanpauno, thanks for looking into it. The ``0.8.1`` comes straight from the Foxy build settings. I guess that one should have been bumped then.

Unless I am mistaken, I'm the one who asked for the rmw versioning stuff to work nicely so that I could have a single source for multiple versions without having to deal with backporting things. That used to be perfectly viable, but I have a feeling https://github.com/ros2/design/pull/250 is going to happen, and that will be too big a change to handle with a bit of conditional compilation. Therefore, it seems multiple branches are likely to become unavoidable. 

Since it turns out a branch had already been created for eloquent, I guess we now should retroactively give dashing a branch as well. Then this PR can simply be closed.

ivanpauno: > @ivanpauno, thanks for looking into it. The 0.8.1 comes straight from the Foxy build settings. I guess that one should have been bumped then.

mmm, it wasn't bumped. The `eloquent` version of `rmw` is indicating `0.8.1`: https://github.com/ros2/rmw/blob/eloquent/rmw/package.xml.
I'm unsure how the versioning was supposed to be used.
@wjwwood @dirk-thomas can you clarify?

> Unless I am mistaken, I'm the one who asked for the rmw versioning stuff to work nicely so that I could have a single source for multiple versions without having to deal with backporting things

Yes, that's true. I agree that mantaining similar versions from the same source is desired.

> Since it turns out a branch had already been created for eloquent, I guess we now should retroactively give dashing a branch as well. Then this PR can simply be closed.

I guess that probably was an error, if it's easier to mantain `eloquent` and `dashing` from the same branch it would be reasonable to merge them again.

> That used to be perfectly viable, but I have a feeling ros2/design#250 is going to happen, and that will be too big a change to handle with a bit of conditional compilation. Therefore, it seems multiple branches are likely to become unavoidable.

Yeah, I agree that after that change it will be easier to branch that to mantain all the versions from the same branch.

dirk-thomas: > I'm unsure how the versioning was supposed to be used.

Whenever the API is broken in `rmw` it should declare that by bumping the dev version to the next upcoming version number: e.g. https://github.com/ros2/rmw/commit/688e53a976970de837c39c11e965b9d56c438e07

wjwwood: Yeah, we should have updated it when merging that. I forgot to do so. It's tricky to remember to do that when pull requests span releases, but we'll just have to be more diligent about it.

wjwwood: I'm working on a pull request to fix it.

wjwwood: https://github.com/ros2/rmw/pull/194

ivanpauno: Thanks for the PR @wjwwood !
@eboasson it should be greater or equal than `0.8.2` then, LGTM with that update.

dirk-thomas:
@eboasson What is the status on this? It would be nice to get the Dashing builds passing again.

eboasson:
In my understanding it requires me (or someone) to change the ``#if RMW_VERSION_GTE(0, 8, 1)`` to ``#if RMW_VERSION_GTE(0, 8, 2)`` and merge it. I've been buried a bit, I am afraid, and doing it now would be too risky ...

dirk-thomas:
While I understand the problem of having little time for many tasks I just want to point out that building Dashing from source using CycloneDDS has been broken since the other PR was merged 16 days ago. That does look bad for ADLINK as well as ROS 2 as a whole. So I would think it is in ADLINKs interest to address such a major regressions earlier than later. (More than two weeks sounds already way to long to me.)

PR #96, opened 2020-02-05
Support ros1-bridge (#72)

eboasson:
This is a companion to https://github.com/eclipse-cyclonedds/cyclonedds/pull/382, implementing the small interface extension for custom topic implementations it requires.

I've verified that the RMW layer with this change applied doesn't break anything (I guess it is pretty obvious from the changes that it won't) when used with a version of cyclone that doesn't have the aforementioned PR applied. Without this PR it won't work with an updated cyclone, so it should be merged before the cyclone PR gets merged. I would prefer to wait until the latter one has had a chance of getting reviewed, just in case there will be changes affecting this PR.

For me locally, using Foxy and Melodic on Ubuntu 18.04, this makes the ros1 bridge demo described in the readme at https://github.com/ros2/ros1_bridge work as expected. I have even tried it in the other direction.

### DIFF
--- a/rmw_cyclonedds_cpp/src/serdata.cpp
+++ b/rmw_cyclonedds_cpp/src/serdata.cpp
@@ -152,6 +152,25 @@ static struct ddsi_serdata * serdata_rmw_from_ser(
   return d.release();
 }
 
+#if DDSI_SERDATA_HAS_FROM_SER_IOV
+static struct ddsi_serdata * serdata_rmw_from_ser_iov(
+  const struct ddsi_sertopic * topic,
+  enum ddsi_serdata_kind kind,
+  ddsrt_msg_iovlen_t niov, const ddsrt_iovec_t * iov,
+  size_t size)
+{
+  auto d = std::make_unique<serdata_rmw>(topic, kind);
+  d->resize(size);
+
+  auto cursor = d->data();
+  for (ddsrt_msg_iovlen_t i = 0; i < niov; i++) {
+    memcpy(cursor, iov[i].iov_base, iov[i].iov_len);
+    cursor = byte_offset(cursor, iov[i].iov_len);
+  }
+  return d.release();
+}
+#endif
+
 static struct ddsi_serdata * serdata_rmw_from_keyhash(
   const struct ddsi_sertopic * topic,
   const struct nn_keyhash * keyhash)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I triggered a packaging job with this change ([Build Status](https://ci.ros2.org/job/ci_packaging_linux/291/)) but it doesn't look like it fixes the failing tests.

eboasson:
@dirk-thomas I had a quick look at the console output and it looks to me like this build is without the changes in cyclone — and if that is indeed the case, no change is behaviour is to be expected. It'd be interesting to try again with an updated cyclone as well (i.e., from https://github.com/eboasson/cyclonedds, also branch ``multiple-sertopic-per-topic``), if you have the time.

dirk-thomas:
The following job also uses the custom branch from CycloneDDS: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/292/)

nuclearsandwich:
I have to bump this from Dashing Patch 6 since the CycloneDDS change is not yet released there. Dashing's source branch of this repository is also still tracking master, although there is an eloquent branch. We'll have to figure out whether Dashing, Eloquent, and Foxy can all be supported from master where we need to start the dashing branch from.

eboasson:
@nuclearsandwich Dashing-thru-Foxy are all fine with the current head of master. We've updated the eloquent branch (GitHub played a nasty trick on me there — I'm sure through a misunderstanding on my part) as part of trying to release the binaries to a newly tagged 0.5.1 using bloom.

Note that the dashing branch never got created simply because rmw_cyclonedds is younger than Dashing ... creating it is easy enough, but it also means updating the overall repos file and I am afraid I might overlook a few things, so I'd prefer it if someone with more knowledge of the system would do that.

nuclearsandwich:
> Dashing-thru-Foxy are all fine with the current head of master.

For most of the repositories supported by the ROS 2 team, we've been pre-emptively creating rosdistro branches when a new releases is made. But if the goal is to develop for all three distros from master I think updating Eloquent to use the master branch and removing the eloquent branch is a reasonable course of action. I'm sure @mjcarroll created it as part of a general sweep when Eloquent was first released.

The alternative is creating a dashing branch and fast-forwarding both it and the eloquent branch to keep them in sync with master when making releases which to me just sounds like extra work.

mjcarroll:
> I'm sure @mjcarroll created it as part of a general sweep when Eloquent was first released.

This is precisely what happened.

PR #99, opened 2020-02-11
Fix dashing builds wrt rmw_topic_endpoint_info_t

dennis-adlink:
Add version check >= 0.8.2 to fix ros2 dashing builds; types rmw_topic_endpoint_info_t and rmw_topic_endpoint_info_array_t don't exist in dashing.


eboasson:
Thanks @dennis-adlink, and with my apologies to everyone for forgetting to pay attention to dashing when reviewing #97 ...

PR #101, opened 2020-02-13
Update usage of rmw_topic_endpoint_info_array

ivanpauno:
Updated after https://github.com/ros2/rmw/pull/196.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -3108,7 +3108,6 @@ get_endpoint_info_by_topic(
       err_msg.str);
     return ret;
   }
-  endpoint_info->count = 0;
 
   node_cache_t node_cache;
   dds_entity_t ppant_rd;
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3120,25 +3119,23 @@ get_endpoint_info_by_topic(
...
-        endpoint_info->info_array[endpoint_info->count])) != RMW_RET_OK)
+        endpoint_info->info_array[i])) != RMW_RET_OK)
     {
-      endpoint_info->count++;
       handle_topic_endpoint_info_array_fini(endpoint_info, allocator);
       dds_delete(ppant_rd);
       return ret;
     }
-    endpoint_info->count++;
+    i++;
```
hidmic: @ivanpauno nit: use prefix increment

ivanpauno: The code base of this repository uses extensively post-increment when the return value is not used.
The google [style-guide](https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement) accepts both pre-increment and post-increment for scalar types.

PR #102, opened 2020-02-18
Correct std::hash return type sizes

mjcarroll:
Corrects recent build warnings in CI (for example https://ci.ros2.org/view/packaging/job/packaging_windows/1648/), because `std::hash` returns a `size_t`

```
serdata.cpp:468, MSBuild, Priority: Normal
--
'initializing': conversion from 'size_t' to 'uint32_t', possible loss of data
```

Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/rmw_cyclonedds_cpp/src/serdata.cpp
+++ b/rmw_cyclonedds_cpp/src/serdata.cpp
@@ -465,8 +465,10 @@ bool sertopic_rmw_equal(
 uint32_t sertopic_rmw_hash(const struct ddsi_sertopic * tpcmn)
 {
   const struct sertopic_rmw * tp = static_cast<const struct sertopic_rmw *>(tpcmn);
-  uint32_t h2 = std::hash<bool>{} (tp->is_request_header);
-  uint32_t h1 = std::hash<std::string>{} (std::string(tp->type_support.typesupport_identifier_));
+  uint32_t h2 = static_cast<uint32_t>(std::hash<bool>{} (tp->is_request_header));
+  uint32_t h1 =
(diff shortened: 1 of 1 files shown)


mjcarroll:
Windows: [Build Status](https://ci.ros2.org/job/ci_windows/9292/)

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -465,9 +465,9 @@ bool sertopic_rmw_equal(
 uint32_t sertopic_rmw_hash(const struct ddsi_sertopic * tpcmn)
 {
   const struct sertopic_rmw * tp = static_cast<const struct sertopic_rmw *>(tpcmn);
-  uint32_t h2 = std::hash<bool>{} (tp->is_request_header);
-  uint32_t h1 = std::hash<std::string>{} (std::string(tp->type_support.typesupport_identifier_));
-  return h1 ^ h2;
+  size_t h2 = std::hash<bool>{} (tp->is_request_header);
+  size_t h1 = std::hash<std::string>{} (std::string(tp->type_support.typesupport_identifier_));
+  return static_cast<uint32_t>(h1 ^ h2);
```
dirk-thomas: Nitpick: first doing the operator and then throwing away part of the result seems unnecessary. How about casting both arguments (or the variables above) and then only do the operator for the necessary bits? (Not that it will matter performance wise.)

mjcarroll: Done in 02d0e18

PR #106, opened 2020-02-28
API changes to sync with one Participant per Context change in rmw_fastrtps

ivanpauno:
Needed since https://github.com/ros2/rmw/pull/189.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -646,7 +646,7 @@ static std::string get_node_user_data(const char * node_name, const char * node_
 extern "C" rmw_node_t * rmw_create_node(
   rmw_context_t * context, const char * name,
   const char * namespace_, size_t domain_id,
-  const rmw_node_security_options_t * security_options
+  const rmw_security_options_t * security_options
 #if RMW_VERSION_GTE(0, 8, 1)
   , bool localhost_only
 #endif
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -646,7 +646,11 @@ static std::string get_node_user_data(const char * node_name, const char * node_
 extern "C" rmw_node_t * rmw_create_node(
   rmw_context_t * context, const char * name,
   const char * namespace_, size_t domain_id,
+#if RMW_VERSION_GTE(0, 8, 2)
   const rmw_security_options_t * security_options
+#else
+  const rmw_node_security_options_t * security_options
+#endif
 #if RMW_VERSION_GTE(0, 8, 1)
   , bool localhost_only
 #endif
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -358,6 +358,10 @@ extern "C" rmw_ret_t rmw_init_options_init(
   init_options->implementation_identifier = eclipse_cyclonedds_identifier;
   init_options->allocator = allocator;
   init_options->impl = nullptr;
+  init_options->localhost_only = RMW_LOCALHOST_ONLY_DEFAULT;
+  init_options->domain_id = RMW_DEFAULT_DOMAIN_ID;
+  init_options->name = NULL;
+  init_options->security_options = rmw_get_zero_initialized_security_options();
   return RMW_RET_OK;
 }
 
@@ -374,19 +378,37 @@ extern "C" rmw_ret_t rmw_init_options_copy(const rmw_init_options_t * src, rmw_i
     RMW_SET_ERROR_MSG("expected zero-initialized dst");
     return RMW_RET_INVALID_ARGUMENT;
   }
+  const rcutils_allocator_t * allocator = &src->allocator;
+  rmw_ret_t ret = RMW_RET_OK;
+
+  allocator->deallocate(dst->name, allocator->state);
   *dst = *src;
-  return RMW_RET_OK;
+  dst->name = NULL;
+  dst->security_options = rmw_get_zero_initialized_security_options();
+
+  dst->name = rcutils_strdup(src->name, *allocator);
+  if (src->name && !dst->name) {
+    ret = RMW_RET_BAD_ALLOC;
+    goto fail;
+  }
+  return rmw_security_options_copy(&src->security_options, allocator, &dst->security_options);
+fail:
+  allocator->deallocate(dst->name, allocator->state);
+  return ret;
 }
 
 extern "C" rmw_ret_t rmw_init_options_fini(rmw_init_options_t * init_options)
 {
   RMW_CHECK_ARGUMENT_FOR_NULL(init_options, RMW_RET_INVALID_ARGUMENT);
-  RCUTILS_CHECK_ALLOCATOR(&init_options->allocator, return RMW_RET_INVALID_ARGUMENT);
+  rcutils_allocator_t & allocator = init_options->allocator;
+  RCUTILS_CHECK_ALLOCATOR(&allocator, return RMW_RET_INVALID_ARGUMENT);
   RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
     init_options,
     init_options->implementation_identifier,
     eclipse_cyclonedds_identifier,
     return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  allocator.deallocate(init_options->name, allocator.state);
+  rmw_security_options_fini(&init_options->security_options, &allocator);
   *init_options = rmw_get_zero_initialized_init_options();
   return RMW_RET_OK;
 }
@@ -645,11 +667,9 @@ static std::string get_node_user_data(const char * node_name, const char * node_
 
 extern "C" rmw_node_t * rmw_create_node(
   rmw_context_t * context, const char * name,
-  const char * namespace_, size_t domain_id,
-#if RMW_VERSION_GTE(0, 8, 2)
-  const rmw_security_options_t * security_options
-#else
-  const rmw_node_security_options_t * security_options
+  const char * namespace_, size_t domain_id
+#if !RMW_VERSION_GTE(0, 8, 2)
+  , const rmw_security_options_t * security_options
 #endif
 #if RMW_VERSION_GTE(0, 8, 1)
   , bool localhost_only
@@ -671,7 +691,9 @@ extern "C" rmw_node_t * rmw_create_node(
   static_cast<void>(domain_id);
   const dds_domainid_t did = DDS_DOMAIN_DEFAULT;
 #endif
+#if !RMW_VERSION_GTE(0, 8, 2)
   (void) security_options;
+#endif
   rmw_ret_t ret;
   int dummy_validation_result;
   size_t dummy_invalid_index;
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -3044,7 +3044,8 @@ extern "C" rmw_ret_t rmw_get_node_names(
   RET_WRONG_IMPLID(node);
   auto node_impl = static_cast<CddsNode *>(node->data);
   if (rmw_check_zero_rmw_string_array(node_names) != RMW_RET_OK ||
-    rmw_check_zero_rmw_string_array(node_namespaces) != RMW_RET_OK)
+    rmw_check_zero_rmw_string_array(node_namespaces) != RMW_RET_OK ||
+    rmw_check_zero_rmw_string_array(security_contexts) != RMW_RET_OK)
   {
     return RMW_RET_ERROR;
   }
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -3037,77 +3037,8 @@ static rmw_ret_t do_for_node_user_data(
   return do_for_node(node_impl, f);
 }
 
-extern "C" rmw_ret_t rmw_get_node_names(
-  const rmw_node_t * node,
-  rcutils_string_array_t * node_names,
-  rcutils_string_array_t * node_namespaces)
-{
-  RET_WRONG_IMPLID(node);
-  auto node_impl = static_cast<CddsNode *>(node->data);
-  if (rmw_check_zero_rmw_string_array(node_names) != RMW_RET_OK ||
-    rmw_check_zero_rmw_string_array(node_namespaces) != RMW_RET_OK)
-  {
-    return RMW_RET_ERROR;
-  }
-
-  std::vector<std::pair<std::string, std::string>> ns;
-  const auto re = std::regex("^name=([^;]*);namespace=([^;]*);", std::regex::extended);
-  auto oper =
-    [&ns, re](const dds_builtintopic_participant_t & sample, const char * ud) -> bool {
-      std::cmatch cm;
-      static_cast<void>(sample);
-      if (std::regex_search(ud, cm, re)) {
-        ns.push_back(std::make_pair(std::string(cm[1]), std::string(cm[2])));
-      }
-      return true;
-    };
-  rmw_ret_t ret;
-  if ((ret = do_for_node_user_data(node_impl, oper)) != RMW_RET_OK) {
-    return ret;
-  }
-
-  rcutils_allocator_t allocator = rcutils_get_default_allocator();
-  if (rcutils_string_array_init(node_names, ns.size(), &allocator) != RCUTILS_RET_OK ||
-    rcutils_string_array_init(node_namespaces, ns.size(), &allocator) != RCUTILS_RET_OK)
-  {
-    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
-    goto fail_alloc;
-  }
-  size_t i;
-  i = 0;
-  for (auto && n : ns) {
-    node_names->data[i] = rcutils_strdup(n.first.c_str(), allocator);
-    node_namespaces->data[i] = rcutils_strdup(n.second.c_str(), allocator);
-    if (!node_names->data[i] || !node_namespaces->data[i]) {
-      RMW_SET_ERROR_MSG("rmw_get_node_names for name/namespace");
-      goto fail_alloc;
-    }
-    i++;
-  }
-  return RMW_RET_OK;
-
-fail_alloc:
-  if (node_names) {
-    if (rcutils_string_array_fini(node_names) != RCUTILS_RET_OK) {
-      RCUTILS_LOG_ERROR_NAMED(
-        "rmw_cyclonedds_cpp",
-        "failed to cleanup during error handling: %s", rcutils_get_error_string().str);
-      rcutils_reset_error();
-    }
-  }
-  if (node_namespaces) {
-    if (rcutils_string_array_fini(node_namespaces) != RCUTILS_RET_OK) {
-      RCUTILS_LOG_ERROR_NAMED(
-        "rmw_cyclonedds_cpp",
-        "failed to cleanup during error handling: %s", rcutils_get_error_string().str);
-      rcutils_reset_error();
-    }
-  }
-  return RMW_RET_BAD_ALLOC;
-}
-
-#if RMW_VERSION_GTE(0, 8, 2)
-extern "C" rmw_ret_t rmw_get_node_names_with_security_contexts(
+extern "C" rmw_ret_t rmw_get_node_names_impl(
+  std::regex re,
   const rmw_node_t * node,
   rcutils_string_array_t * node_names,
   rcutils_string_array_t * node_namespaces,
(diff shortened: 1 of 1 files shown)


eboasson:
It certainly looks perfectly reasonable to me. The answer to @hidmic's question is a bit more difficult: this involves a few rather large pull requests and we need to first figure out what changes are actually needed. The change to using a single participant is trivial, it is the bit that relates to introspection that is less obvious.

Currently, the Cyclone RMW layer just queries its discovery data on demand and otherwise ignores all changes save for triggering the graph guard condition. I guess it will now have to call  functions in ``rmw_dds_common`` for each change, shuttle the new discovery topics back-and-forth, and forward the introspection operations to it. How much work those changes will be I don't know without having a good look at the interfaces.

So the only thing I can say definitively is that it we'll be looking into the details and hope to be able to this soon.

ivanpauno:
> The change to using a single participant is trivial, it is the bit that relates to introspection that is less obvious.

Yes, completly agreed.

> How much work those changes will be I don't know without having a good look at the interfaces.

Please, let me know if I can help, or if anything has to be modified in `rmw_dds_common` to make it more suitable for `cyclonedds`.

> So the only thing I can say definitively is that it we'll be looking into the details and hope to be able to this soon.

Great, thanks!

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2932,7 +2965,7 @@ extern "C" rmw_ret_t rmw_get_node_names(
   }
 
   std::vector<std::pair<std::string, std::string>> ns;
-  const auto re = std::regex("^name=(.*);namespace=(.*);$", std::regex::extended);
+  const auto re = std::regex("^name=([^;]*);namespace=([^;]*);", std::regex::extended);
```
hidmic: @ivanpauno a hidden bugfix? Maybe we can keep this one in its own commit and rebase-merge afterwards?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2987,6 +3020,87 @@ extern "C" rmw_ret_t rmw_get_node_names(
   return RMW_RET_BAD_ALLOC;
 }
 
+extern "C" rmw_ret_t rmw_get_node_names_with_security_contexts(
```
hidmic: @ivanpauno should this (and every other change in this patch) be conditionally compiled based on RMW version too?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2987,6 +3020,87 @@ extern "C" rmw_ret_t rmw_get_node_names(
...
+  auto node_impl = static_cast<CddsNode *>(node->data);
+  if (rmw_check_zero_rmw_string_array(node_names) != RMW_RET_OK ||
+    rmw_check_zero_rmw_string_array(node_namespaces) != RMW_RET_OK)
+  {
+    return RMW_RET_ERROR;
+  }
+
+  std::vector<std::tuple<std::string, std::string, std::string>> ns;
+  const auto re = std::regex(
+    "^name=([^;]*);namespace=([^;]*);securitycontext=([^;]*);", std::regex::extended);
```
hidmic: @ivanpauno nit: why not calling the constructor directly i.e. `std::regex re(...);`?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2987,6 +3020,87 @@ extern "C" rmw_ret_t rmw_get_node_names(
...
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  if (rcutils_string_array_init(node_names, ns.size(), &allocator) != RCUTILS_RET_OK ||
+    rcutils_string_array_init(node_namespaces, ns.size(), &allocator) != RCUTILS_RET_OK ||
+    rcutils_string_array_init(security_contexts, ns.size(), &allocator) != RCUTILS_RET_OK)
+  {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    goto fail_alloc;
+  }
+  size_t i;
+  i = 0;
```
hidmic: @ivanpauno nit: last two lines can be collapsed into one.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2987,6 +3020,87 @@ extern "C" rmw_ret_t rmw_get_node_names(
...
+extern "C" rmw_ret_t rmw_get_node_names_with_security_contexts(
+  const rmw_node_t * node,
+  rcutils_string_array_t * node_names,
+  rcutils_string_array_t * node_namespaces,
+  rcutils_string_array_t * security_contexts)
+{
+  RET_WRONG_IMPLID(node);
+  auto node_impl = static_cast<CddsNode *>(node->data);
+  if (rmw_check_zero_rmw_string_array(node_names) != RMW_RET_OK ||
+    rmw_check_zero_rmw_string_array(node_namespaces) != RMW_RET_OK)
```
hidmic: @ivanpauno should the same check be performed on `security_contexts` if not `nullptr`?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2987,6 +3020,87 @@ extern "C" rmw_ret_t rmw_get_node_names(
...
+  if (rcutils_string_array_init(node_names, ns.size(), &allocator) != RCUTILS_RET_OK ||
+    rcutils_string_array_init(node_namespaces, ns.size(), &allocator) != RCUTILS_RET_OK ||
+    rcutils_string_array_init(security_contexts, ns.size(), &allocator) != RCUTILS_RET_OK)
+  {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    goto fail_alloc;
+  }
+  size_t i;
+  i = 0;
+  for (auto && n : ns) {
```
hidmic: @ivanpauno why rvalue references?

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: Not exactly, as it wasn't a bug.
I'm adding `securitycontext=CONTEXT_NAME;` at the end of user data.
The changes are needed because of:
- I want the regex to match `name` and `namespace` also in old versions that doesn't have the `securitycontext` part at the end. So I dropped the `$` from the regex.
- To match correctly with `name=NODE_NAME;namespace=NAMESPACE;securitycontext=CONTEXT;`, I needed to replace the `(.*)` groups with `([^;]*)`. If not, the result would be `name="NODE_NAME"`, `namespace="NAMESPACE;securitycontext=CONTEXT"`.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: Thanks for the reminder, see https://github.com/ros2/rmw_cyclonedds/pull/106/commits/756fa27bc547a00ad36b80327ef825185a205d87.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: preexisting, I can address it if you mind strongly.

P.S.: As file as I understand, this is calling just the constructor.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: Same as above, it comes from `rmw_get_node_names`.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: It seems to me that you're looking to the wrong function:
https://github.com/ros2/rmw_cyclonedds/blob/ad4879663c2244ff88604292b239cc0400bf94df/rmw_cyclonedds_cpp/src/rmw_node.cpp#L3093-L3100

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: Copied from `rmw_get_node_names`.

P.S.: I can maybe do something better than copy-pasting ...

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: Yeah, copy elision will eventually do the job but it's odd. OK to leave as-is if pre-existing.

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: I think not. The snippet you refer to takes care of finalizing it if present, but I don't see a `rmw_check_zero_rmw_string_array` counterpart and [right below](https://github.com/ros2/rmw_cyclonedds/blob/ad4879663c2244ff88604292b239cc0400bf94df/rmw_cyclonedds_cpp/src/rmw_node.cpp#L3057) what is referred to by `security_context` is initialized without checking whether it was `nullptr` or not.

ivanpauno: See https://github.com/ros2/rmw_cyclonedds/pull/106/commits/1c072f236b94853a1042b468215625726d5c8011.

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: Don't feel like you have to leave non-idiomatic C++ bits untouched. Chances are that I wrote them, and as I'm more used to C than to C++ and have grown to dislike both, my C++ tends to come a bit weird sometimes.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: See https://github.com/ros2/rmw_cyclonedds/pull/106/commits/3c2f89c4414f5f9df88b3bc7d71197b0a069152d.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: There is a pretty fun reason: when done in the same line, the compiler fails saying that the above `goto` is skipping an initialization. In general, it's safe to skip the initialization of a trivially destructible object (AFAIU), but the compiler shows the warning anyways (and we transform that warning in an error, because of our compiler flags).

When split in two lines, the compiler doesn't show a warning.

P.S.: This kind of warning never happens in pure C code, it only applies when you use C++.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: Done in https://github.com/ros2/rmw_cyclonedds/pull/106/commits/3c2f89c4414f5f9df88b3bc7d71197b0a069152d.

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: Hmm interesting, odd but interesting.

wjwwood: Please define ‘i’ higher in the scope to avoid this instead. This what is done in all the other functions that use goto.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3188,6 +3127,37 @@ extern "C" rmw_ret_t rmw_get_node_names_with_security_contexts(
...
+#if RMW_VERSION_GTE(0, 8, 2)
+extern "C" rmw_ret_t rmw_get_node_names_with_security_contexts(
+  const rmw_node_t * node,
+  rcutils_string_array_t * node_names,
+  rcutils_string_array_t * node_namespaces,
+  rcutils_string_array_t * security_contexts)
+{
+  if (RMW_RET_OK != rmw_check_zero_rmw_string_array(security_contexts)) {
+    return RMW_RET_ERROR;
+  }
```
hidmic: @ivanpauno is it invalid to call `rmw_get_node_names_with_security_contexts()` with a null `security_contexts`? If not, I think this check should be relocated to the implementation function and conditionally executed based on `security_contexts` value.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3037,27 +3037,28 @@ static rmw_ret_t do_for_node_user_data(
...
-  std::vector<std::pair<std::string, std::string>> ns;
-  const auto re = std::regex("^name=([^;]*);namespace=([^;]*);", std::regex::extended);
+  std::vector<std::tuple<std::string, std::string, std::string>> ns;
   auto oper =
     [&ns, re](const dds_builtintopic_participant_t & sample, const char * ud) -> bool {
       std::cmatch cm;
       static_cast<void>(sample);
       if (std::regex_search(ud, cm, re)) {
-        ns.push_back(std::make_pair(std::string(cm[1]), std::string(cm[2])));
+        ns.push_back(std::make_tuple(std::string(cm[1]), std::string(cm[2]), std::string(cm[3])));
```
hidmic: @ivanpauno this bit is still assuming that there's a third group, regardless of `security_contexts` value.

Is having a `security_context` stanza in `ud` mandatory? If not, why not simply making it optional in the regex? Then, `rmw_get_node_names(...)` reduces to `rmw_get_node_names_with_security_contexts(..., nullptr)`.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: In `rmw_get_node_names_impl` `security_contexts` can be `NULL`.
In `rmw_get_node_names_with_security_contexts` `security_contexts` cannot be `NULL`.

I think the current logic is fine.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: This is wrong, thanks for detecting it!

ivanpauno: See https://github.com/ros2/rmw_cyclonedds/pull/106/commits/e09008b242ce747ee08ac0075ed8823d1253fa57.
Based on the [docs](https://en.cppreference.com/w/cpp/regex/match_results/operator_at), the previous code wasn't wrong. I actually ran tests, and it was working (forget about the detail that used the same regex both time and it wasn't going to match with old implementations).

If you have a better way to write the regex, please provide it.

ivanpauno:
I closed and reopened to trigger the new checks.

ivanpauno:
@rotu the check called `ubuntu-18,04, eloquent` seems to be using `master` and not `eloquent`.

Also, it's impossible that the checks using master would ever pass, if it's not possible to specify a branch in other package where we're doing simultaneous changes (e.g. `rmw`).
That's because we bump the API version is bumped once after the first API change, but not again after next release.

I'm not completely sure how to solve the problem, but I think we should only have actions for checking compatibility with past releases, and just rely on CI for master (which should be run manually by the user when the PR is ready).

ivanpauno:
>@rotu the check called ubuntu-18,04, eloquent seems to be using master and not eloquent.

Sounds like https://github.com/ros2/rmw_cyclonedds/pull/133 fixed this.
But checks checking `source` (which I understand is `master`), will continue failing.

rotu:
@ivanpauno I'm aware of the issue and it's due to my mistake in setting this up. Should be fixed by #133.

Yes, the checks ONLY check changes in this repo - the goal is to prevent accidental breakage. A change like this PR, involving a constellation of different commits, does tend to break things and presents a versioning challenge. Assuming that it would be too burdensome to preserve compatibility, the answer is to merge it with these checks failing.

rotu:
FYI: here's me blundering through setting up CI. Using "source" was a mistake. https://github.com/ros-tooling/action-ros-ci/issues/120

ivanpauno:
> Yes, the checks ONLY check changes in this repo - the goal is to prevent accidental breakage. A change like this PR, involving a constellation of different commits, does tend to break things and presents a versioning challenge. Assuming that it would be too burdensome to preserve compatibility, the answer is to merge it with these checks failing.

That sounds good to me.

I know that the actions using`source` will fail in this PR, but I would like to see if I'm maintaining compatibility with `dashing` and `eloquent` correctly.
Of course, I can manually run CI for `dashing` and `eloquent`, but that's what usually people don't do and end up breaking them.

Most of the PRs that have broken `dashing` and `eloquent` builds were merged with simultaneous changes in `rmw`, so ignoring the checks in those cases is ignoring the cases where we are most likely to break previous releases :smile: .

To avoid ignoring the result of the checks in those cases, I think we should either:
- Continue running `eloquent` and `dashing` checks if `source` check fails (i.e. not cancelling them).
- Delete the `source` checks, and rely on manually ran CI for `master`.

rotu:
>     * Continue running `eloquent` and `dashing` checks if `source` check fails (i.e. not cancelling them).

Agreed. That *should* be the behavior since #133 (the setting `fail-fast: false` in the yml)

> Delete the source checks, and rely on manually ran CI for master.

Manually run CI is still part of the picture. The actions in this repo only check things *up to* `rmw_cyclonedds`. It would certainly be nice to test things upstream as well with GitHub Actions and add more tests to this repo, testing both correctness and performance.

rotu:
To rerun this against latest CI workflow, I think you need to rebase onto master. I think if you fork this repo on Github, you can simply enable CI in the **Actions** tab of your fork and it will run on every push.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -636,23 +672,36 @@ static void node_gone_from_domain_locked(dds_domainid_t did)
...
+  const char * namespace_, size_t domain_id
+#if !RMW_VERSION_GTE(0, 8, 2)
+  , const rmw_node_security_options_t * security_options
+#endif
 #if RMW_VERSION_GTE(0, 8, 1)
   , bool localhost_only
 #endif
 )
 {
-  static_cast<void>(context);
```
hidmic, line 655: @ivanpauno isn't this up to the RMW version ?

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: I think @wjwwood still applies.

ivanpauno:
Note for reviewers: All checks are passing, except the ones using `master`. That's because there are simultaneous changes in others repos, and it's expected.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: The code is pre-existent, and I don't want to retrigger CI because of this. If you don't mind, I will ignore the comment.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: A warning didn't appear in the `Dashing` and `Eloquent` checks, but you're right.
Fixed in https://github.com/ros2/rmw_cyclonedds/pull/106/commits/d213440d429a2638de95ead7f2dee1bdc03f229c.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: Ok, finally fixed in https://github.com/ros2/rmw_cyclonedds/pull/106/commits/d213440d429a2638de95ead7f2dee1bdc03f229c.
I will rely on the `Eloquent` and `Dashing` checks here for validity, because the same code is exercised there.

eboasson:
@ivanpauno I've finally carved out the time to work on this, but before really commencing I wanted to make sure there won't be any duplication of work, which there might be if you happen to also be looking into it. (If you are, I'm totally happy to leave you to it, by the way 😁)

Also looking at the scope of the changes, and as I have mentioned before, I think this is the point where we should stop providing compatibility with dashing and eloquent in rmw_cyclonedds master. Hopefully we'll be able to merge #123 into master first: that would then be a single version that supports all features and works with Dashing, Eloquent and Foxy, and so would also be valuable for anyone who needs to integrate multiple versions of ROS2 in a single system.

It'll take me some time to get all the changes done and tested locally, so I'd say there's plenty of time for updating #123 to match the current state of master and merging it.

ivanpauno:
> @ivanpauno I've finally carved out the time to work on this, but before really commencing I wanted to make sure there won't be any duplication of work, which there might be if you happen to also be looking into it. (If you are, I'm totally happy to leave you to it, by the way grin)

Sounds good to me.
I'm not currently working on migrating the remaining rmw implementations to use one `Participant` per `Context`, neither I'm allocated to do it in a near future.

> Also looking at the scope of the changes, and as I have mentioned before, I think this is the point where we should stop providing compatibility with dashing and eloquent in rmw_cyclonedds master.

100% agreed, creating a new branch makes sense.

> Hopefully we'll be able to merge #123 into master first: that would then be a single version that supports all features and works with Dashing, Eloquent and Foxy, and so would also be valuable for anyone who needs to integrate multiple versions of ROS2 in a single system.

Sounds great to me!

----

If you need any reviews or if you have any questions, let me know, and I will be willing to help.
I will update https://github.com/ros2/design/pull/250 soon, so it better reflects what was done in `rmw_fastrtps`.

PR #115, opened 2020-03-17
Add rmw_*_event_init() functions

mm318:
Related to https://github.com/ros2/ros2/issues/822

This adds the `rmw_publisher_event_init()` and `rmw_subscription_event_init()` functions to `rmw_cyclonedds`, moved from `rmw`, so that these functions will be able to check for whether or not the given type of `rmw_event_type_t` is supported by the chosen middleware.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1872,7 +1872,7 @@ static const std::unordered_map<rmw_event_type_t, uint32_t> mask_map{
 
 static bool is_event_supported(const rmw_event_type_t event_t)
 {
-  return mask_map.count(event_t) > 0;
+  return mask_map.count(event_t) == 1;
 }
 
 static uint32_t get_status_kind_from_rmw(const rmw_event_type_t event_t)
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1856,6 +1856,71 @@ extern "C" rmw_ret_t rmw_return_loaned_message_from_subscription(
...
+static const std::unordered_map<rmw_event_type_t, uint32_t> mask_map{
+  {RMW_EVENT_LIVELINESS_CHANGED, DDS_LIVELINESS_CHANGED_STATUS},
+  {RMW_EVENT_REQUESTED_DEADLINE_MISSED, DDS_REQUESTED_DEADLINE_MISSED_STATUS},
+  {RMW_EVENT_LIVELINESS_LOST, DDS_LIVELINESS_LOST_STATUS},
+  {RMW_EVENT_OFFERED_DEADLINE_MISSED, DDS_OFFERED_DEADLINE_MISSED_STATUS},
+};
+
+static bool is_event_supported(const rmw_event_type_t event_t)
+{
+  return mask_map.count(event_t) > 0;
```
ivanpauno: I would use `==1`.

mm318: Done.

PR #123, opened 2020-03-20
Enable use of Cyclone DDS security features

SidFaber:
Add utility function to insert security settings to the cyclone QOS object used to create nodes. Include a utility to find security files and properly format their location to use with DDS.

Code testing included running the ROS2 tutorials with encryption enabled, error checking for missing certificates, testing bad signatures on governance and permissions files, invalid or mismatched domain IDs, and invalid encryption keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ Cyclone DDS is ready to use. It seeks to give the fastest, easiest, and most rob
 
 ## Building from source and contributing
 
-
 Note the `master` branch maintains compatibility with ROS releases Dashing and later, including the not-yet-released [*Foxy*](https://index.ros.org/doc/ros2/Releases/Release-Foxy-Fitzroy/).
 
 If building ROS2 from source ([ros2.repos](https://github.com/ros2/ros2/blob/master/ros2.repos)), you already have this package and Cyclone DDS:
--- a/README.md
+++ b/README.md
@@ -30,6 +30,5 @@ DDS directly instead of via the ROS2 abstraction.
 
 ## Known limitations
 
-Cyclone DDS doesn't yet implement the DDS Security standard, nor does it fully implement
-the Lifespan, Deadline and some of the Liveliness QoS modes.  Consequently these features
-of ROS2 are also not yet supported when using Cyclone DDS.
+Cyclone DDS doesn't yet fully implement the Lifespan, Deadline and some of the Liveliness QoS modes.
+Consequently these features of ROS2 are also not yet supported when using Cyclone DDS.
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -147,6 +147,18 @@ struct builtin_readers
   dds_entity_t rds[sizeof(builtin_topics) / sizeof(builtin_topics[0])];
 };
 
+#if RMW_SUPPORT_SECURITY
+struct dds_security_files_t
+{
+  char * identity_ca_cert = nullptr;
+  char * cert = nullptr;
+  char * key = nullptr;
+  char * permissions_ca_cert = nullptr;
+  char * governance_p7s = nullptr;
+  char * permissions_p7s = nullptr;
+};
+#endif
+
 struct CddsEntity
 {
   dds_entity_t enth;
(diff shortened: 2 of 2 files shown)


SidFaber:
This does not pay attention to the [`ENABLE_SECURITY` Cmake option](https://github.com/eclipse-cyclonedds/cyclonedds/blob/0768ad59edfacc203db8f047c2321472a2c243e8/src/CMakeLists.txt#L45)  because Cyclone does not generate an include file that uses the option as far as I can tell. 

@eboasson, is this something we want to add?

joespeed:
@rotu because rmw

README.md
```diff
@@ -30,6 +30,5 @@ DDS directly instead of via the ROS2 abstraction.
 
 ## Known limitations
 
-Cyclone DDS doesn't yet implement the DDS Security standard, nor does it fully implement
```
eboasson: Argh! The lifespan, deadline and liveliness QoS are fully supported nowadays, there aren't really any known limitations anymore! To be fixed in another PR, obviously.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -642,6 +644,80 @@ static std::string get_node_user_data(const char * node_name, const char * node_
...
+    qos, "dds.sec.access.permissions_ca", "permissions_ca.cert.pem",
+    security_options);
+  store_security_filepath_in_qos(
+    qos, "dds.sec.access.governance", "governance.p7s",
+    security_options);
+  store_security_filepath_in_qos(
+    qos, "dds.sec.access.permissions", "permissions.p7s",
+    security_options);
+
+  dds_qset_prop(qos, "dds.sec.auth.library.path", "libdds_security_auth.so");
```
eboasson: The "lib" and ".so" affixes will cause trouble on macOS and Windows. If the specified file does not exist, Cyclone also tries the names extended in the platform's native way (so lib/.dylib for macOS and .dll for Windows). So just "dds_security_auth" seems better (and analogously for the other libraries).

mikaelarguedas: this looks platfrom specific, maybe there is support for the library basename and cyclone then modifies and look for specific names based on the platform (ala class_loader https://github.com/ros/class_loader/blob/11982ab3efc390608f10f2db1e97edf4037afec5/src/class_loader.cpp#L51-L73)

SidFaber:
> ... My preference is to define a macro `DDS_HAS_QOS_PROPERTY_LIST` in Cyclone's public header files if the property list QoS is implemented, and test for that in the RMW layer.
 
@eboasson do you plan to add this to the security branch of Cyclone?  If so I'll update this with some code that should handle the items you pointed out.  Nice catch on preventing a silent fail-open scenario!

eboasson:
> > ... My preference is to define a macro `DDS_HAS_QOS_PROPERTY_LIST` in Cyclone's public header files if the property list QoS is implemented, and test for that in the RMW layer.
> 
> @eboasson do you plan to add this to the security branch of Cyclone? If so I'll update this with some code that should handle the items you pointed out. Nice catch on preventing a silent fail-open scenario!

Yep, working on it: https://github.com/eclipse-cyclonedds/cyclonedds/pull/448

SidFaber:
> Yep, working on it: [eclipse-cyclonedds/cyclonedds#448](https://github.com/eclipse-cyclonedds/cyclonedds/pull/448)

@eboasson, I have updates here ready to land once this is ready, thanks for the improvements

eboasson:
@SidFaber FIY: I just merged #448

rmw_cyclonedds_cpp/src/rmw_node.cpp
SidFaber: Updated in bca0852

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -26,9 +26,11 @@
 #include <utility>
 #include <regex>
 
+#include "rcutils/filesystem.h"
 #include "rcutils/get_env.h"
 #include "rcutils/logging_macros.h"
 #include "rcutils/strdup.h"
+#include "rcutils/format_string.h"
```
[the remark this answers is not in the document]
SidFaber: Updated in bca0852

rmw_cyclonedds_cpp/src/rmw_node.cpp
SidFaber: Fixed with bca0852

SidFaber:
@eboasson, @mikaelarguedas, thanks for your comments, they should be all addressed in bca0852. Important to note that now node creation fails when security is requested (ROS_SECURITY_ENABLE=true); this change the current fail-open behavior to fail-closed which should be desired.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -688,8 +789,17 @@ extern "C" rmw_node_t * rmw_create_node(
 #endif
 
   dds_qos_t * qos = dds_create_qos();
+  RCUTILS_CHECK_FOR_NULL_WITH_MSG(
+    security_options, "rmw_create_node: Unable to create qos", return nullptr);
```
eboasson: I suspect this was meant to check ``qos`` rather than ``security_options``; and if bailing out, a domain will have been instantiated/refcounted already by ``check_create_domain_locked`` and, as noted, its work must be undone using ``node_gone_from_domain_locked`` (see a few lines below).

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -688,8 +789,17 @@ extern "C" rmw_node_t * rmw_create_node(
...
   std::string user_data = get_node_user_data(name, namespace_);
   dds_qset_userdata(qos, user_data.c_str(), user_data.size());
+
+  if (security_options->enforce_security) {
+    if (configure_qos_for_security(qos, security_options) != RMW_RET_OK) {
+      return nullptr;
```
eboasson: This, too, can leak the domain reference, and moreover leaks ``qos``. Isn't C just a lovely language, always being on the lookout for ways to get you? 😄

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -688,8 +789,17 @@ extern "C" rmw_node_t * rmw_create_node(
 #endif
 
   dds_qos_t * qos = dds_create_qos();
+  RCUTILS_CHECK_FOR_NULL_WITH_MSG(
+    security_options, "rmw_create_node: Unable to create qos", return nullptr);
   std::string user_data = get_node_user_data(name, namespace_);
   dds_qset_userdata(qos, user_data.c_str(), user_data.size());
+
+  if (security_options->enforce_security) {
```
mikaelarguedas: This logic is incomplete, it sets the security QoS *only if* `enforce_security` is set.

The expected behavior is to:
If the security files can be found: setup the plugins
If they cannot be found:
- if enforce_security is true -> exit with error message
- else -> create a participant without the security plugins instanciated

This is most likely the cause of the failing tests at https://github.com/ros2/system_tests/pull/408

rmw_cyclonedds_cpp/src/rmw_node.cpp
kyrofa: > Isn't C just a lovely language, always being on the lookout for ways to get you?

Regarding that: is there a reason you're not making more substantial use of C++ features, here?

eboasson: Not really, the RMW layer is intended to be in C++. Part of the reason that it is so C-like is that I am much more experienced with C than with C++ and so whenever I try to write C++ it still ends up looking like C. There is also the fact that Cyclone is in C, that this is therefore relying on a C API, and that some of the niceties of C++ (like destructors) are simply not available on these objects.

There is an implementation of a standard C++11 API of DDS waiting for approval by the Eclipse Foundation's IP review process. That interface covers all the standard features but not all the additional tricks. At some point it might be worth rewriting the RMW implementation to use that API, it's only about a handful of lines of code anyway.

kyrofa: > Part of the reason that it is so C-like is that I am much more experienced with C than with C++ and so whenever I try to write C++ it still ends up looking like C.

Fair enough, that's a perfectly valid reason.

> There is also the fact that Cyclone is in C, that this is therefore relying on a C API, and that some of the niceties of C++ (like destructors) are simply not available on these objects.

In case you're unaware, shared/unique ptrs can be handed custom deleters that might help with that particular issue.

eboasson: > In case you're unaware, shared/unique ptrs can be handed custom deleters that might help with that particular issue.

That sounds like something worth reading up on, thanks!

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -688,8 +770,18 @@ extern "C" rmw_node_t * rmw_create_node(
 #endif
 
   dds_qos_t * qos = dds_create_qos();
+  if (qos == nullptr) {
```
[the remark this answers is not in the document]
SidFaber: Have to call a cleanup routine here before exiting

rmw_cyclonedds_cpp/src/rmw_node.cpp
SidFaber: @mikaelarguedas this is how the other middleware behaves, but help me understand if this is the behavior we truly want.

The rcl sets security_options->enforce_security to either RMW_SECURITY_ENFORCEMENT_PERMISSIVE or RMW_SECURITY_ENFORCEMENT_ENFORCE. It's always set to RMW_SECURITY_ENFORCEMENT_PERMISSIVE when ROS_SECURITY_ENABLE is anything other than "true"; that is why it's used to completely bypass security qos here.

With what you propose security will be enabled anytime ROS_SECURITY_ROOT_DIRECTORY exists regardless of ROS_SECURITY_ENABLE and ROS_SECURITY_STRATEGY.  The [rcl only checks for the directory](https://github.com/ros2/rcl/blob/73948da4c50b11d5d133e068b804ce7c2b4f9cb6/rcl/src/rcl/node.c#L289), it (rightfully) does not ensure all six security files exist and are properly formatted. So node creation would fail any time the middleware cannot start with files in the security directory, even when ROS_SECURITY_ENABLE is set to false. 

Similarly ROS_SECURITY_ENABLE=false would not disable security for testing. Instead security would have to be disabled by changing ROS_SECURITY_ROOT_DIRECTORY or moving the security directory on disk.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -688,8 +787,22 @@ extern "C" rmw_node_t * rmw_create_node(
...
+    RCUTILS_LOG_ERROR_NAMED("rmw_cyclonedds_cpp", "rmw_create_node: Unable to create qos");
+    node_gone_from_domain_locked(did);
+    return nullptr;
+  }
   std::string user_data = get_node_user_data(name, namespace_);
   dds_qset_userdata(qos, user_data.c_str(), user_data.size());
+
+  if (security_options->enforce_security) {
```
SidFaber: This needs to use the rmw_security_enforcement_policy_t enum:

```suggestion
  if (security_options->enforce_security==RMW_SECURITY_ENFORCEMENT_ENFORCE) {
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
mikaelarguedas: If don't think the described behavior matches how the other middlewares behave, or the desired behavior.
Overall the expected behavior is the one described at: https://design.ros2.org/articles/ros2_dds_security.html

Putting it in a table for ease of read:

| ROS_SECURITY_ENABLE  | ROS_SECURITY_STRATEGY | valid set of security files located under ROS_SECURE_ROOT_DIRECTORY | Expected outcome |
| ----- | ---------- | --- | ----------------------------------------------------------------- |
| false | *          | *   | No security plugins instanciated, creating non-secure participant |
| true  | Enforce    | Yes | creates secure participant                                        |
| true  | Enforce    | No  | fails to create participant                                       |
| true  | Permissive | Yes | creates secure participant                                        |
| true  | Permissive | No  | creates unsecure participant                                      |

Anything not matching that behavior is a bug IMO. 

I agree that the line you are referring to is misleading https://github.com/ros2/rcl/blame/73948da4c50b11d5d133e068b804ce7c2b4f9cb6/rcl/src/rcl/node.c#L316-L317 and in practice serve no purpose and at first glance could be removed completely.
The one that matters in our case is the else branch of that if: https://github.com/ros2/rcl/blame/73948da4c50b11d5d133e068b804ce7c2b4f9cb6/rcl/src/rcl/node.c#L318-L323
If ROS_SECURITY_ENABLE is false: node_security_options.security_root_path is **not** set and this is the variable that tells you if you should try to instantiate the security plugins or not.

Mimicking the logic of the other RMWs could be a good hint of how to implement it here: https://github.com/ros2/rmw_fastrtps/blob/bd5507b61c7285d6083e514ccd49f00b2945baeb/rmw_fastrtps_shared_cpp/src/rmw_node.cpp#L292

rmw_cyclonedds_cpp/src/rmw_node.cpp
rotu: @eboasson better yet, define a destructors for your classes. Then you don't need to pass a custom deleter to unique_ptr/shared_ptr

kyrofa: @rotu that's not always possible:

> [...] this is therefore relying on a C API, and that some of the niceties of C++ (like destructors) are simply not available on these objects.

rotu: I find it's easier to maintain a wrapper class than a deleter, but YMMV. Where there's a matched initializer + deleter pair, I find it easiest to make them into a constructor + destructor. With a factory method + deleter pair, i prefer to use a class method + protected constructor + destructor.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -59,6 +61,7 @@
 #include "namespace_prefix.hpp"
 
 #include "dds/dds.h"
+#include "dds/ddsc/dds_public_qos.h"
```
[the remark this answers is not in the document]
SidFaber: Removed in 1ca2269

rmw_cyclonedds_cpp/src/rmw_node.cpp
SidFaber: Fixed with 1ca2269, thanks for catching!

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -666,7 +742,13 @@ extern "C" rmw_node_t * rmw_create_node(
   static_cast<void>(domain_id);
   const dds_domainid_t did = DDS_DOMAIN_DEFAULT;
 #endif
-  (void) security_options;
+
+  if (security_options == nullptr) {
```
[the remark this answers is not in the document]
SidFaber: Updated with 1ca2269

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -642,6 +652,95 @@ static std::string get_node_user_data(const char * node_name, const char * node_
...
+    allocator.deallocate(security_file_path, allocator.state);
+  }
+}
+#endif  /* RMW_SUPPORT_SECURITY */
+
+/*  Set all the qos properties needed to enable DDS security  */
+rmw_ret_t configure_qos_for_security(
+  dds_qos_t * qos, const rmw_node_security_options_t * security_options)
+{
+
```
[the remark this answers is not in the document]
SidFaber: Removed in 1ca2269

rmw_cyclonedds_cpp/src/rmw_node.cpp
SidFaber: @eboasson, this should all be covered with 1ca2269.

rmw_cyclonedds_cpp/src/rmw_node.cpp
SidFaber: @mikaelarguedas, thanks for the discussion. It was simple to switch the logic here, but you'll also see changes to `configure_qos_for_security`: all files are checked before any properties are added to qos to enable cyclone dds security.

This is now passing all everything in the `test_security`, that's been a great help and I'd like to see those test routines continue to mature!

mikaelarguedas: That's awesome.

> all files are checked before any properties are added to qos to enable cyclone dds security.

Great improvement, we should consider modifying the other rmw implementations to make sure they do the same

> This is now passing all everything in the test_security, that's been a great help and I'd like to see those test routines continue to mature!

:tada: 


I didnt test the new version as it's conflicting since https://github.com/ros2/rmw_cyclonedds/pull/106

eboasson: @SidFaber I gave it a go (with Dashing to verify it supports that, too) and I noticed that it warns
```
[INFO] [rmw_cyclonedds_cpp]: rmw_create_node: Unable to configure security
```
whenever it starts when security either isn't supported or couldn't be configured. That is, when security is:
1. disabled, regardless of whether security support is included in the build;
2. enabled in permissive mode, when security support not included in the build;
3. enabled in permissive mode, when security support is included in the build but something is awry with the configuration.

Case (1) seems wrong (there is no expectation of security so also no expectation of getting warnings about it) and case (3) seems like a good thing to me. I am not quite sure what to make of case (2), I'd say it should be a hard fail but perhaps that's not the intent behind permissive mode (I don't quite get that mode anyway). @mikaelarguedas could you comment on that?

Just for completeness: if security is enabled in enforced mode, it only starts when the security configuration is valid and the implementation supports it.

Apart from the warning not being appropriate in all cases, it looks good to me!

mikaelarguedas: > Case (1) seems wrong

:+1:

> case (3) seems like a good thing to me

:+1:

> I am not quite sure what to make of case (2), I'd say it should be a hard fail but perhaps that's not the intent behind permissive mode (I don't quite get that mode anyway)

Originally the "Permissive" mode is to allow someone to launch in a single environment a system with some parts using security and some parts without.
I don't think it's been used much in practice.
In the case of 2, the rmw implementation itself doesnt support security (because not included in the build) so there's no chance of having any part of the system using security. So IMO it should hard fail and provide a clearer error message as of why.

An example is such error message in rmw_fastrtps_cpp https://github.com/ros2/rmw_fastrtps/blob/48b403a211b269ab383e6adfcf89d1b4333b2c39/rmw_fastrtps_shared_cpp/src/participant.cpp#L238-L242

SidFaber: How about I just give an info message "Security is enabled" when the cyclone QOS object gets set up properly? It will normally follow the "found security directory" message that comes from rcl. I feel some messaging is important since the rcl message implies that security is enabled which may not be the case.

mikaelarguedas: As I find the rcl message by itself pretty noisy when launching systems with many nodes, I'd rather not add an info message *always*. I would prefer a warning when users need to know something is not what they requested (case (3)), and nothing (or debug message) otherwise.

SidFaber: @mikaelarguedas, I just removed the info message: there are some tests where you'll get an rcl "found security" message and create a plain text node which is similar to how other middleware behaves. Although not ideal, this is something that would be better solved with logging.

SidFaber:
This has been updated for the recent context changes.

@eboasson, I followed the existing style to conditionally include based on rmw versions. fastrtps has diverged from this code now as their latest update does more with contexts, but everything here seems to be working fine.

@mikaelarguedas, this passed the current `test_security` after changing the makefile slightly to also test `rmw_cyclonedds_cpp`. 

Thanks for your help with this one!

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -690,6 +711,131 @@ static std::string get_node_user_data(
...
+  }
+  finalize_security_file_URIs(dds_security_files, allocator);
+  return ret;
+#else
+  (void) qos;
+  (void) security_options;
+  RMW_SET_ERROR_MSG(
+    "Security was requested but the Cyclone DDS being used does not have security "
+    "support enabled. Recompile Cyclone DDS with the '-DENABLE_SECURITY=ON' "
+    "CMake option");
```
ivanpauno, line 834: I'm always getting this error message if I build a `ros2` workspace without passing `-DENABLE_SECURITY=ON` and run any demo (using cyclone)
I think `rmw_cyclonedds_cpp` should be able to automatically detect if security is available or not at build time.

kyrofa: You're right, this error message should probably only be set if security was specifically requested by the user, but that's only checked when this function is called, which is too late. @SidFaber what do you think?

ivanpauno: See https://github.com/ros2/rmw_cyclonedds/pull/172 for a possible fix.

SidFaber: This should only happen when ROS_SECURITY_STRATEGY=Enforce and ROS_SECURITY_ENABLE=True. The full interplay of environment variables, RCL and RMW_CYCLONE is outlined [here](https://docs.google.com/spreadsheets/d/15Wf2MjGRxTUnpQl971zV75V2X4JhvA4tf3F1qzgqPZM/edit#gid=0), the message should only show in lines 19, 20 and 22.

@ivanpauno, can you confirm that is or is not what you're seeing? This should work with both the current and the security branches of rmw_cyclonedds_cpp. Thanks!

ivanpauno: > @ivanpauno, can you confirm that is or is not what you're seeing? This should work with both the current and the security branches of rmw_cyclonedds_cpp. Thanks!

I'm seeing this error without setting any security related environment variable.

SidFaber: https://github.com/ros2/rmw_cyclonedds/pull/175 should fix this up, thanks for pointing this out.

PR #125, opened 2020-03-24
Support for ON_REQUESTED_INCOMPATIBLE_QOS and ON_OFFERED_INCOMPATIBLE_QOS events

mm318:
Related to [this](https://github.com/ros2/ros2/issues/822) feature request. The design and implementation details can also be found there.

Handling events of type `REQUESTED_INCOMPATIBLE_QOS_STATUS` and `OFFERED_INCOMPATIBLE_QOS_STATUS`

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -33,14 +33,15 @@
 #include "rmw/allocators.h"
 #include "rmw/convert_rcutils_ret_to_rmw_ret.h"
 #include "rmw/error_handling.h"
-#include "rmw/names_and_types.h"
+#include "rmw/event.h"
+#include "rmw/get_node_info_and_types.h"
 #include "rmw/get_service_names_and_types.h"
 #include "rmw/get_topic_names_and_types.h"
-#include "rmw/get_node_info_and_types.h"
-#include "rmw/event.h"
-#include "rmw/validate_node_name.h"
+#include "rmw/incompatible_qos_events_statuses.h"
+#include "rmw/names_and_types.h"
 #include "rmw/rmw.h"
 #include "rmw/sanity_checks.h"
+#include "rmw/validate_node_name.h"
 
 #include "Serialization.hpp"
 #include "rmw/impl/cpp/macros.hpp"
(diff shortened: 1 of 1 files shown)


dennis-adlink:
@mm318 thanks for adding support for these events! The changes look good to me, I don't see any issues.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -33,14 +33,15 @@
...
 #include "rmw/get_service_names_and_types.h"
 #include "rmw/get_topic_names_and_types.h"
-#include "rmw/get_node_info_and_types.h"
-#include "rmw/event.h"
-#include "rmw/validate_node_name.h"
+#include "rmw/incompatible_qos_events_statuses.h"
+#include "rmw/names_and_types.h"
 #include "rmw/rmw.h"
 #include "rmw/sanity_checks.h"
+#include "rmw/validate_node_name.h"
```
ivanpauno: nit: The reordering is unrelated to the PR, though I don't mind strongly.

mm318:
Thanks, @eboasson. I have added `#if RMW_VERSION_GTE(0,8,2)` guards for the new code.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -33,14 +33,15 @@
...
 #include "rmw/error_handling.h"
-#include "rmw/names_and_types.h"
+#include "rmw/event.h"
+#include "rmw/get_node_info_and_types.h"
 #include "rmw/get_service_names_and_types.h"
 #include "rmw/get_topic_names_and_types.h"
-#include "rmw/get_node_info_and_types.h"
-#include "rmw/event.h"
-#include "rmw/validate_node_name.h"
+#include "rmw/incompatible_qos_events_statuses.h"
```
ivanpauno: @mm318 this should have been conditionally included with a `#if RMW_VERSION_GTE(0, 8, 2)` guard (the header didn't exist before).
Can you open a follow-up?

ivanpauno: Fixed in https://github.com/ros2/rmw_cyclonedds/pull/127.

mm318: Sorry, I missed this!

eboasson: @mm318 it happens to us all ... and in this case had a rather positive side-effect: @rotu decided to go set-up GitHub actions to do builds on PRs against dashing and eloquent. I don't know if he has already verified that a build against dashing indeed fails when a >=0.8.2 check is removed, but once we know we can trust it, the risk will have been reduced significantly.

PR #127, opened 2020-03-29
Include incompatible_qos_events_statuses.h only on Foxy (and newer)

eboasson:
The absence of compile-time guards breaks Dashing & Eloquent builds since #125 got merged.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -37,7 +37,6 @@
 #include "rmw/get_node_info_and_types.h"
 #include "rmw/get_service_names_and_types.h"
 #include "rmw/get_topic_names_and_types.h"
-#include "rmw/incompatible_qos_events_statuses.h"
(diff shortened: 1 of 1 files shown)


mm318:
Sorry, I missed the `#if` guard!

Should we move to using multiple branches with backporting instead of using lots of `#if` guards?

eboasson:
> Sorry, I missed the `#if` guard!
> 
> Should we move to using multiple branches with backporting instead of using lots of `#if` guards?

I rather like getting all the fixes and feature improvements automatically available to dashing and eloquent without having to do any backporting. For example, @SidFaber and @kyrofa have a PR as good as ready for merging that adds SROS support. Take that PR and the security branch of cyclone — and suddenly Dashing and Cyclone will have security support, no further work needed.

However, we are at the stage that the Cyclone RMW layer is pretty much "done", and so the benefit of having a single source is not as great anymore. As this is not the usual model in the ROS community, the model should be one that most contributing to this repository are comfortable with, and not necessarily the one that I happen to like myself.

PR #131, opened 2020-03-30
Freshen up the readme

rotu:


eboasson:
You might as well delete the "known limitations" section, although technically the security stuff isn't available in a default build yet.

rotu:
@eboasson Done. Other suggestions?

rotu:
You have a great point there. Changed “targets” to “maintains compatibility with”. Yes we should hone the release model to hone expectations about “what should users use?” I expect a “dashing-release” and “eloquent-release” branches which track the binary packages would be more self-descriptive than the current “eloquent”.

dirk-thomas:
> Eloquent is on a separate branch for no other reason than that that branch happened to get created when Eloquent got released.

While this is currently the case it doesn't have to stay like that. I would highly recommend that for each ROS distro there is one and only one answer to which branch should be used for it. That could be separate branches for ROS distros or the same branch or a mixture.

But saying that for e.g. Eloquent there is a branch but also master and both should work makes it a) difficult for users to choose since it is nuclear which one is used for making releases and b) requires the maintainers to keep those branches in sync.

PR #137, opened 2020-04-03
Add debug tips to README.md

rotu:


rotu:
@eboasson, any additional suggestions would be nice

eboasson:
> @eboasson, any additional suggestions would be nice

Based on an "accident report" I received the other day and the simple workaround for it: if there are occasional extreme spikes in latency (like seconds, or even minutes) when sending large samples, make sure to set Internal/MinimumSocketReceiveBufferSize to a value a fair bit larger than the maximum sample size.

The typical cause of the latency spikes is a very eager retransmit behaviour in combination with those really large samples as well as many of the subsequent requests for retransmission of lost fragments overrunning the UDP receive buffer. Making the latter large enough to easily hold such a very large sample dramatically reduces this problem.

I really need to do something about that ...

rotu:
@eboasson Oof!!! What is that latency of? Is that in ROS, Cyclone, or the network? And how would/did you diagnose it?

eboasson:
That latency is from sending a sample to receiving it on the other end. Assuming my understanding is correct and it is not a lucky accident that changing this setting did the trick, it has nothing to do with ROS or bad networks: it is basically a retransmit storm caused by consistently overruning the receiver. Make it a little less eager to retransmit and it will do a lot better without tweaking.

I only got a description of the application-visible symptoms — a clear and detailed one including message sizes (~10MB) — and confirmation that changing this setting makes it work reliably instead of working often but going berserk every now and then. Diagnosis is mostly a bit of experience :) And the consequent "rule" is simply: make sure your samples are not much larger than the UDP socket receive buffers, or long latency spikes can happen. (Until the retransmit behaviour gets some TLC, that is.

There haven't been all that many >1MB samples in DDS systems so far, and so it is a "new" problem that started surfacing with sending images and point clouds. While I don't understand the engineering rationale for the large samples in, e.g.,  point clouds, there's no excuse for not handling them well.

rotu:
I’m aiming the readme at those *don’t* have that level of experience, and I don’t want it to be a bug bulletin or troubleshooting steps (though we could have those too). can the user identify send-to-receive latency? How can the user detect a “retransmit storm” or at least capture the details so another user can look at it and say “yes that’s what’s happening” or “no, it’s something else”? Cyclone logging? Wireshark? Something else?

README.md
```diff
@@ -23,6 +23,22 @@ Cyclone DDS is ready to use. It seeks to give the fastest, easiest, and most rob
...
+So Cyclone isn't playing nice or not giving you the performance you had hoped for? That's not good... Please [file an issue against this repository](https://github.com/ros2/rmw_cyclonedds/issues/new)!
+Here are some ways to generate additional debugging info that can help identify the problem faster.
+
+Set update the config to create richer debugging output.
+
+To see the output live:
+`export CYCLONEDDS_URI=‘<Tracing><Verbosity>trace</><Out>stderr</></>’`
+
+To put log files to /var/log/:
+`export CYCLONEDDS_URI=‘<Tracing><Verbosity>trace</><Out>/var/log/CycloneDDS/${CYCLONEDDS_PID}.log</></>’`
```
eboasson, line 37: This assumes /var/log/CycloneDDS exists (it won't make the directory if it doesn't exist). For that reason, perhaps it would be better to suggest /var/log/cyclonedds.${CYCLONEDDS_PID}.log instead?

rotu:
> The trouble with the advice to turn on the traces (or capture a wireshark thing) is that it is not always so easy to interpret. The general comment to create an issue is a good catch-all in any case.

Yeah, I know, I want to make sure issues are captured, and I'm not sure exactly what DIY solutions are out there.

> I think it would be good to have a list of known symptoms and what to do to debug as well. For the most part those are misbehaviours that need to be fixed, but until the fix is available, it would be good to help people a bit.

Agreed. Please feel free to add.

>     * Excessive or erratic latencies with large samples (say > 250kB) are often caused by overruning the UDP socket receive buffer and increasing the size of the UDP socket receive buffer helps. That is in part a kernel configuration issue (Ubuntu 18 by default limits these buffers to about 250kB; see e.g. https://cuwb.io/2018/02/24/linux-udp-buffers.html), and in part a Cyclone configuration issue (it defaults to asking for 1MB and accepting whatever it gets, but you can use Internal/MinimumSocketReceiveBufferSize to request more.

That recommends making it 100MB which seems unreasonable. I added some recommendations to make it 8 MB. Any way to make Cyclone take advantage of that by default?

>     * If nodes aren't discovering each other, it can help to fall back to a pure unicast configuration with UDP datagrams no larger than a single packet in order to investigate.

Added.

>     * Simple connectivity, throughput and latency tests can also be done with Cyclone's ddsperf tool (it is in install/cyclonedds/bin)
>       that is just about the simplest configuration as far as networking

Awesome. Added.

> Edit: Giuseppe Peano must be rolling in his grave that I think 2 equals 3 ...

There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors.

rotu:
I don't quite understand the other suggestions you're making with MaxMessageSize, AllowMulticast, ParticipantIndex, etc. If the user has to do this, I think we want to find the underlying issue instead of having them work around it.

eboasson:
> I don't quite understand the other suggestions you're making with MaxMessageSize, AllowMulticast, ParticipantIndex, etc. If the user has to do this, I think we want to find the underlying issue instead of having them work around it.

Oh yes, that's definitely something that should work well automatically. It is a step in the procedure when it just won't work, arguably anyone who cares enough to try fixing things in such a situation would probably care enough to ask a question.

README.md
```diff
@@ -7,23 +7,55 @@ Cyclone DDS is ready to use. It seeks to give the fastest, easiest, and most rob
...
+   ```
 
-2) Set env variable and run ROS2 apps as usual:
+2. Set env variable and run ROS2 apps as usual:
 
-  ```export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp```
+   ```export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp```
 
-3) Confirm RMW: In Eloquent, to confirm which RMW you're using:
+3. Confirm RMW: In Eloquent, to confirm which RMW you're using:
```
eboasson, line 22: shouldn't that be "Eloquent and later"?

README.md
```diff
@@ -7,23 +7,55 @@ Cyclone DDS is ready to use. It seeks to give the fastest, easiest, and most rob
...
 
    ```ros2 doctor --report```
 
 
+## Performance recommendations
+
+With large samples (100s of kilobytes), excessive latency can be caused by running out of space in the OS-level receive buffer. For this reason, on Linux, we recommend increasing the buffer size:
+* Temporarily (until reboot): `sudo sysctl -w net.core.rmem_max=8388608`
+* Permanently: `echo "net.core.rmem_max=8388608\n" | sudo tee /etc/sysctl.d/60-cyclonedds.conf`
```
eboasson, line 31: The kernel has a default and a maximum. This only raises the maximum, which means cyclone will still get the default, then check if it is < 1MB and if so ask for 1MB.

If you want to actually get more than 1MB, you either have to raise the kernel default (in which cyclone will observe it is > 1MB and leave it alone) or configure cyclone to ask for more. That's what Internal/MinimumSocketReceiveBufferSize size is for.

rotu: That's a bit confusing.... I have an idea.

rotu: My Idea: https://github.com/eclipse-cyclonedds/cyclonedds/pull/489

PR #145, opened 2020-04-09
Switch to one participant per context model

eboasson:
@ivanpauno this seems to work and it would be great if you could find the time to look it over. I have only tested it locally using rcl/test and test_communication.

* Secure ROS can't find the "sros2"-generated key stores, but that seems to be the case with Fast-RTPS as well.
* I've really relied on the Fast-RTPS RMW layer, so various parts were copied nearly verbatim (like demangling, type support and introspection functions), and some parts (e.g., context setup, discovery-related extensions to creating/destroying nodes, publishers, subscriptions) borrow heavily from the work done there.
* I was surprised that local readers/writers require a add_entity call in addition to the associate call, but it is the only logical conclusion after observing that skipping the local entities in my discovery thread it makes a ton of tests fail.
* It does seem to me that it would make sense to define the (de)mangling of topic and type names in ``rmw_dds_common``.
* I suspect some of the tests have become timing-dependent now because the local readers/writers get added to the graph cache asynchronously at the moment. I could of course do the local add_reader/add_writer calls synchronously, but it would be good to know what is expected.

I think there'll be at least one round of cleaning up to come 😄 and I'm also not quite sure I've done the right thing with the unique_ptrs (not that much of a C++ man ... @rotu will hopefully be able to advise there).


ivanpauno:
> Secure ROS can't find the "sros2"-generated key stores, but that seems to be the case with Fast-RTPS as well.

There's a problem in `Focal` with the version of SSL, but I tried in both Bionic and Windows and it's working. You can check also Windows nightlies (from some days ago, as it's broken for other reasons right now).

> I was surprised that local readers/writers require a add_entity call in addition to the associate call, but it is the only logical conclusion after observing that skipping the local entities in my discovery thread it makes a ton of tests fail.

That's needed, so a reader/writer can be matched with a `Node`. If not, you can only match it with the `Participant` that created it.

> It does seem to me that it would make sense to define the (de)mangling of topic and type names in rmw_dds_common.

Yeah, I agree. I was thinking to move all that logic there, but I didn't have the time to do it.

ivanpauno:
> I suspect some of the tests have become timing-dependent now because the local readers/writers get added to the graph cache asynchronously at the moment. I could of course do the local add_reader/add_writer calls synchronously, but it would be good to know what is expected.

What can happen is that a reader is shown when listing subscriptions, but not shown when listing the subscription that a node created.
I think it's fine if that continues being asynchronous.

This could be probably avoided if the information is somehow communicated through reader/writer userdata (see https://github.com/ros2/design/pull/250#discussion_r391160903).
The main reason I didn't choose that path, is that there were not support in all DDS vendors that ROS2 is using.

ivanpauno:
@eboasson this was fast! I will do a review ASAP.

eboasson:
@rotu just wanted to remark that it conflicts with your #119 but only in a trivial way: a rename of one local variable to eliminate a warning that it shadowed another one. I purposely left the code alone where it had nothing to do with the change in mapping.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -436,207 +431,233 @@ extern "C" rmw_ret_t rmw_init_options_fini(rmw_init_options_t * init_options)
...
+  /* deleting ppant will delete waitset as well, so there is no real need to delete
+     the waitset here on error, but it is more hygienic */
+  if ((ws = dds_create_waitset(impl->ppant)) < 0) {
+    RCUTILS_SAFE_FWRITE_TO_STDERR(
+      "ros discovery info listener thread: failed to create waitset, will shutdown ...\n");
+    return;
+  }
+  /* I suppose I could attach lambda functions one way or another, which would
+     definitely be more elegant, but this avoids having to deal with the C++
+     freakishness that is involved and works, too. */
```
ivanpauno, line 556: I think this is fine, using lambdas would probably be more confusing.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -836,148 +823,282 @@ rmw_ret_t configure_qos_for_security(
...
 
-#if SUPPORT_LOCALHOST
   /* Take domains_lock and hold it until after the participant creation succeeded or
      failed: otherwise there is a race with rmw_destroy_node deleting the last participant
      and tearing down the domain for versions of Cyclone that implement the original
      version of dds_create_domain that doesn't return a handle.  */
-  std::lock_guard<std::mutex> lock(gcdds.domains_lock);
-  if (!check_create_domain_locked(did, localhost_only)) {
-    return nullptr;
+  if (!check_create_domain(domain_id, options->localhost_only)) {
```
ivanpauno, line 865: In `rmw_fastrtps`, a `Participant` is created when the first `Node` is created within a `Context`.
This is also a valid option, and shouldn't create cross-vendor problems.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -836,148 +823,282 @@ rmw_ret_t configure_qos_for_security(
...
-    if (security_options->enforce_security == RMW_SECURITY_ENFORCEMENT_ENFORCE) {
-      dds_delete_qos(qos);
-      node_gone_from_domain_locked(did);
-      return nullptr;
+  std::unique_ptr<dds_qos_t, std::function<void(dds_qos_t *)>>
+  ppant_qos(dds_create_qos(), &dds_delete_qos);
+  if (ppant_qos == nullptr) {
+    return RMW_RET_BAD_ALLOC;
+  }
+  std::string user_data = std::string("securitycontext=") + std::string(
```
ivanpauno, line 878: `securitycontext` was renamed `enclave`

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -836,148 +823,282 @@ rmw_ret_t configure_qos_for_security(
...
+
+  context->impl = impl.get();
+
+  // FIXME: not sure if I could get away with using a bunch of listeners
+  // - Locally published samples shouldn't make it to the reader, so there shouldn't be a deadlock
+  //   caused by the graph cache's mutex already having been locked by (e.g.) rmw_create_node.
+  // - Whatever the graph cache implementation does, it shouldn't involve much more than local state
+  //   updates and triggering a guard condition, and so that should be safe.
+  // - There might be a significant cost to calling the graph cache function to update its internal
+  //   state, and having that occur on the thread receiving data from the network may not be wise
```
ivanpauno, line 1003: Yeah, the last point is the main reason I've used a separate thread.
`rmw_dds_common` can probably be largely improved (and the way ros discovery data is shared too).

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -836,148 +823,282 @@ rmw_ret_t configure_qos_for_security(
...
+  impl->common.graph_cache.set_on_change_callback(
+    [guard_condition = graph_guard_condition.get()]() {
+      static_cast<void>(rmw_trigger_guard_condition(guard_condition));
+    });
+
+  impl->common.pub = publisher.get();
+  impl->common.sub = subscription.get();
+  impl->common.graph_guard_condition = graph_guard_condition.get();
+  get_entity_gid(impl->ppant, impl->common.gid);
+  impl->common.graph_cache.add_participant(impl->common.gid, context->options.security_context);
```
ivanpauno, line 991: `security_context` should be replace with `enclave`, everywhere.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -836,148 +823,282 @@ rmw_ret_t configure_qos_for_security(
...
+    return nullptr;
   }
-#endif
 
   /* Since ROS2 doesn't require anything fancy from DDS Subscribers or Publishers, create a single
      pair & reuse that */
+  const dds_entity_t pp = context->impl->ppant;
   dds_entity_t pub, sub;
   if ((pub = dds_create_publisher(pp, nullptr, nullptr)) < 0) {
-    node_gone_from_domain_locked(did);
```
ivanpauno, line 939: just out of curiosity, what are these `publisher` and `subscription` for?

eboasson: These are the DDS publisher and subscriber objects underneath which the individual writers and readers are created. As there is nothing related to the node in them (the "group data" setting would be about the only thing that could've been used to some purpose) there is no benefit to having a pair of these per node. I've moved them to the context.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -836,148 +823,282 @@ rmw_ret_t configure_qos_for_security(
...
+
+extern "C" rmw_ret_t rmw_context_fini(rmw_context_t * context)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    context,
+    context->implementation_identifier,
+    eclipse_cyclonedds_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  delete context->impl;
```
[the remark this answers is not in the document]
eboasson: Good catch! The RMW publisher, subscription and guard condition objects for interfacing wiht the graph cache weren't deleted (the underlying DDS objects did get deleted as a consequence of deleting the participant).

I've added deleting them to the `rmw_context_impl_t` destructor and adjusted the initialisation code a bit so that I think it should all work fine now. I've tried it, also with some fault injection, but I might be getting some of the C++-ey bits wrong. @rotu perhaps you could have a look as well?

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: done

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: I have no idea when the context gets initialized relative to when the first node is created. If the two always go hand in hand, then it is not worth changing. But if applications often create the context much earlier (say, before checking their command-line arguments), it would be better to do it the same way as in `rmw_fastrtps`. I'm happy to go with this for now.

eboasson:
@ivanpauno I've rebased it to get all the updates `rmw_dds_common` and some of the other changes. That's the force-push. The other changes I made based on your comments are in separate comments.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -880,6 +891,12 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t
     return RMW_RET_BAD_ALLOC;
   }
 
+  /* "impl"'s destructor relies on these being initialized properly */
+  impl->common.thread_is_running.store(false);
+  impl->common.graph_guard_condition = nullptr;
+  impl->common.pub = nullptr;
+  impl->common.sub = nullptr;
```
ivanpauno, line 898: can these be initialized in `rmw_context_impl_t` constructor?

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: Ah makes sense, thanks for the change

ivanpauno:
Full CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10249/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5791/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8332/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/10133/)

ivanpauno:
@eboasson `test_echo_pub` of `ros2topic` has hung in all platforms.
In some of them, I tried twice.

Either this PR introduced a race condition, or this change randomly triggered a race condition elsewhere.

I wouldn't discard the second option, considering https://github.com/ros2/build_farmer/issues/248.
I will test locally if I have the time, but I don't promise anything.

Let me know if you can dig into the problem.

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: Yes, that'd be better, thanks

eboasson:
> @eboasson `test_echo_pub` of `ros2topic` has hung in all platforms.
> In some of them, I tried twice.
> 
> Either this PR introduced a race condition, or this change randomly triggered a race condition elsewhere.
> 
> I wouldn't discard the second option, considering [ros2/build_farmer#248](https://github.com/ros2/build_farmer/issues/248).
> I will test locally if I have the time, but I don't promise anything.
> 
> Let me know if you can dig into the problem.

I should be able to at least do some investigation on macOS

eboasson:
@ivanpauno instead of hanging, that test fails for me:
```
======================================================================
FAIL: test_echo_basic (test_echo_pub.TestROS2TopicEchoPub) (topic='/clitest/topic/echo_incompatible_qos', provide_qos=True, compatible_qos=False)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/erik/ros2_ws/src/ros2/ros2cli/ros2topic/test/test_echo_pub.py", line 250, in test_echo_basic
    assert not command.output, (
AssertionError: Echo CLI should not have received anything with incompatible QoS
assert not 'Incompatible QoS Policy detected: RELIABILITY_QOS_POLICY\n'
 +  where 'Incompatible QoS Policy detected: RELIABILITY_QOS_POLICY\n' = <launch_testing.tools.process.ProcessProxy object at 0x1090be7d0>.output
```
There's no way the QoS matching suddenly stopped working, but it could be a publisher process that keeps running for too long. That also bears some similarity to a hanging test.

I also get, at the very end:
```
ros discovery info listener thread: wait failed, will shutdown ...
```
and that is weird, too, and might be a clue to the other symptoms. (@dennis-adlink has kindly put some effort into reproducing it as well so I could do some other urgent things, but last I heard he didn't encounter any failure.) These details need to be understood first ...

Also, the CI run now fails because I added a function to Cyclone (`dds_get_guid`, https://github.com/eclipse-cyclonedds/cyclonedds/pull/483) to avoid having to write some truly ridiculous code in `rmw_cyclonedds`, but the CI run now uses Cyclone's `releases/0.5.x` branch which doesn't have that. I think for Foxy the CI should be using master. @rotu, I know I saw a few PRs here and there regarding the switching of Cyclone branches used by ROS2 but I never delved into the details. Would you have time to check why our local CI (and perhaps the "real" one, too) switched branches?

eboasson:
The
```
ros discovery info listener thread: wait failed, will shutdown ...
```
message was caused by a global destructor shutting down Cyclone DDS while the discovery thread was still running, which in turn was a consequence of the process not deleting the RMW objects before terminating. I've pushed a fix for that (which simply means it ends abruptly, but that is probably the intended behaviour).

Other than that, the error I'm getting:
```
test_echo_basic (test_echo_pub.TestROS2TopicEchoPub) ... [WARN] [1587305691.385956000] [cli_echo_pub.cli_echo_pub_test_node]: New subscription discovered on this topic, requesting incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
```
is caused by test script being unable to handle the warnings for incompatible QoS pairs that ROS2 now prints. When using Fast-RTPS the test succeeds because that Fast-RTPS doesn't support the feature yet (https://github.com/eProsima/Fast-RTPS/issues/852).

Each test takes ridiculously long, but I have not seen any hanging tests. So I assume your hunch that the hangs tests might be caused by some things in other packages is correct.

I think we ought to do another CI run with the fixes I pushed since the previous one and see what comes out.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -738,7 +735,6 @@ static bool check_create_domain(dds_domainid_t did, rmw_localhost_only_t localho
       RCUTILS_LOG_ERROR_NAMED(
         "rmw_cyclonedds_cpp",
         "rmw_create_node: failed to create domain, error %s", dds_strretcode(dom.domain_handle));
-      dds_delete(dom.domain_handle);
```
ivanpauno: why was this line deleted?

ivanpauno:
> Would you have time to check why our local CI (and perhaps the "real" one, too) switched branches?

We're also using `releases/0.5.x` in CI.

> I think we ought to do another CI run with the fixes I pushed since the previous one and see what comes out.

I will give it a try locally, and then run CI with both this change and `cyclonedds` master.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: I think this was an outdated diff, resolving.

ivanpauno:
Locally, it works fine.
Another CI round:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10334/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5861/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8418/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/10242/)

ivanpauno:
@eboasson Failures are unrelated, so I will merge this after the update in `ros2.repos` file is merged.

Thanks for working on this!

ivanpauno:
Ok, I will merge this one, as `cyclonedds` branch was moved to `master` again, thanks @eboasson.

@rotu I think it would be good to update the checks here to use `master` again.

rotu:
> @rotu I think it would be good to update the checks here to use `master` again.

If you mean https://github.com/ros2/ros2/pull/905, then yes, this corrects my mistake. If you mean something else, please elaborate.

clalancette:
@ivanpauno @eboasson I think this commit is causing build warnings on Linux:

```
15:37:21 [ 10%] Building CXX object CMakeFiles/rmw_cyclonedds_cpp.dir/src/rmw_node.cpp.o
15:37:21 /home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp: In function ‘void discovery_thread(rmw_context_impl_t*)’:
15:37:21 /home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:579:80: warning: type qualifiers ignored on cast result type [-Wignored-qualifiers]
15:37:21   579 |   auto sub = static_cast<CddsSubscription const * const>(impl->common.sub->data);
15:37:21       |                                                                                ^
15:37:21 /home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:580:96: warning: type qualifiers ignored on cast result type [-Wignored-qualifiers]
15:37:21   580 |   auto gc = static_cast<CddsGuardCondition const * const>(impl->common.listener_thread_gc->data);
15:37:21       |                              
```

ivanpauno:
Sorry, I checked the test failures but I didn't see the warnings.
I will work on a fix.

rotu:
I think this created a build mess by introducing a dependency on FastRTPS.

```
colcon graph --packages-up-to rmw_cyclonedds_cpp --packages-above fastrtps   
fastrtps                         +**.......
rosidl_typesupport_fastrtps_cpp   +*..***..
rosidl_typesupport_fastrtps_c      +*..**..
rosidl_typesupport_c                +****..
rosidl_generator_py                  + **..
rosidl_typesupport_cpp                +**..
rosidl_default_generators              + * 
rosidl_default_runtime                  +*.
rmw_dds_common                           +*
rmw_cyclonedds_cpp                        +
```

ivanpauno:
> I think this created a build mess by introducing a dependency on FastRTPS.
> 
> ```
> colcon graph --packages-up-to rmw_cyclonedds_cpp --packages-above fastrtps   
> fastrtps                         +**.......
> rosidl_typesupport_fastrtps_cpp   +*..***..
> rosidl_typesupport_fastrtps_c      +*..**..
> rosidl_typesupport_c                +****..
> rosidl_generator_py                  + **..
> rosidl_typesupport_cpp                +**..
> rosidl_default_generators              + * 
> rosidl_default_runtime                  +*.
> rmw_dds_common                           +*
> rmw_cyclonedds_cpp                        +
> ```

No, it doesn't. Please continue the discussion in https://github.com/ros2/rmw_dds_common/issues/16.

PR #146, opened 2020-04-09
security-context -> enclave

mikaelarguedas:
Signed-off-by: Mikael Arguedas <mikael.arguedas@gmail.com>

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -705,7 +705,7 @@ static std::string get_node_user_data(
   const char * node_name, const char * node_namespace, const char * enclave)
 {
   return get_node_user_data_name_ns(node_name, node_namespace) +
-         std::string("securitycontext=") + std::string(enclave) +
+         std::string("enclave=") + std::string(enclave) +
          std::string(";");
 }
 #else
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
CI at https://github.com/ros2/sros2/pull/198#issuecomment-612017072

ivanpauno:
You can ignore the failing checks, as they don't take in account simultaneous in other repos.
About the `Dashing` check, it seems to be failing because of unrelated reasons.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3215,7 +3215,7 @@ extern "C" rmw_ret_t rmw_get_node_names_impl(
   }
 
   std::regex re {
-    "^name=([^;]*);namespace=([^;]*);(securitycontext=([^;]*);)?",
+    "^name=([^;]*);namespace=([^;]*);(enclave=([^;]*);)?",
```
ruffsl, line 3218: What is this for? Why didn't it use an underscore before?

mikaelarguedas: I believe that;s for user data.
looks like no underscore was used in any implementations 
https://github.com/ros2/rmw_connext/pull/407/files#diff-2469e9c52ef0878455668e9233e96b0aL79
https://github.com/ros2/rmw_fastrtps/pull/365/files#diff-9fc4d7c39bf46122c5317462daae5306L76

ivanpauno: >  What is this for? Why didn't it use an underscore before?

The others `rmw` implementations use a function to parse the key that don't accept a underscores: https://github.com/ros2/rmw/blob/6c513772fec7ccafe4eaaf5f838fbcfcc0d92393/rmw/include/rmw/impl/cpp/key_value.hpp#L70.

Using `enclave` instead of `security_context` here is correct.

PR #147, opened 2020-04-10
Register RMW output filters

hidmic:
This pull request registers regular expressions for output filtering in launch tests that use `rmw_cyclonedds_cpp` as their RMW implementation of choice. It aims to solve test failures as seen in https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/.

CI up to `ros2topic`, `ros2action` and `ros2service`:

* Debug MacOS [Build Status](http://ci.ros2.org/job/ci_osx/8193/)

CC @nuclearsandwich


rotu:
There are a *lot* of test failures in your link. Which test failure[s] does this address?

I think a more direct fix would be to suppress the output from CycloneDDS that's making the tests fail.

hidmic:
> There are a `lot` of test failures in your link. Which test failure[s] does this address?

That's an absolutely fair question. It's:

* https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/ros2lifecycle.src.ros2.ros2cli.ros2lifecycle.test/test_cli/test_cli/
* https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/ros2service.src.ros2.ros2cli.ros2service.test/test_cli/test_cli/
* https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/ros2topic.src.ros2.ros2cli.ros2topic.test/test_cli/test_cli/
* https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/ros2action.src.ros2.ros2cli.ros2action.test/test_cli/test_cli/

> I think a more direct fix would be to suppress the output from CycloneDDS that's making the tests fail.

I don't mind either way. I figured that log shows up in debug builds for a (good) reason.

rotu:
Oof those failure messages are so meaningless as to be inactionable. And the stack traces aren't even stack traces! Is there already an issue for this or should I submit one?

I think a much more sound solution is to use the already defined `rmw_set_log_severity` and for us to respect it. I'll add it to `rmw_cyclonedds_cpp`.

rotu:
Actually looking at this further, it seems like this might simply be fixed more simply and robustly by removing the `strict` flag to `expect_output`, which makes it match the text *in* the output, as opposed to requiring a full text match.

dirk-thomas:
Since this is an incremental improvement and resolves a test failure and there is precedence for this approach (for other rmw impl. similar output filters are defined) we can merge this as is. Of course this can be revised in follow up PR if a better approach is available (and if applicable this change can be reverted then).

rotu:
> Since this is an incremental improvement and resolves a test failure and there is precedence for this approach (for other rmw impl. similar output filters are defined) we can merge this as is. Of course this can be revised in follow up PR if a better approach is available (and if applicable this change can be reverted then).

Whoa! NO. I disagree. None of this justifies compromising code quality for a hacky fix. Switching the `strict` option in the test would be a similar incremental improvement resolving a test failure and there is more precedence for "fixing broken tests".

rotu:
Adding log severity in #149

dirk-thomas:
> Switching the `strict` option in the test would be a similar incremental improvement resolving a test failure and there is more precedence for "fixing broken tests".

Since you didn't want to continue the conversation on the PR you opened as a replacement to this I will repeat my comment from [there](https://github.com/ros2/rmw_cyclonedds/pull/149#issuecomment-612332017) here:

These CLI tests checks for strict output for a reason. They not only ensure that certain output is present but also ensure that no other output like warnings / errors / wrong information is present. So no, `strict=False` is not an appropriate solution.

Please either suggest a viable alternative or this PR should be reconsidered since it actually resolves the test failures.

I am going to reopen this PR to ensure the issue doesn't get lost.

rotu:
That PR is not a replacement. It is a problem I noticed when investigating the root cause and possible, more correct, fixes to these test failures.

rotu:
Closing again so this doesn’t get accidentally merged. I’m confident that the correct fix does not lie in the RMW layer. Created issue https://github.com/ros2/ros2cli/issues/482 so underlying issue doesn’t get lost.

dirk-thomas:
Since the pull request is solving the test failures it should stay open. Especially since there is not acceptable alternative proposal atm. If you are worried that it gets merged by accident it can be marked as a draft.

ivanpauno:
> Actually looking at this further, it seems like this might simply be fixed more simply and robustly by removing the strict flag to expect_output, which makes it match the text in the output, as opposed to requiring a full text match.

Removing `strict` from the checks isn't always a good option.
For example [here](https://github.com/ros2/ros2cli/blob/bf4941afd06064d20914c19687a8ca98186687e6/ros2topic/test/test_cli.py#L190-L198), that wouldn't control if a node appears twice, or if something else appears in the output.

The output filters registered here are just used to filter logs in tests (like [here](https://github.com/ros2/ros2cli/blob/bf4941afd06064d20914c19687a8ca98186687e6/ros2topic/test/test_cli.py#L157)), and not used elsewhere.
So they don't do any harm.

> Adding log severity in #149

That sounds like a good option. But if it's not ready yet, the proposal here doesn't sound like a bad idea.

rotu:
These output filters may be used to filter logs in tests but they imply that the log message is spurious and other packages may pick up the bad habit. Also, the message strings may change in the future, are not part of the compatibility interface, and a change to them in the cyclonedds repo should not require a change here to keep the build passing.

For that test you mentioned, you don’t check that the output of the tool show up on stdout - which is really important! (and would automatically filter out diagnostic input like this)

ivanpauno:
> For that test you mentioned, you don’t check that the output of the tool show up on stdout - which is really important! (and would automatically filter out diagnostic input like this)

I'm not sure of what stream the test is checking, but checking only `stdout` sounds like a reasonable thing to do.

dirk-thomas:
> Also, the message strings may change in the future, are not part of the compatibility interface, and a change to them in the cyclonedds repo should not require a change here to keep the build passing.

it is certainly the case that when the message changes the registered filters have to be updated accordingly. But since no viable alternative is currently only the table this is an acceptable approach in order to get these tests to pass.

> I'm not sure of what stream the test is checking, but checking only `stdout` sounds like a reasonable thing to do.

The current test explicitly ensures that no error messages (or anything else) is printed on `stderr`. Only checking `stdout` would remove that check and additional `stderr` wouldn't be catched by the test anymore. Imo that is undesired since the test tries to ensure that the whole output a users sees is as expected.

rotu:
> But since no viable alternative is currently only the table this is an acceptable approach in order to get these tests to pass.

Here are some alternatives I'm willing to entertain: https://github.com/ros2/ros2cli/issues/482#issuecomment-612473234
None of them involve modifying this repo to involve suppressing that warning message, which is what this PR as-is would do. The best fix would be to confirm stdout output and to flag stderr output for manual review, similar to how pytest handles unanticipated warnings. Please let it go, @dirk-thomas.

dirk-thomas:
> Please let it go

I will "let it go" as soon as these tests are not failing anymore since the goal for ROS 2 is to have all tests passing and CI being green.

rotu:
Cyclone and the RMW are working correctly. If you want to fix the tests correctly, it involves updating the test environment or the test expectations, not adding a new feature to the rmw codebase in this repo.

eboasson:
My 2 cents:

* https://github.com/eclipse-cyclonedds/cyclonedds/issues/485 addresses the root cause of this, but I don't think it is realistic to assume it'll get done in the few days left before Foxy;
* #149 is in principle the right way to supressing the message, but it runs into some unfinished business in Cyclone configuration: all the logging & tracing settings are really per domain and the function call you use doesn't affect any logging done once it has enabled the per-domain logging/tracing. It therefore requires a companion fix in Cyclone: take the programmatically set options as a default for the per-domain one. The tricky part is making sure it all behaves somewhat intuitively (not that the current situation is all that intuitive ...)

I believe one's got to know when to be pragmatic and when to go for perfection. I don't see why the output filter isn't a pragmatic *interim* solution while we work on the logging & network configuration of cyclone.

dirk-thomas:
> I believe one's got to know when to be pragmatic and when to go for perfection. I don't see why the output filter isn't a pragmatic interim solution while we work on the logging & network configuration of cyclone.

@eboasson I couldn't agree more. Please consider merging this PR then (or let me know if it needs any kind of update). I rather don't want to push the merge button myself after this discussion.

eboasson:
> > I believe one's got to know when to be pragmatic and when to go for perfection. I don't see why the output filter isn't a pragmatic interim solution while we work on the logging & network configuration of cyclone.
> 
> @eboasson I couldn't agree more. Please consider merging this PR then (or let me know if it needs any kind of update). I rather don't want to push the merge button myself after this discussion.

Sensible!

@rotu I'm happy to make good on my promise on the logging (https://github.com/eclipse-cyclonedds/cyclonedds/issues/270) which will allow #149 to do what is supposed to do. I'll do my best to get it done quickly.

rotu:
@ebasson, I agree with everything except the conclusion. Interim solutions become all too often permanent, it doesn’t belong in this repo, and it entrenches a “worst practice” of ignoring warning messages.

PR #148, opened 2020-04-10
Add support for taking a sequence of messages

mjcarroll:
Implementation of ros2/rmw#212.

Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -2163,6 +2163,61 @@ static rmw_ret_t rmw_take_int(
   return RMW_RET_OK;
 }
 
+static rmw_ret_t rmw_take_seq(
+  const rmw_subscription_t * subscription,
+  size_t count,
+  rmw_message_sequence_t * message_sequence,
+  rmw_message_info_sequence_t * message_info_sequence,
+  size_t * taken)
+{
+  RET_NULL(taken);
+  RET_NULL(message_sequence);
+  RET_NULL(message_info_sequence);
+  RET_WRONG_IMPLID(subscription);
+
+  if (count > message_sequence->capacity) {
+    RMW_SET_ERROR_MSG("Insuffient capacity in message_sequence");
+    return RMW_RET_ERROR;
+  }
+
+  if (count > message_info_sequence->capacity) {
+    RMW_SET_ERROR_MSG("Insuffient capacity in message_info_sequence");
+    return RMW_RET_ERROR;
+  }
+
+  CddsSubscription * sub = static_cast<CddsSubscription *>(subscription->data);
+  RET_NULL(sub);
+
+  std::vector<dds_sample_info_t> infos(count);
+  auto ret = dds_take(sub->enth, message_sequence->data, infos.data(), count, count);
+
+  if (ret <= 0) {
+    return RMW_RET_ERROR;
+  }
+
+  for (int ii = 0; ii < ret; ++ii) {
+    const dds_sample_info_t & info = infos[ii];
+    rmw_message_info_t * message_info = &message_info_sequence->data[ii];
+
+    if (info.valid_data) {
+      if (message_info) {
+        message_info->publisher_gid.implementation_identifier = eclipse_cyclonedds_identifier;
+        memset(message_info->publisher_gid.data, 0, sizeof(message_info->publisher_gid.data));
+        assert(sizeof(info.publication_handle) <= sizeof(message_info->publisher_gid.data));
+        memcpy(
+          message_info->publisher_gid.data, &info.publication_handle,
+          sizeof(info.publication_handle));
+      }
+    }
+  }
+
+  *taken = ret;
+  message_sequence->size = ret;
+  message_info_sequence->size = ret;
+
+  return RMW_RET_OK;
+}
+
 static rmw_ret_t rmw_take_ser_int(
   const rmw_subscription_t * subscription,
   rmw_serialized_message_t * serialized_message, bool * taken,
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1816,6 +1816,61 @@ static rmw_ret_t rmw_take_int(
...
+
+    if (info.valid_data) {
+      if (message_info) {
+        message_info->publisher_gid.implementation_identifier = eclipse_cyclonedds_identifier;
+        memset(message_info->publisher_gid.data, 0, sizeof(message_info->publisher_gid.data));
+        assert(sizeof(info.publication_handle) <= sizeof(message_info->publisher_gid.data));
+        memcpy(
+          message_info->publisher_gid.data, &info.publication_handle,
+          sizeof(info.publication_handle));
+      }
```
wjwwood, line 1863: What's the else case here? Why would that be null, and if it is, is that ok?

eboasson: message_info can't be null, I dare say. What I am more concerned about is what happens when `!valid_data`: that's a case the ROS2 applications don't expect. In Cyclone's case, the sample will be all-zero (except for the key fields, but there are none of those in ROS2).

You only get those in some edge cases: when the last publisher disappears, when a publisher loses its liveliness, stuff like that. I think those need to be hidden from the caller, just like `rmw_take` does.

mjcarroll: @eboasson Since we aren't currently controlling any allocation here, would it be sufficient to rearrange the sequences to keep all of the valid samples/infos at the beginning of the sequence, and indicate with the `size` field how many are valid?

mjcarroll: I went ahead and implemented it that way in ef76435

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1995,6 +1995,79 @@ static rmw_ret_t rmw_take_int(
...
+    return RMW_RET_ERROR;
+  }
+
+  CddsSubscription * sub = static_cast<CddsSubscription *>(subscription->data);
+  RET_NULL(sub);
+
+  std::vector<dds_sample_info_t> infos(count);
+  auto ret = dds_take(sub->enth, message_sequence->data, infos.data(), count, count);
+
+  if (ret <= 0) {
```
eboasson, line 2026: `ret == 0` simply means no data was taken, which is (correctly, in my opinion) not an error according to https://github.com/ros2/rmw/blob/257ed01c5e90c701d086842f6013bc2416c5ebd9/rmw/include/rmw/rmw.h#L714-L750 
I suspect this should be `ret < 0`.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1995,6 +1995,79 @@ static rmw_ret_t rmw_take_int(
...
+
+  // Keep track of taken/not taken to reorder sequence with valid messages at the front
+  std::vector<void *> taken_msg;
+  std::vector<void *> not_taken_msg;
+
+  for (int ii = 0; ii < ret; ++ii) {
+    const dds_sample_info_t & info = infos[ii];
+
+    void * message = &message_sequence->data[ii];
+    rmw_message_info_t * message_info = &message_info_sequence->data[*taken];
```
eboasson, line 2038: I see no initialization of `*taken` to 0 and according to the interface description it is an out argument, not an in/out one.

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: @mjcarroll my apologies for not responding, but yes, that works as long as the upper layers don't mind the pointers getting reordered. I suppose those pointers ultimately come from the application and I don't know what the expectation of applications will be in this.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1995,6 +1995,79 @@ static rmw_ret_t rmw_take_int(
...
+
+  for (size_t ii = 0; ii < taken_msg.size(); ++ii) {
+    message_sequence->data[ii] = taken_msg[ii];
+  }
+
+  for (size_t ii = 0; ii < not_taken_msg.size(); ++ii) {
+    message_sequence->data[ii + taken_msg.size()] = not_taken_msg[ii];
+  }
+
+  *taken = taken_msg.size();
```
eboasson, line 2064: `*taken` is already equal to `taken_msg.size()` at this point if it is always 0 at the start of the loop over messages returned by `dds_take`.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1995,6 +1995,79 @@ static rmw_ret_t rmw_take_int(
...
+  message_sequence->size = taken_msg.size();
```
eboasson, line 2065: With this, I wonder if it wouldn't make more sense to simply remove `taken`: with the number of taken samples always stored in `message_sequence->size` it doesn't really add anything.

rmw_cyclonedds_cpp/src/rmw_node.cpp
mjcarroll: In general I would expect the upper layers to do the allocation outside of the message_sequence structure, and only populate the pointers for the purpose of retrieving from the middleware.  In that case, I can't imagine that reordering the pointers should cause substantial issues, as long as it is documented.

rmw_cyclonedds_cpp/src/rmw_node.cpp
mjcarroll: You are correct, it should have been initialized to zero.

rmw_cyclonedds_cpp/src/rmw_node.cpp
mjcarroll: Agreed.

rmw_cyclonedds_cpp/src/rmw_node.cpp
mjcarroll: Agreed.

rmw_cyclonedds_cpp/src/rmw_node.cpp
mjcarroll: I was trying to mirror the `rmw_take` API, which has a similar flag, but it would appear there is some duplication here.  It was my intention that the API consumer would be explicitly checking the `taken` flag to determine if messages were taken before iterating over the structures.

PR #149, opened 2020-04-10
Implement rmw_set_log_severity

rotu:
The existing `rmw_set_log_severity` currently just throws an error. This implements it.

PR #153, opened 2020-04-14
Make case fallthrough explicit

rotu:
as per https://github.com/ros2/rmw_cyclonedds/pull/149#issuecomment-613130793

### DIFF
--- a/rmw_cyclonedds_cpp/CMakeLists.txt
+++ b/rmw_cyclonedds_cpp/CMakeLists.txt
@@ -22,7 +22,7 @@ if(NOT CMAKE_CXX_STANDARD)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Wextra -Wpedantic)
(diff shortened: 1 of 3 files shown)


rotu:
@Karsten1987, wanna take a look? -Wimplicit-fallthrough is disabled on Clang by default, but if we're obeying it, no reason not to turn it on!

rmw_cyclonedds_cpp/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ if(NOT CMAKE_CXX_STANDARD)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Wextra -Wpedantic)
+  add_compile_options(-Wall -Wextra -Wpedantic -Wimplicit-fallthrough)
```
Karsten1987, line 25: why not disabling the warning all together with `-Wno-implicit-fallthrough` or `-Wimplicit-fallthrough=0`

rotu: Because the goal is not to remove fallthrough - it's to make sure wherever we us it, it's intentional and obvious.

PR #154, opened 2020-04-14
correct fallthrough macro

Karsten1987:
The current fallthrough macro gave me a warning now on OSX when compiling with `-Werror`:

```
error: use of the 'fallthrough' attribute is a C++17 extension [-Werror,-Wc++17-extensions]
      FALLTHROUGH;
```

The reason, as far as I can tell, is that the compiler is c++17 enabled, yet we're compiling the project with a C++14 standard. 
Also I think there was a typo when choosing the compiler version for 17, according to
https://gcc.gnu.org/onlinedocs/cpp/Standard-Predefined-Macros.html
> 201703L for the 2017 C++ standard 

EDIT: ~Looking into a default behavior for gcc7, I found this: https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/ and applied the `__attribute__  ((fallthrough))`~
As to be expected, MSVC doesn't know that attribute. Going back to comment as default behavior.

@rotu let me know what you think

### DIFF
--- a/rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
+++ b/rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
@@ -15,7 +15,7 @@
 #ifndef FALLTHROUGH_MACRO_HPP_
 #define FALLTHROUGH_MACRO_HPP_
 
-#if __has_cpp_attribute(fallthrough) && (__cplusplus >= 201703L)
+#if __cplusplus >= 201603L
 // C++17
 #define FALLTHROUGH [[fallthrough]]
 #elif __has_cpp_attribute(clang::fallthrough)
--- a/rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
+++ b/rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
@@ -15,7 +15,7 @@
 #ifndef FALLTHROUGH_MACRO_HPP_
 #define FALLTHROUGH_MACRO_HPP_
 
-#if __has_cpp_attribute(fallthrough) || (__cplusplus >= 201603L)
+#if __has_cpp_attribute(fallthrough) && (__cplusplus >= 201703L)
 // C++17
 #define FALLTHROUGH [[fallthrough]]
 #elif __has_cpp_attribute(clang::fallthrough)
--- a/rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
+++ b/rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
@@ -26,7 +26,7 @@
 #define FALLTHROUGH [[gnu::fallthrough]]
 #else
 // gcc
-#define FALLTHROUGH __attribute__ ((fallthrough));
+#define FALLTHROUGH /* FALLTHROUGH */
 #endif
 
 #endif  // FALLTHROUGH_MACRO_HPP_
--- a/rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
+++ b/rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
@@ -21,9 +21,12 @@
 #elif __has_cpp_attribute(clang::fallthrough)
 // Clang
 #define FALLTHROUGH [[clang::fallthrough]]
+#elif __has_cpp_attribute(gnu::fallthrough)
+// gcc with gnu extension
+#define FALLTHROUGH [[gnu::fallthrough]]
 #else
 // gcc
-#define FALLTHROUGH /* fallthrough */
+#define FALLTHROUGH __attribute__ ((fallthrough));
 #endif
 
 #endif  // FALLTHROUGH_MACRO_HPP_


Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10184/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5739/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8275/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10067/)

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10185/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5740/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8276/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10068/)

rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
```diff
@@ -15,15 +15,18 @@
...
 #elif __has_cpp_attribute(clang::fallthrough)
 // Clang
 #define FALLTHROUGH [[clang::fallthrough]]
+#elif __has_cpp_attribute(gnu::fallthrough)
+// gcc with gnu extension
+#define FALLTHROUGH [[gnu::fallthrough]]
 #else
 // gcc
-#define FALLTHROUGH /* fallthrough */
+#define FALLTHROUGH __attribute__ ((fallthrough));
```
rotu: The `/* fallthrough */` comment actually *does* suppress the warning on GCC, so the `[[gnu::fallthrough]]` case is unneeded and redundant. If you prefer `[[gnu::fallthrough]]` for GCC we can definitely keep it, as it does lend credibility to the code. The default case should not be `__attribute__((fallthrough))`, as you've discovered.

rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
```diff
@@ -15,15 +15,18 @@
 #ifndef FALLTHROUGH_MACRO_HPP_
 #define FALLTHROUGH_MACRO_HPP_
 
-#if __has_cpp_attribute(fallthrough) || (__cplusplus >= 201603L)
+#if __has_cpp_attribute(fallthrough) && (__cplusplus >= 201703L)
```
rotu, line 18: The `&&` is the real fix here. I think `201603L` and `201703L` are functionally the same for this. 2016 is when it was added to the working draft, 2017 is when it was released. Whichever you prefer!

https://en.cppreference.com/w/cpp/feature_test
https://en.cppreference.com/w/cpp/preprocessor/replace#Predefined_macros

rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
```diff
@@ -15,15 +15,18 @@
...
 #elif __has_cpp_attribute(clang::fallthrough)
 // Clang
 #define FALLTHROUGH [[clang::fallthrough]]
+#elif __has_cpp_attribute(gnu::fallthrough)
+// gcc with gnu extension
+#define FALLTHROUGH [[gnu::fallthrough]]
 #else
 // gcc
-#define FALLTHROUGH /* fallthrough */
+#define FALLTHROUGH /* FALLTHROUGH */
```
rotu, line 29: good call. this should be capital so it works with `-Wimplicit-fallthrough=4`

rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
rotu: Actually, yeah! I like the `[[gnu::fallthrough]]` since it works with `-Wimplicit-fallthrough=5` https://gcc.gnu.org/onlinedocs/gcc-9.3.0/gcc.pdf

rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
rotu: I might even prefer removing the `__has_cpp_attribute(fallthrough)` and just leaving cpp version detection.

Karsten1987: Is the change you requested here to remove the `__has_cpp_attribute(fallthrough)` then?

rotu: Yes, I think that would be the best move to remove the `__has_cpp_attribute(fallthrough)`, since even if it exists, we're pretending it doesn't. Also pick a `__cplusplus` version. It was not a typo, but whichever version feels better to you.

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10197/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5749/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8289/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10084/)

rmw_cyclonedds_cpp/src/fallthrough_macro.hpp
Karsten1987: done in 2f42980

PR #160, opened 2020-04-22
Provide a dummy readcdr function if needed

eboasson:
Current master has a custom reader history cache (RHC) for monitoring the changes to the node graph without having to waste memory on it. https://github.com/eclipse-cyclonedds/cyclonedds/pull/499 at long last adds a companion `dds_readcdr` to the existing `dds_takecdr` that this RMW layer has depended on for a long time, and so the custom RHC needs to updated with a trivial implementation.

#145 changes the mapping of ROS nodes to participants and uses the "graph cache" implemented in `rmw_dds_common`. In that setting, there is no use for this custom RHC and so this PR doesn't apply after #145 has been merged.

### DIFF
--- a/rmw_cyclonedds_cpp/src/graphrhc.cpp
+++ b/rmw_cyclonedds_cpp/src/graphrhc.cpp
@@ -118,6 +118,26 @@ static int graphrhc_take(
   return 0;
 }
 
+#if DDS_HAS_READCDR
+static int graphrhc_readcdr(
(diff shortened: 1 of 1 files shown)


eboasson:
No longer applicable now that #145 has been merged. The change would be potentially be useful for dashing/eloquent, but just retargetting the PR is not a practical approach.

Closing.

PR #163, opened 2020-04-22
Fill in message_info timestamps

iluetkeb:
This fills the timestamps in the message_info.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1979,6 +1979,10 @@ static rmw_ret_t rmw_take_int(
         memcpy(
           message_info->publisher_gid.data, &info.publication_handle,
           sizeof(info.publication_handle));
+        message_info->source_timestamp = info.source_timestamp;
+        if(rcutils_system_time_now(&(message_info->received_timestamp)) != RCUTILS_RET_OK) {
+          message_info->received_timestamp = 0;
+        }
       }
 #if REPORT_LATE_MESSAGES > 0
       dds_time_t tnow = dds_time();
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1972,6 +1972,7 @@ static rmw_ret_t rmw_take_int(
   dds_sample_info_t info;
   while (dds_take(sub->enth, &ros_message, &info, 1, 1) == 1) {
     if (info.valid_data) {
+      *taken = true;
       if (message_info) {
         message_info->publisher_gid.implementation_identifier = eclipse_cyclonedds_identifier;
         memset(message_info->publisher_gid.data, 0, sizeof(message_info->publisher_gid.data));
@@ -1980,8 +1981,10 @@ static rmw_ret_t rmw_take_int(
           message_info->publisher_gid.data, &info.publication_handle,
           sizeof(info.publication_handle));
         message_info->source_timestamp = info.source_timestamp;
+        // TODO(iluetkeb) this must be replaced by a better received timestamp
         if(rcutils_system_time_now(&(message_info->received_timestamp)) != RCUTILS_RET_OK) {
-          message_info->received_timestamp = 0;
+          // this may appear rather severe, but if the clock failed, something is really wrong
+          return RMW_RET_ERROR;
         }
       }
 #if REPORT_LATE_MESSAGES > 0
@@ -1991,7 +1994,6 @@ static rmw_ret_t rmw_take_int(
         fprintf(stderr, "** sample in history for %.fms\n", static_cast<double>(dt) / 1e6);
       }
 #endif
-      *taken = true;
       return RMW_RET_OK;
     }
   }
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1979,6 +1979,10 @@ static rmw_ret_t rmw_take_int(
         memcpy(
           message_info->publisher_gid.data, &info.publication_handle,
           sizeof(info.publication_handle));
+        message_info->source_timestamp = info.source_timestamp;
+        if(rcutils_system_time_now(&(message_info->received_timestamp)) != RCUTILS_RET_OK) {
```
wjwwood, line 1983: We shouldn't just silently ignore a failure to take the current time... This should fail the function, in my opinion.

iluetkeb: I'm now returning an error in this case.

However, this leads to the somewhat counter-intuitive situation that the message has been taken from DDS (and this is marked properly), but the return code is an error. AFAICT, the user needs to free the message in this case. We should check the places where this is used to ensure that actually happens.

wjwwood: The user always needs to destruct (destructor in C++ and fini in C) and free (opposite of malloc or delete, opposite of new) the message, even if take fails...

So I don't think that's an issue.

Taking the data but not giving it to the user is is unfortunate, but we can have a mixed result, like:

> If the error code `RMW_TAKE_SUCCEEDED_BUT_META_DATA_FAILED` is returned, then you should still check `taken` to see if data was taken, but the data in the message info may be incomplete.

Then the user can decide what to do in that case.

I think it is unlikely that this function fails, but in that event, I think failing with a return code (even if the post condition w.r.t. the message being taken is inconvenient) is better than silently ignoring it and not filling the meta data.

wjwwood: To be clear, I think returning `RMW_RET_ERROR` for now is fine. I was just suggesting how we could improve the error reporting/documentation if we felt we needed to do so. I don't feel strongly that is the case, however, since I think this function failing is pretty unlikely in practice.

iluetkeb: I agree on the error case, but I'm thinking we should always return 0 here, to signal that the received timestamp is not really supported.

Regarding the error, I think you're right. Aleast on POSIX, CLOCK_REALTIME is guaranteed to always be present, so in practice this can't fail (at least not with any of its documented error codes). Moreover, the C++ version of this, std::chrono::system_clock::now, is even declared noexcept.

However, obtaining the received_timestamp at the point of taking the message is not representative of when the message is really received. We really need the DDS implementation to support an internal received timestamp (like Fast-RTPS 1.10 does). As long as we cannot rely on all of the implementations to support this, we cannot really use the received timestamp. The only way to distinguish that would be that those implementations that do not support it return 0 to signal this.

Unless you object to this reasoning, I would replace this with 0.

wjwwood:
I'm curious why you changed this, to not have the received timestamp? Does it matter so long as it is consistent within a single implementation?

I don't really mind either way, but it seemed like the received timestamp was more useful than wrong.

iluetkeb:
> I'm curious why you changed this, to not have the received timestamp?

The timestamps are intended to see which event came first. However, when you just take system time at the moment of calling rcl_take_request, the timestamps will merely reflect the call order and so carry no value. In contrast, DDS implementations like Fast-RTPS take the timestamp when they really receive the data and store it in the SampleInfo, so they are actually useful in that case.

For this reason, I think it is important to communicate to the user whether the middleware really supports a genuine received timestamp or not. 0 is used to signify that it doesn't.

eboasson:
Thanks @iluetkeb for helping with implementing this also for cyclone dds, it really makes a difference to me. I just wanted to add a suggestion on how to get a receive time stamp without having to rely on Cyclone to provide it.

The RMW layer relies on a custom sample implementation to perform direct ROS2-to-CDR conversion and to get access to the serialized form for `rmw_take_serialized_message` and any received message will be turned into a sample instance as soon as it is ready to be inserted in the reader history caches. If that’s what you mean by “the receive time stamp” (as I pointed out in https://github.com/ros2/design/issues/259#issuecomment-612440879 there are very many versions), then one way to do it would be to store that time stamp in this custom sample: just add it to https://github.com/ros2/rmw_cyclonedds/blob/2638f4abe87982f45e5301e90fc4200371a613c3/rmw_cyclonedds_cpp/src/serdata.hpp#L48-L61

It won’t be returned in the sample info because cyclone currently only implements the fields defined in the standard, but that can be worked around by using `dds_takecdr` instead of `dds_take`, like when taking a serialized message:
https://github.com/ros2/rmw_cyclonedds/blob/2638f4abe87982f45e5301e90fc4200371a613c3/rmw_cyclonedds_cpp/src/rmw_node.cpp#L2178
That gives you a *reference* to the sample that you can then deserialize by calling `ddsi_serdata_to_sample` on it. That way you get everything.

(I don’t know what the rules are for loaned samples, but with this sample implementation, and for those topics where the CDR representation matches the in-memory representation and where there are no endianness issues, the same trick would be feasible.)

iluetkeb:
Thanks @eboasson, I've now put this into an issue for future reference: https://github.com/ros2/rmw_cyclonedds/issues/167

PR #168, opened 2020-04-25
Exclude other RMW from Github CI build

### LINKED ISSUE #164: Master CI installing FastRTPS from binary
Discovered when investigating https://github.com/eProsima/Fast-RTPS/issues/1173

```
2020-04-22T17:04:56.5980077Z [1mexecuting command [sudo -H apt-get install -y ros-eloquent-fastcdr][0m
 2020-04-22T17:04:56.5981853Z [1mexecuting command [sudo -H apt-get install -y ros-eloquent-fastrtps][0m
```

(1) rosdistro isn't Eloquent, so this shouldn't happen
(2) I'm not sure why it's installing FastRTPS from binaries.

rotu:
Since #145, the CI build of rmw_cyclonedds_cpp has been failing on Windows due to inadvertently injecting fastrtps into the build process.
fastrtps fails to build (https://github.com/eProsima/Fast-RTPS/issues/1173) causing the CI to fail.
There doesn't seem to be a better way to suppress this in action-ros-ci https://github.com/ros-tooling/action-ros-ci/issues/177

Fixes #164

### DIFF
--- a/.github/resources/suppress_other_rmw.repos
+++ b/.github/resources/suppress_other_rmw.repos
@@ -1,0 +1,11 @@
+definitions:
+  - &empty_repo
+    type: zip
+    url: data:application/zip;base64,UEsFBgAAAAAAAAAAAAAAAAAAAAAAAA==
+
+repositories:
+  ros2/rosidl_typesupport_connext/COLCON_IGNORE: *empty_repo
+  ros2/rmw_connext/COLCON_IGNORE: *empty_repo
+
+  ros2/rosidl_typesupport_fastrtps/COLCON_IGNORE: *empty_repo
+  ros2/rmw_fastrtps/COLCON_IGNORE: *empty_repo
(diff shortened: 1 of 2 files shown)


.github/resources/suppress_other_rmw.repos
```diff
@@ -0,0 +1,11 @@
+definitions:
+  - &empty_repo
+    type: zip
+    url: data:application/zip;base64,UEsFBgAAAAAAAAAAAAAAAAAAAAAAAA==
```
ivanpauno, line 4: To avoid the discussion if `COLCON_IGNORE` can be a directory or only a file, could this zip contain an empty file called `COLCON_IGNORE`?

Then `ros2/rosidl_typesupport_*/` should be used instead of `ros2/rosidl_typesupport_*/COLCON_IGNORE`.

rotu: I tried that.

The problem with having a file is that I would need to overwrite the repo to put it in the right place. This breaks on Window: https://github.com/dirk-thomas/vcstool/issues/130

PR #169, opened 2020-04-25
Provide a dummy readcdr function if needed

eboasson:
The RMW layer in dashing/eloquent depends on a still-evolving interface for custom reader history caches in Cyclone. https://github.com/eclipse-cyclonedds/cyclonedds/pull/499 extends that interface, this PR ensures the sources in the RMW layer will be compatible with Cyclone also after merging that PR.

I believe future proofing by a change of this nature makes sense anyway, but a mishap in implementing #142 means dashing today still relies on cyclonedds' master branch (it is a good thing Cyclone DDS regressions are pretty rare!). Therefore, https://github.com/eclipse-cyclonedds/cyclonedds/pull/499 can only go into Cyclone after either this PR, or https://github.com/ros2/ros2/pull/907 or both have been merged.

Note that Eloquent and Foxy are "safe": Eloquent because it uses cyclonedds:releases/0.5.x and Foxy because #145 removed the reliance on this interface.

### DIFF
--- a/rmw_cyclonedds_cpp/src/graphrhc.cpp
+++ b/rmw_cyclonedds_cpp/src/graphrhc.cpp
@@ -118,6 +118,26 @@ static int graphrhc_take(
   return 0;
 }
 
+#if DDS_HAS_READCDR
+static int graphrhc_readcdr(
+  struct dds_rhc * rhc_cmn, bool lock, struct ddsi_serdata ** values,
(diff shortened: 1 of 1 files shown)


eboasson:
Clearly there is still a need for it (see https://github.com/eclipse-cyclonedds/cyclonedds/issues/538) and it has now been confirmed independently, I'm merging it.

PR #171, opened 2020-04-27
Cast size_t to uint32_t explicitly.

hidmic:
This pull request aims to fix [a warning seen on nightly Windows builds](https://ci.ros2.org/view/nightly/job/nightly_win_rel/1533/msbuild/category.63474818/). Said warning was introduced in https://github.com/ros2/rmw_cyclonedds/pull/148. This was first observed when running CI for [ros2/launch#408](https://github.com/ros2/launch/pull/408#issuecomment-620162001). 

CI up to `rmw_connext` and `rmw_cyclonedds_cpp`:
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/10425/)

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -2187,10 +2187,10 @@ static rmw_ret_t rmw_take_seq(
     return RMW_RET_ERROR;
   }
 
-  if (count > std::numeric_limits<uint32_t>::max()) {
+  if (count > (std::numeric_limits<uint32_t>::max)()) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
       "Cannot take %ld samples at once, limit is %d",
-      count, std::numeric_limits<uint32_t>::max());
+      count, (std::numeric_limits<uint32_t>::max)());
     return RMW_RET_ERROR;
   }
 
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -26,6 +26,7 @@
 #include <tuple>
 #include <utility>
 #include <regex>
+#include <limits>
 
 #include "rcutils/filesystem.h"
 #include "rcutils/format_string.h"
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2190,7 +2190,8 @@ static rmw_ret_t rmw_take_seq(
   RET_NULL(sub);
 
   std::vector<dds_sample_info_t> infos(count);
-  auto ret = dds_take(sub->enth, message_sequence->data, infos.data(), count, count);
+  auto maxsamples = static_cast<uint32_t>(count);
```
clalancette, line 2193: It's not really safe to assume that a `size_t` will fit into a `uint32_t`; on LP64 platforms (like Linux), size_t is 64-bits while uint32_t is 32-bits.  I think we need an earlier check here to make sure that is not the case.

rotu: I think it's okay here. If the code is asking for more than 4294967295 messages at a time, there's not much to do about it. It *would* be nice to have a [`numeric_cast`](https://www.boost.org/doc/libs/1_38_0/libs/numeric/conversion/doc/html/boost_numericconversion/improved_numeric_cast__.html) in `rcpputils` for less safe conversions, though.

clalancette: Clearly if the caller is asking for more than 4294967295, something has gone wrong further up in the stack.  But my point is that if it is asking for more than 4294967295, you'll get weird behavior once you've cast it to a uint32_t.  Take the following code:

```
  size_t count = 0xffffffff;
  fprintf(stderr, "size_t: %lu, uint32_t: %u\n", count, static_cast<uint32_t>(count));
  count++;
  fprintf(stderr, "size_t %lu, uint32_t: %u\n", count, static_cast<uint32_t>(count));
```

On my 64-bit machine, this outputs the following:

```
size_t: 4294967295, uint32_t: 4294967295
size_t 4294967296, uint32_t: 0
```

This would make it tough to debug.  So I still think we need to check that the size will fit into a uint32_t before casting it to that.

hidmic: Good point, done in https://github.com/ros2/rmw_cyclonedds/pull/171/commits/fe8ed460edff790cf2551177e098230bb4aadf12.

PR #172, opened 2020-04-28
Reset security related RMW error state.

hidmic:
If security support is not required, the lack of it is not an error. I think this was introduced in #145.

Without this patch, if building with `ENABLE_SECURITY=off` and  `RCUTILS_REPORT_ERROR_HANDLING_ERRORS` set, the error string set in `configure_qos_security()` is always printed out on any downstream override.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -922,6 +922,8 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t 
     if (context->options.security_options.enforce_security == RMW_SECURITY_ENFORCEMENT_ENFORCE) {
       return RMW_RET_ERROR;
     }
+    // No security enforced, lack of support is not an error.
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -922,6 +922,8 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t
     if (context->options.security_options.enforce_security == RMW_SECURITY_ENFORCEMENT_ENFORCE) {
       return RMW_RET_ERROR;
     }
+    // No security enforced, lack of support is not an error.
+    rmw_reset_error();
```
kyrofa, line 926: I think setting the error state is a bug anyway, fixing it like this feels a tad kludgey. I instead suggest surrounding the generation of the error message with a `if (security_options->security_root_path != nullptr)`. If `security_root_path == nullptr` I'd just return OK, since nothing was requested and we did nothing.

ivanpauno:
Closing in favor of #175

Issue #174, opened 2020-04-29
Lots of tests failing on Windows


ivanpauno:
See for example:

https://ci.ros2.org/view/nightly/job/nightly_win_rel/1535/#showFailuresLink

ivanpauno:
@eboasson FYI.

I will try to narrow down what commit in `rmw_cyclonedds` or `cyclonedds` generated this.

eboasson:
Thanks @ivanpauno!

dirk-thomas:
This is unfortunately not new but was introduced last week between Wednesday and Friday in two stages.

ivanpauno:
@eboasson I'm trying to fix the issue, but if you have any clues help is welcomed.

eboasson:
@ivanpauno @dirk-thomas: a few hours later ... `rmw_cyclonedds_cpp` currently shuts down everything in the `rmw_context_impl_t` destructor but it appears that Windows "helpfully" kills all spawned threads before running the global destructors. This causes the destructor to hang somewhere in deleting a DDS entity because things that must happen suddenly no longer happen. The best solution is obviously to stop supporting Windows altogether.

I guess second best would be starting and stopping Cyclone on node creation/destruction instead of on context creation/destruction: that's when it used to do it before #145, so in practice that should work well enough. Apart from having to rework some of the RMW implementation, there's no downside to that change.

ivanpauno:
> @ivanpauno @dirk-thomas: a few hours later ... rmw_cyclonedds_cpp currently shuts down everything in the rmw_context_impl_t destructor but it appears that Windows "helpfully" kills all spawned threads before running the global destructors

lovely ...

> This causes the destructor to hang somewhere in deleting a DDS entity because things that must happen suddenly no longer happen. The best solution is obviously to stop supporting Windows altogether.

That doesn't sound like something we would want to do.

> I guess second best would be starting and stopping Cyclone on node creation/destruction instead of on context creation/destruction: that's when it used to do it before #145, so in practice that should work well enough. Apart from having to rework some of the RMW implementation, there's no downside to that change.

The approach taken in `rmw_fastrtps_cpp` is similar to that one:
- One participant is created per context.
- It's created after the first `Node` is created in that `Context`.
- It's destructed after the last `Node` is destroyed in that `Context.

Doing the same thing should avoid the problem.

ivanpauno:
@eboasson I was able to reproduce the problem (it's hanging on `ctrl-c`).
[...]

PR #176, opened 2020-04-30
Initialize participant on first use. Destroy participant after last node is destroyed.

ivanpauno:
The same approach was used in `rmw_fastrtps`.
Fixes https://github.com/ros2/rmw_cyclonedds/issues/174.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -280,11 +280,10 @@ struct rmw_context_impl_t
   ~rmw_context_impl_t()
   {
     if (0u != this->node_count) {
-      fprintf(
-        stderr,
+      RCUTILS_SAFE_FWRITE_TO_STDERR(
         "Not all nodes were finished before finishing the context\n."
         "Ensure `rcl_node_fini` is called for all nodes before `rcl_context_fini`,"
-        "to avoid leaking.");
+        "to avoid leaking.\n");
     }
   }
 
@@ -948,7 +947,6 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
       "rmw_cyclonedds_cpp", "rmw_create_node: failed to create DCPSPublication reader");
     return RMW_RET_ERROR;
   }
-
   /* Create DDS publisher/subscriber objects that will be used for all DDS writers/readers
     to be created for RMW publishers/subscriptions. */
   if ((this->dds_pub = dds_create_publisher(this->ppant, nullptr, nullptr)) < 0) {
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -287,6 +287,10 @@ struct rmw_context_impl_t
         "to avoid leaking.");
     }
   }
+
+private:
+  void
+  clean_up();
 };
 
 struct CddsNode
@@ -742,6 +746,21 @@ static bool check_create_domain(dds_domainid_t did, rmw_localhost_only_t localho
   }
 }
 
+static
+void
+check_destroy_domain(dds_domainid_t domain_id)
+{
+  if (domain_id != UINT32_MAX) {
+    std::lock_guard<std::mutex> lock(gcdds.domains_lock);
+    CddsDomain & dom = gcdds.domains[domain_id];
+    assert(dom.refcount > 0);
+    if (--dom.refcount == 0) {
+      static_cast<void>(dds_delete(dom.domain_handle));
+      gcdds.domains.erase(domain_id);
+    }
+  }
+}
+
 #if RMW_SUPPORT_SECURITY
 /*  Returns the full URI of a security file properly formatted for DDS  */
 bool get_security_file_URI(
@@ -863,7 +882,7 @@ rmw_ret_t
 rmw_context_impl_t::init(rmw_init_options_t * options)
 {
   std::lock_guard<std::mutex> guard(initialization_mutex);
-  if (1u != ++this->node_count) {
+  if (0u != this->node_count) {
     // initialization has already been done
     return RMW_RET_OK;
   }
@@ -880,6 +899,7 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
   std::unique_ptr<dds_qos_t, std::function<void(dds_qos_t *)>>
   ppant_qos(dds_create_qos(), &dds_delete_qos);
   if (ppant_qos == nullptr) {
+    this->clean_up();
     return RMW_RET_BAD_ALLOC;
   }
   std::string user_data = std::string("enclave=") + std::string(
@@ -890,12 +910,14 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
       &options->security_options) != RMW_RET_OK)
   {
     if (RMW_SECURITY_ENFORCEMENT_ENFORCE == options->security_options.enforce_security) {
+      this->clean_up();
       return RMW_RET_ERROR;
     }
   }
 
   this->ppant = dds_create_participant(this->domain_id, ppant_qos.get(), nullptr);
   if (this->ppant < 0) {
+    this->clean_up();
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_cyclonedds_cpp", "rmw_create_node: failed to create DDS participant");
     return RMW_RET_ERROR;
@@ -905,6 +927,7 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
   if ((this->rd_participant =
     dds_create_reader(this->ppant, DDS_BUILTIN_TOPIC_DCPSPARTICIPANT, nullptr, nullptr)) < 0)
   {
+    this->clean_up();
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_cyclonedds_cpp", "rmw_create_node: failed to create DCPSParticipant reader");
     return RMW_RET_ERROR;
@@ -912,6 +935,7 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
   if ((this->rd_subscription =
     dds_create_reader(this->ppant, DDS_BUILTIN_TOPIC_DCPSSUBSCRIPTION, nullptr, nullptr)) < 0)
   {
+    this->clean_up();
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_cyclonedds_cpp", "rmw_create_node: failed to create DCPSSubscription reader");
     return RMW_RET_ERROR;
@@ -919,6 +943,7 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
   if ((this->rd_publication =
     dds_create_reader(this->ppant, DDS_BUILTIN_TOPIC_DCPSPUBLICATION, nullptr, nullptr)) < 0)
   {
+    this->clean_up();
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_cyclonedds_cpp", "rmw_create_node: failed to create DCPSPublication reader");
     return RMW_RET_ERROR;
@@ -927,11 +952,13 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
   /* Create DDS publisher/subscriber objects that will be used for all DDS writers/readers
     to be created for RMW publishers/subscriptions. */
   if ((this->dds_pub = dds_create_publisher(this->ppant, nullptr, nullptr)) < 0) {
+    this->clean_up();
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_cyclonedds_cpp", "rmw_create_node: failed to create DDS publisher");
     return RMW_RET_ERROR;
   }
   if ((this->dds_sub = dds_create_subscriber(this->ppant, nullptr, nullptr)) < 0) {
+    this->clean_up();
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_cyclonedds_cpp", "rmw_create_node: failed to create DDS subscriber");
     return RMW_RET_ERROR;
@@ -954,6 +981,7 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
     &pubsub_qos,
     &publisher_options);
   if (this->common.pub == nullptr) {
+    this->clean_up();
     return RMW_RET_ERROR;
   }
 
@@ -968,11 +996,13 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
     &pubsub_qos,
     &subscription_options);
   if (this->common.sub == nullptr) {
+    this->clean_up();
     return RMW_RET_ERROR;
   }
 
   this->common.graph_guard_condition = create_guard_condition();
   if (this->common.graph_guard_condition == nullptr) {
+    this->clean_up();
     return RMW_RET_BAD_ALLOC;
   }
 
@@ -996,19 +1026,16 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
   // the thread receiving data from the network may not be wise.
   rmw_ret_t ret;
   if ((ret = discovery_thread_start(this)) != RMW_RET_OK) {
+    this->clean_up();
     return ret;
   }
+  ++this->node_count;
   return RMW_RET_OK;
 }
 
-rmw_ret_t
-rmw_context_impl_t::fini()
+void
+rmw_context_impl_t::clean_up()
 {
-  std::lock_guard<std::mutex> guard(initialization_mutex);
-  if (0u != --this->node_count) {
-    // destruction shouldn't happen yet
-    return RMW_RET_OK;
-  }
   discovery_thread_stop(common);
   common.graph_cache.clear_on_change_callback();
   if (common.graph_guard_condition) {
@@ -1024,15 +1051,18 @@ rmw_context_impl_t::fini()
     RCUTILS_SAFE_FWRITE_TO_STDERR(
       "Failed to destroy domain in destructor\n");
   }
-  if (domain_id != UINT32_MAX) {
-    std::lock_guard<std::mutex> lock(gcdds.domains_lock);
-    CddsDomain & dom = gcdds.domains[domain_id];
-    assert(dom.refcount > 0);
-    if (--dom.refcount == 0) {
-      static_cast<void>(dds_delete(dom.domain_handle));
-      gcdds.domains.erase(domain_id);
-    }
+  check_destroy_domain(domain_id);
+}
+
+rmw_ret_t
+rmw_context_impl_t::fini()
+{
+  std::lock_guard<std::mutex> guard(initialization_mutex);
+  if (0u != --this->node_count) {
+    // destruction shouldn't happen yet
+    return RMW_RET_OK;
   }
+  this->clean_up();
   return RMW_RET_OK;
 }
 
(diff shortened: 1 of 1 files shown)


eboasson:
Wow, that's super quick!

Of course I had to try, and I get:
```
6: libc++abi.dylib: terminating with uncaught exception of type rclcpp::exceptions::RCLError: failed to create interrupt guard condition: failed to create guardcondition, at /Users/erik/ros2_ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:2618, at /Users/erik/ros2_ws/src/ros2/rcl/rcl/src/rcl/guard_condition.c:90
```
which is
```
There is a running process, kill it and restart?: [Y/n] 
Process 37486 exited with status = 9 (0x00000009) 
Process 37490 launched: '/Users/erik/ros2_ws/build/test_communication/test_publisher_subscriber_cpp' (x86_64)
libddsc.0.5.0.dylib was compiled with optimization - stepping may behave oddly; variables may not be available.
Process 37490 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 2.1
    frame #0: 0x00000001019ea88e libddsc.0.5.0.dylib`dds_create_guardcondition(owner=0) at dds_guardcond.c:42:13 [opt]
   39  	     init here is cheap.  If it is DDS_CYCLONEDDS_HANDLE, we may have to initialise the library, so
   40  	     have to call it.  If it is some bogus value and the library is not initialised yet ... so be
   41  	     it.  Naturally, this requires us to call delete on DDS_CYCLONEDDS_HANDLE afterward. */
-> 42  	  if ((rc = dds_init ()) < 0)
   43  	    return rc;
   44  	
   45  	  if ((rc = dds_entity_lock (owner, DDS_KIND_DONTCARE, &e)) != DDS_RETCODE_OK)
Target 0: (test_publisher_subscriber_cpp) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 2.1
  * frame #0: 0x00000001019ea88e libddsc.0.5.0.dylib`dds_create_guardcondition(owner=0) at dds_guardcond.c:42:13 [opt]
    frame #1: 0x000000010190d77a librmw_cyclonedds_cpp.dylib`::rmw_create_guard_condition(rmw_context_t *) [inlined] create_guard_condition(impl=0x0000000100f04d30) at rmw_node.cpp:2617:29 [opt]
    frame #2: 0x000000010190d75b librmw_cyclonedds_cpp.dylib`::rmw_create_guard_condition(context=<unavailable>) at rmw_node.cpp:2633 [opt]
    frame #3: 0x0000000100b1f215 librcl.dylib`__rcl_guard_condition_init_from_rmw_impl(guard_condition=0x0000000100f05160, rmw_guard_condition=0x0000000000000000, context=0x0000000100f04940, options=rcl_guard_condition_options_t @ 0x00007ffeefbf7e50) at guard_condition.c:86:41 [opt]
    frame #4: 0x000000010071d54a librclcpp.dylib`rclcpp::node_interfaces::NodeBase::NodeBase(this=0x0000000100f050c0, node_name="test_publisher_subscriber_Arrays", namespace_="", context=<unavailable>, rcl_node_options=0x0000000100f05180, use_intra_process_default=<unavailable>, enable_topic_statistics_default=<unavailable>) at node_base.cpp:49:19 [opt]
```

I may be able to help you here: the DDS spec says waitsets and guard conditions are things that you can create and delete at any time and that have no relationship with anything else. Cyclone DDS takes a slightly different tack and considers them regular entities that have a parent.

Because the participant used to be created immediately in the context initialisation, it could use the participant as the owner of all guard conditions, which also means they are guaranteed to be cleaned up when you delete the participant entity. Now the participant is created later and `rclcpp` tries to create a guard condition before the first node/the participant is created.

In Cyclone you don't have to use a participant as the parent (once upon a time, yes, but that's not been the case for quite some time). Above the participant sits a domain entity (created at the same time as the first node), and above that sits the Cyclone library itself, which is a singleton object with constant, pre-defined handle: `DDS_CYCLONEDDS_HANDLE` (this is already used for waitsets: see `rmw_create_wait_set`).

So if you use `DDS_CYCLONEDDS_HANDLE` as the owner when calling `dds_create_guardcondition`, this problem should go away. There are two things to keep in mind here:
* That means it is no longer deleted automatically by the call to `dds_delete` on the participant handle. I believe all guard conditions used are RMW ones, and those are all cleaned up properly anyway, so it is probably not an issue. Otherwise there'll be a memory leak if you create and delete a node in a loop.
* The waitset will only accept entities in the same scope — if the waitset is owned by a participant, it will only accept entities in that participant; and if it is owned by a domain entity, then only in that domain. Here the RMW waitsets should be ok because they have `DDS_CYCLONEDDS_HANDLE` as parent, but it will probably cause trouble with the guard condition in the waitset used in the `discovery_thread`. Changing the owner of that waitset should help.

ivanpauno:
@eboasson I was going to ask how to handle that situation, thanks for answering it beforehand.

> That means it is no longer deleted automatically by the call to dds_delete on the participant handle. I believe all guard conditions used are RMW ones, and those are all cleaned up properly anyway, so it is probably not an issue. Otherwise there'll be a memory leak if you create and delete a node in a loop.

Yes, all `create_guard_condition` calls seem to have a matching `destroy_guard_condition` call.

> The waitset will only accept entities in the same scope — if the waitset is owned by a participant, it will only accept entities in that participant; and if it is owned by a domain entity, then only in that domain. Here the RMW waitsets should be ok because they have DDS_CYCLONEDDS_HANDLE as parent, but it will probably cause trouble with the guard condition in the waitset used in the discovery_thread. Changing the owner of that waitset should help.

Considering that wait sets can be created in `rmw` without passing a context or a node, I think this approach is more adequate.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1091,6 +1133,11 @@ extern "C" rmw_node_t * rmw_create_node(
     return nullptr;
   }
 
+  ret = context->impl->init(&context->options);
+  if (RMW_RET_OK != ret) {
+    return nullptr;
```
eboasson, line 1138: It used to clean up on failure via the destructor, but that no longer happens. Calling `fini` here would have exactly the same effect, I think.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1170,6 +1218,7 @@ extern "C" rmw_ret_t rmw_destroy_node(rmw_node_t * node)
   rmw_free(const_cast<char *>(node->namespace_));
   rmw_node_free(node);
   delete node_impl;
+  node->context->impl->fini();
```
eboasson, line 1221: This looks like a use-after-free of `node` to me

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -861,107 +859,79 @@ rmw_ret_t configure_qos_for_security(
...
+    if (RMW_SECURITY_ENFORCEMENT_ENFORCE == options->security_options.enforce_security) {
       return RMW_RET_ERROR;
     }
   }
-  impl->ppant = dds_create_participant(domain_id, ppant_qos.get(), nullptr);
-  if (impl->ppant < 0) {
+
+  this->ppant = dds_create_participant(this->domain_id, ppant_qos.get(), nullptr);
+  if (this->ppant < 0) {
     RCUTILS_LOG_ERROR_NAMED(
```
sloretz, line 921: If `dds_create_participant()` fails, is it guaranteed that it internally "unref'd" the domain? I doesn't look like `~rmw_context_impl_t()` can be relied on as before because there's no attempt to clean up the domain in there.


https://github.com/ros2/rmw_cyclonedds/blob/3b9bd8056de24c7e5c0ff06c44c3dca6adc4bb11/rmw_cyclonedds_cpp/src/rmw_node.cpp#L280-L289

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -861,107 +859,79 @@ rmw_ret_t configure_qos_for_security(
...
+  if ((this->rd_publication =
+    dds_create_reader(this->ppant, DDS_BUILTIN_TOPIC_DCPSPUBLICATION, nullptr, nullptr)) < 0)
   {
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_cyclonedds_cpp", "rmw_create_node: failed to create DCPSPublication reader");
     return RMW_RET_ERROR;
   }
 
   /* Create DDS publisher/subscriber objects that will be used for all DDS writers/readers
-     to be created for RMW publishers/subscriptions. */
```
sloretz, line 958: nit, accidental whitespace change?

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: Oh, that's a good one. Fixed in https://github.com/ros2/rmw_cyclonedds/pull/176/commits/34c9a437c9123f89ac6b7df609b4efc7c37f141a

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: I think that functions that fail shouldn't leak (i.e.: if `init` failed, there shouldn't be need to explicitly call `fini`).
I used that approach in https://github.com/ros2/rmw_cyclonedds/pull/176/commits/34c9a437c9123f89ac6b7df609b4efc7c37f141a.

ivanpauno:
@eboasson PTAL, I will re-run CI after approval.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -583,7 +580,7 @@ static void discovery_thread(rmw_context_impl_t * impl)
   dds_entity_t ws;
   /* deleting ppant will delete waitset as well, so there is no real need to delete
      the waitset here on error, but it is more hygienic */
-  if ((ws = dds_create_waitset(impl->ppant)) < 0) {
+  if ((ws = dds_create_waitset(DDS_CYCLONEDDS_HANDLE)) < 0) {
```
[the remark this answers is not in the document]
ivanpauno: That phrase applies to lot of places in `rmw_cyclonedds`, `rmw_fastrtps` and `rmw_connext`, so I will ignore it for the moment.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: When `dds_create_participant` fails, `clean_up` is called which calls `check_destroy_domain`, which is "unref'd" the domain.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: Solved in https://github.com/ros2/rmw_cyclonedds/pull/176/commits/ded89f9ab699905649468ff68e128f3587f9c09d.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -263,31 +267,24 @@ struct rmw_context_impl_t
...
-    if (domain_id != UINT32_MAX) {
-      std::lock_guard<std::mutex> lock(gcdds.domains_lock);
-      CddsDomain & dom = gcdds.domains[domain_id];
-      assert(dom.refcount > 0);
-      if (--dom.refcount == 0) {
-        static_cast<void>(dds_delete(dom.domain_handle));
-        gcdds.domains.erase(domain_id);
-      }
+    if (0u != this->node_count) {
+      fprintf(
```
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/rmw_cyclonedds/pull/176/commits/ded89f9ab699905649468ff68e128f3587f9c09d.

rmw_cyclonedds_cpp/src/rmw_node.cpp
sloretz: What if the `dds_create_reader()` calls below fail? Do those unref the participant and domain? It looks like they might also return from this function without cleaning up.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10576/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6025/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8612/)

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: This ones? https://github.com/ros2/rmw_cyclonedds/blob/ded89f9ab699905649468ff68e128f3587f9c09d/rmw_cyclonedds_cpp/src/rmw_node.cpp#L926-L949

All of them also call `clean_up` on failure.
I agree that RAII would be nicer here, but I think this is cleaning up things correctly.

rmw_cyclonedds_cpp/src/rmw_node.cpp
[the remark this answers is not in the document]
eboasson: Also, in Cyclone you’re dealing with integers so I don’t know how easy it would be to retrofit that (I know you can wrap pointers in `unique_ptr`, perhaps there are other tricks as well). Rewriting it to use the “new” OMG C++ DDS API would allow you to use RAII everywhere, but I’m not sure it’d be worth the bother to rewrite it.

ivanpauno:
While investigating another issue, I run in a problem similar to the one commented here https://github.com/ros2/rmw_cyclonedds/pull/176#issuecomment-621915946.
Solved in https://github.com/ros2/rmw_cyclonedds/pull/176/commits/a8e5131ce8b3b2796d12077e8bedc30a2aab3588.

New CI, this time limited to `test_communication`, `test_rclcpp`, `rmw_cyclonedds_cpp` to get results faster:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10586/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6032/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8621/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/10483/)

ivanpauno:
@eboasson @sloretz PTAL

ivanpauno:
Ok, this solves most of the failures, but we've this new one https://ci.ros2.org/job/ci_windows/10483/testReport/junit/test_rclcpp/test_executor__rmw_cyclonedds_cpp/notify/ across all implementations.

ivanpauno:
https://github.com/ros2/rmw_cyclonedds/pull/176/commits/3d807fef26317930e18c1d299d3bc6282974b915 :man_facepalming:.
I think that one was the last bizarre error.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10591/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6037/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8626/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/10489/)

sloretz:
CI summary: Looks good to merge to me

* Linux: no new failures
* Linux aarch64: green
* osx: Possibly one new failure `test_rclcpp.test_parameter_server_cpp__rmw_cyclonedds_cpp.xunit.missing_result`
* windows: no new failures

While there appears to be a new failure of test_parameter_server with cylconedds, there is a longstanding issue https://github.com/ros2/system_tests/issues/420 for the same thing with Fast-RTPS. The console output shows this new failure happening in the same place as Fast-RTPS, so I'm inclined to think it's a test issue rather than caused by this PR

```
64: Test that the second executable terminates after a finite amount of time. ... [INFO] [test_parameters_server_cpp-1]: process started with pid [19605]
64: [INFO] [test_remote_parameters_cpp-2]: process started with pid [19606]
64: [test_remote_parameters_cpp-2] Running main() from /Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/gtest_vendor/src/gtest_vendor/src/gtest_main.cc
64: [test_remote_parameters_cpp-2] [==========] Running 6 tests from 2 test cases.
64: [test_remote_parameters_cpp-2] [----------] Global test environment set-up.
64: [test_remote_parameters_cpp-2] [----------] 3 tests from parameters__rmw_implementation
64: [test_remote_parameters_cpp-2] [ RUN      ] parameters__rmw_implementation.test_remote_parameters_async
64: [test_remote_parameters_cpp-2] Setting parameters
64: [test_remote_parameters_cpp-2] Got set_parameters result
64: [test_remote_parameters_cpp-2] Listing parameters with recursive depth
64: [test_remote_parameters_cpp-2] Listing parameters with depth 1
64: [test_remote_parameters_cpp-2] Listing parameters with depth 2
64: [test_remote_parameters_cpp-2] Getting parameters
64: [test_remote_parameters_cpp-2] Getting nonexistent parameters
64: [test_remote_parameters_cpp-2] Listing parameters with recursive depth
64: [test_remote_parameters_cpp-2] [       OK ] parameters__rmw_implementation.test_remote_parameters_async (195 ms)
64: [test_remote_parameters_cpp-2] [ RUN      ] parameters__rmw_implementation.test_remote_parameters_sync
64/77 Test #64: test_parameter_server_cpp__rmw_fastrtps_cpp .....................***Timeout  60.23 sec
```

```
40: [INFO] [test_remote_parameters_cpp-2]: process started with pid [19388]
40: [test_remote_parameters_cpp-2] Running main() from /Users/osrf/jenkins-agent/workspace/ci_osx/ws/install/gtest_vendor/src/gtest_vendor/src/gtest_main.cc
40: [test_remote_parameters_cpp-2] [==========] Running 6 tests from 2 test cases.
40: [test_remote_parameters_cpp-2] [----------] Global test environment set-up.
40: [test_remote_parameters_cpp-2] [----------] 3 tests from parameters__rmw_implementation
40: [test_remote_parameters_cpp-2] [ RUN      ] parameters__rmw_implementation.test_remote_parameters_async
40: [test_parameters_server_cpp-1] [ERROR] [1588287331.386203070] [rclcpp]: executor taking a service server request from service '/test_parameters_server_allow_undeclared/list_parameters' unexpectedly failed: error not set
40: [test_parameters_server_cpp-1] [ERROR] [1588287331.390162050] [rclcpp]: executor taking a service server request from service '/test_parameters_server_allow_undeclared/set_parameters' unexpectedly failed: error not set
40: [test_parameters_server_cpp-1] [ERROR] [1588287331.396424426] [rclcpp]: executor taking a service server request from service '/test_parameters_server_allow_undeclared/get_parameters' unexpectedly failed: error not set
40: [test_remote_parameters_cpp-2] Setting parameters
40: [test_remote_parameters_cpp-2] Got set_parameters result
40: [test_remote_parameters_cpp-2] Listing parameters with recursive depth
40: [test_remote_parameters_cpp-2] Listing parameters with depth 1
40: [test_remote_parameters_cpp-2] Listing parameters with depth 2
40: [test_remote_parameters_cpp-2] Getting parameters
40: [test_remote_parameters_cpp-2] Getting nonexistent parameters
40: [test_remote_parameters_cpp-2] Listing parameters with recursive depth
40: [test_remote_parameters_cpp-2] [       OK ] parameters__rmw_implementation.test_remote_parameters_async (441 ms)
40: [test_remote_parameters_cpp-2] [ RUN      ] parameters__rmw_implementation.test_remote_parameters_sync
40/77 Test #40: test_parameter_server_cpp__rmw_cyclonedds_cpp ...................***Timeout  60.23 sec
```

PR #177, opened 2020-04-30
Fix topic name mangling when not using ros conventions

ivanpauno:
Fixes https://github.com/ros2/system_tests/pull/428.


ivanpauno:
> Skipping the prefix matches rmw_fastrtps and rmw_connext. I haven't figured out what the suffix is for, though Connext doesn't have an option for a suffix so it probably doesn't matter.

The suffix is added to services `Request` and `Reply` topics.

sloretz:
CI ok in https://github.com/ros2/system_tests/pull/428#issuecomment-622171835 ; merging ahead of nightlies

PR #178, opened 2020-05-07
Remove MANUAL_BY_NODE liveliness API

ivanpauno:
See https://github.com/ros2/rmw/pull/227.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -3045,7 +3045,6 @@ static rmw_ret_t rmw_take_response_request(
       }
     }
   }
-  fprintf(stderr, "taken is false !!!\n");
   *taken = false;
   return RMW_RET_OK;
 }
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1135,15 +1135,6 @@ extern "C" rmw_ret_t rmw_context_fini(rmw_context_t * context)
 ///////////                                                                   ///////////
 /////////////////////////////////////////////////////////////////////////////////////////
 
-extern "C" rmw_ret_t rmw_node_assert_liveliness(const rmw_node_t * node)
-{
-  RET_WRONG_IMPLID(node);
-  if (dds_assert_liveliness(node->context->impl->ppant) != DDS_RETCODE_OK) {
-    return RMW_RET_ERROR;
-  }
-  return RMW_RET_OK;
-}
-
 extern "C" rmw_node_t * rmw_create_node(
   rmw_context_t * context, const char * name,
   const char * namespace_, size_t domain_id,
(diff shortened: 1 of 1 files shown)


rotu:
This is not yet deprecated. Shouldn't we be deprecating instead?

https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#deprecation-strategy
https://semver.org/#how-should-i-handle-deprecating-functionality

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1695,9 +1683,6 @@ static bool dds_qos_to_rmw_qos(const dds_qos_t * dds_qos, rmw_qos_profile_t * qo
       case DDS_LIVELINESS_AUTOMATIC:
         qos_policies->liveliness = RMW_QOS_POLICY_LIVELINESS_AUTOMATIC;
         break;
-      case DDS_LIVELINESS_MANUAL_BY_PARTICIPANT:
```
eboasson, line 1698: Removing this case outright is a problem: this conversion routine is also used by the functions that monitor remote publications/subscriptions to feed it into the graph cache, and so `DDS_LIVELINESS_MANUAL_BY_PARTICIPANT` may still show up in its input.

Ignoring the endpoint altogether would be possible but seems unwise (it might be intentional in a more complex setting where there are some non-ROS2 processes involved), so it needs to be mapped in some sensible way. Given the desire to remove `BY_NODE` and the (in my view) undesirability of adding the cop-add of an `OTHER` mode, mapping it to `AUTOMATIC` seems reasonable.

ivanpauno: does using `RMW_QOS_POLICY_LIVELINESS_UNKNOWN` sound reasonable?

eboasson: Yes, that sounds perfectly reasonable. I hadn't realized it already existed 🙂

ivanpauno:
> lgtm, though I'm not sure if the master branch needs to support more than one version simultaneously, it might require ifdef logic instead.

Master is used only for foxy and later.
`dashing-eloquent` is the one used for previous distrubutions.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1566,9 +1557,6 @@ static dds_qos_t * create_readwrite_qos(
     case RMW_QOS_POLICY_LIVELINESS_UNKNOWN:
       dds_qset_liveliness(qos, DDS_LIVELINESS_AUTOMATIC, ldur);
       break;
-    case RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE:
-      dds_qset_liveliness(qos, DDS_LIVELINESS_MANUAL_BY_PARTICIPANT, ldur);
-      break;
     case RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_TOPIC:
       dds_qset_liveliness(qos, DDS_LIVELINESS_MANUAL_BY_TOPIC, ldur);
       break;
```
wjwwood, line 1562: Does this need a default case statement?

ivanpauno: Done https://github.com/ros2/rmw_cyclonedds/pull/178/commits/7f8a45167a74a12f27572e9ba101a8e30d29ea8c.

rotu:
FYI @eboasson, @jacobperron, this is not backwards compatible with 0.6.0 so we need a new release.

jacobperron:
A new release of rmw_cyclonedds is forthcoming (with releases for all other packages involved in https://github.com/ros2/rmw/pull/227).

ivanpauno:
Release done: https://github.com/ros/rosdistro/pull/24849.

PR #179, opened 2020-05-07
Add rmw_implementation to CI workflow

rotu:
This ensures we maintain compatibility with downstream

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,9 @@ jobs:
     - name: Build and test ROS
       uses: ros-tooling/action-ros-ci@0.0.16
       with:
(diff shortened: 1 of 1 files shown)


rotu:
> I don't know which incompatibilities it prevents, but it looks fine to me.

If we fail to implement a declared RMW function or change its signature from that declared in `rmw.h`, it currently won't break the build. This is the case for https://github.com/ros2/rmw_cyclonedds/pull/178.

Whether or not we *want* to introduce a new binary break having just released 0.6.0 is a separate issue, but our CI should be detecting such breakages so they don't come as a surprise.

eboasson:
Ah, I see. Thanks!

PR #186, opened 2020-05-20
implement safer align function

dodsonmg:
avoids unnecessary casts to pointer and pointer manipulation by reference.

this is the same as PR #141, which was previously merged into `master`, but way ahead of the `dashing-eloquent` fork.

### DIFF
--- a/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/TypeSupport_impl.hpp
+++ b/rmw_cyclonedds_cpp/include/rmw_cyclonedds_cpp/TypeSupport_impl.hpp
@@ -120,12 +120,20 @@ void TypeSupport<MembersType>::setName(const std::string & name)
   this->name = std::string(name);
 }
 
-static inline void *
-align_(size_t __align, void * & __ptr) noexcept
+template<typename T>
+static inline T
+align_int_(size_t __align, T __int) noexcept
+{
+  return (__int - 1u + __align) & ~(__align - 1);
+}
+
+template<typename T>
+static inline T *
+align_ptr_(size_t __align, T * __ptr) noexcept
 {
   const auto __intptr = reinterpret_cast<uintptr_t>(__ptr);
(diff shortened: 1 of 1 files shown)


dodsonmg:
Looks like Windows test failed while acquiring dependencies...

rotu:
@dodsonmg No need. I need to backport https://github.com/ros2/rmw_cyclonedds/pull/181 which uses setup-ros@master to fix this bug: https://github.com/ros-tooling/setup-ros/issues/156

@thomas-moulard time to cut a new release or recommend people use master? https://github.com/ros-tooling/setup-ros/blob/master/README.md#usage

dodsonmg:
For my own planning purposes:  are you waiting for the Windows CI to be fixed before merging (same question for #189)?

rotu:
> For my own planning purposes: are you waiting for the Windows CI to be fixed before considering the merge (same question for #189)?

Sorry, please don't wait on me. I'm taking a break from Cyclone stuff (indefinite but hopefully not long). I defer to @eboasson.

eboasson:
@dodsonmg any reason this isn't a cherry-pick? It looks to me like the change is identical, and keeping the metadata intact would be better.

I don't think there is a need to wait for the windows build failure in the CI to be addressed given that the changes are fine on Foxy and don't interact with the ROS libraries. But if, e.g., @nuclearsandwich or @jacobperron says differently, then it is different.

dodsonmg:
Sure.  I just don't know how to do it.

dodsonmg:
Cherry picking seems to have worked.  Windows build still fails, and for some reason the MacOS build was cancelled...  I can push again to see if I can dislodge the MacOS build?

dodsonmg:
Any other feedback on this one (i.e., advice for getting through the cancelled builds)?

Same for #189.

eboasson:
@dodsonmg no further feedback other than thanks. Given the nature of changes, the fact that the build failures here are unrelated and that it works fine on Foxy, I am good with merging it. The delay is just the result of having too much to do ...

Issue #183, opened 2020-05-12
test_parameter_server_cpp__rmw_cyclonedds_cpp is failing most of the time on linux


tfoote:
In triaging CI failures I've identified this test as failing most of the time. 

Linux Debug: https://ci.ros2.org/job/nightly_linux_debug/1550/testReport/(root)/projectroot/test_parameter_server_cpp__rmw_cyclonedds_cpp/history/

<details>

![Screenshot from 2020-05-12 18-26-37](https://user-images.githubusercontent.com/447804/81761200-2e316280-947e-11ea-9f67-43db6dc813b1.png)

</details>

Linux Release: https://ci.ros2.org/job/nightly_linux_release/1539/testReport/(root)/projectroot/test_parameter_server_cpp__rmw_cyclonedds_cpp/history/

LInux Aarch64: https://ci.ros2.org/job/nightly_linux-aarch64_release/1159/testReport/(root)/projectroot/test_parameter_server_cpp__rmw_cyclonedds_cpp/history/

Here's an example of the failure.
<details>

```
[...]

Issue #74, opened 2019-12-05
Not getting service responses reliably when using CycloneDDS


hidmic:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04, both AMD64 and ARM64
- Installation type:
  - Binaries and from source
- Version or commit hash:
  - b92db52bb14ec097946bfe44d2b337359dddfbab (using https://github.com/ros2/rmw_cyclonedds/commit/4089feee815a4687b30f757ab1c87bf028804c93)
- DDS implementation:
  - CycloneDDS
- Client library (if applicable):
  - rclcpp **only**

#### Steps to reproduce issue

1. On one terminal, run:
```console
RMW_IMPLEMENTATION=rmw_cyclonedds_cpp ros2 run demo_nodes_cpp add_two_ints_server
```

2. On another terminal, run:
```console
RMW_IMPLEMENTATION=rmw_cyclonedds_cpp ros2 run demo_nodes_cpp add_two_ints_client_async
```

#### Expected behavior
A single service call takes place and succeeds i.e. you get logs on each terminal:

1.
```console
[INFO] [add_two_ints_server]: Incoming request
a: 2 b: 3
```

2. 
```console
[INFO] [add_two_ints_client]: Result of add_two_ints: 5
```

#### Actual behavior
The client hangs with no output though the server does log so it appears to have received the request.

#### Additional information
It does not happen on `rclpy`. It also does not happen if bringing up both nodes via `launch` e.g.:

```python
from launch import LaunchDescription
from launch_ros.actions import Node


def generate_launch_description():
    return LaunchDescription([
        Node(
            package='demo_nodes_cpp',
            node_executable='add_two_ints_server',
            output='screen'
        ),
        Node(
            package='demo_nodes_cpp',
            node_executable='add_two_ints_client_async',
            output='screen'
        )
    ])
```

So it's likely a race.

hidmic:
I can still reproduce on `master`.

dirk-thomas:
I can confirm the problem.

My first guess is a race that the client discovers the server and sends it request but when the server receives the request and wants to publish the reply it hasn't discovered the client yet.

With a simple hack (adding a sleep of 500ms into the server callback function) the client receives the response reliable.

I guess we might something like a "wait_for_client" in the same way as we already offer a "wait_for_service".

eboasson:
[...]

PR #187, opened 2020-05-20
Lost service responses (#183, #74)

eboasson:
This PR addresses the service invocation problems by fixing one silly bug in the `rmw_service_server_is_available ` code (not actually checking the number of matched endpoints), and by blocking in `rmw_send_response` until there is reasonable evidence that the response reader has been discovered.

The proper solution (as discussed in #74) makes the `rmw_service_server_is_available` return false until this point has been reached, but as of today, neither the DDS specification nor Cyclone DDS provides the means to do that without the application exchanging information on what has been discovered. It can be done easily enough, but it is a rather significant burden for what is ultimately a rare problem.

Without the workaround (but with the bug fix) it rarely fails. With the workaround added, I have not been able to reproduce it anymore. I've only seen multiple waits in sequence by introducing significant packet loss.

I do expect this workaround to be somewhat controversial ...

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -316,14 +316,9 @@ struct CddsSubscription : CddsEntity
 
 struct client_service_id_t
 {
-  // strangely, the writer_guid in a request id is smaller than the rmw_gid_t
-  // cpplint erroneously complains that
-  //
-  //   sizeof((reinterpret_cast<rmw_request_id_t *>(0))->writer_guid)
-  //
-  // is not a constant, and the 16 is a hard-coded magic number in
-  // rmw_request_id_t ...
-  uint8_t data[16];
+  // strangely, the writer_guid in an rmw_request_id_t is smaller than the identifier in
+  // an rmw_gid_t
+  uint8_t data[sizeof((reinterpret_cast<rmw_request_id_t *>(0))->writer_guid)]; // NOLINT
 };
 
 struct CddsCS
@@ -510,16 +505,15 @@ static void get_entity_gid(dds_entity_t h, rmw_gid_t & gid)
 
 static std::map<std::string, std::vector<uint8_t>> parse_user_data(const dds_qos_t * qos)
 {
+  std::map<std::string, std::vector<uint8_t>> map;
   void * ud;
   size_t udsz;
-  if (!dds_qget_userdata(qos, &ud, &udsz)) {
-    std::map<std::string, std::vector<uint8_t>> emptymap;
-    return emptymap;
-  } else {
+  if (dds_qget_userdata(qos, &ud, &udsz)) {
     std::vector<uint8_t> udvec(static_cast<uint8_t *>(ud), static_cast<uint8_t *>(ud) + udsz);
     dds_free(ud);
-    return rmw::impl::cpp::parse_key_value(udvec);
+    map = rmw::impl::cpp::parse_key_value(udvec);
   }
+  return map;
 }
 
 static bool get_user_data_key(const dds_qos_t * qos, const std::string key, std::string & value)
@@ -3046,10 +3040,14 @@ extern "C" rmw_ret_t rmw_wait(
 ///////////                                                                   ///////////
 /////////////////////////////////////////////////////////////////////////////////////////
 
+using get_matched_endpoints_fn_t = dds_return_t (*)(
+  dds_entity_t h,
+  dds_instance_handle_t * xs, size_t nxs);
+using BuiltinTopicEndpoint = std::unique_ptr<dds_builtintopic_endpoint_t,
+    std::function<void (dds_builtintopic_endpoint_t *)>>;
+
 static rmw_ret_t get_matched_endpoints(
-  dds_entity_t h, dds_return_t (* fn)(
-    dds_entity_t h,
-    dds_instance_handle_t * xs, size_t nxs), std::vector<dds_instance_handle_t> & res)
+  dds_entity_t h, get_matched_endpoints_fn_t fn, std::vector<dds_instance_handle_t> & res)
 {
   dds_return_t ret;
   if ((ret = fn(h, res.data(), res.size())) < 0) {
@@ -3075,21 +3073,19 @@ static void free_builtintopic_endpoint(dds_builtintopic_endpoint_t * e)
   dds_free(e);
 }
 
-static std::unique_ptr<dds_builtintopic_endpoint_t,
-  std::function<void(dds_builtintopic_endpoint_t *)>> get_matched_subscription_data(
+static BuiltinTopicEndpoint get_matched_subscription_data(
   dds_entity_t writer, dds_instance_handle_t readerih)
 {
-  std::unique_ptr<dds_builtintopic_endpoint_t, std::function<void(dds_builtintopic_endpoint_t *)>>
-  ep(dds_get_matched_subscription_data(writer, readerih), &free_builtintopic_endpoint);
+  BuiltinTopicEndpoint ep(dds_get_matched_subscription_data(writer, readerih),
+    free_builtintopic_endpoint);
   return ep;
 }
 
-static std::unique_ptr<dds_builtintopic_endpoint_t,
-  std::function<void(dds_builtintopic_endpoint_t *)>> get_matched_publication_data(
+static BuiltinTopicEndpoint get_matched_publication_data(
   dds_entity_t reader, dds_instance_handle_t writerih)
 {
-  std::unique_ptr<dds_builtintopic_endpoint_t, std::function<void(dds_builtintopic_endpoint_t *)>>
-  ep(dds_get_matched_publication_data(reader, writerih), &free_builtintopic_endpoint);
+  BuiltinTopicEndpoint ep(dds_get_matched_publication_data(reader, writerih),
+    free_builtintopic_endpoint);
   return ep;
 }
 
@@ -3098,7 +3094,7 @@ static const std::string request_id_writer_guid_to_string(const rmw_request_id_t
   std::ostringstream os;
   os << std::hex;
   os << std::setw(2) << static_cast<int>(static_cast<uint8_t>(id.writer_guid[0]));
-  for (size_t i = 1; i < sizeof(id.writer_guid) - 1; i++) {
+  for (size_t i = 1; i < sizeof(id.writer_guid); i++) {
     os << "." << static_cast<int>(static_cast<uint8_t>(id.writer_guid[i]));
   }
   return os.str();
@@ -3109,7 +3105,7 @@ static const std::string csid_to_string(const client_service_id_t & id)
   std::ostringstream os;
   os << std::hex;
   os << std::setw(2) << static_cast<int>(id.data[0]);
-  for (size_t i = 1; i < sizeof(id.data) - 1; i++) {
+  for (size_t i = 1; i < sizeof(id.data); i++) {
     os << "." << static_cast<int>(id.data[i]);
   }
   return os.str();
@@ -3290,10 +3286,8 @@ extern "C" rmw_ret_t rmw_send_response(
   }
   switch (st) {
     case client_present_t::ERROR:
-      // break instead of returns makes gcc happy
-      break;
     case client_present_t::MAYBE:
-      return RMW_RET_TIMEOUT;
+      break;
     case client_present_t::YES:
       return rmw_send_response_request(&info->service, *request_header, ros_response);
     case client_present_t::GONE:
@@ -3796,6 +3790,7 @@ static rmw_ret_t get_topic_name(dds_entity_t endpoint_handle, std::string & name
 static rmw_ret_t check_for_service_reader_writer(const CddsCS & client, bool * is_available)
 {
   std::vector<dds_instance_handle_t> rds, wrs;
+  assert(is_available != nullptr && !*is_available);
   if (get_matched_endpoints(client.pub->enth, dds_get_matched_subscriptions, rds) < 0 ||
     get_matched_endpoints(client.sub->enth, dds_get_matched_publications, wrs) < 0)
   {
@@ -3807,18 +3802,18 @@ static rmw_ret_t check_for_service_reader_writer(const CddsCS & client, bool * i
   for (auto rdih : rds) {
     auto rd = get_matched_subscription_data(client.pub->enth, rdih);
     std::string serviceid;
-    if (rd.get() && get_user_data_key(rd->qos, "serviceid", serviceid)) {
+    if (rd && get_user_data_key(rd->qos, "serviceid", serviceid)) {
       needles.insert(serviceid);
     }
   }
-  if (needles.size() == 0) {
+  if (needles.empty()) {
     return RMW_RET_OK;
   }
   // then scan the writers to see if there is at least one with a service id in the set
   for (auto wrih : wrs) {
     auto wr = get_matched_publication_data(client.sub->enth, wrih);
     std::string serviceid;
-    if (wr.get() &&
+    if (wr &&
       get_user_data_key(
         wr->qos, "serviceid",
         serviceid) && needles.find(serviceid) != needles.end())
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -16,6 +16,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <algorithm>
+#include <chrono>
 #include <iomanip>
 #include <map>
 #include <set>
@@ -75,6 +76,8 @@
 #include "rmw_cyclonedds_cpp/serdes.hpp"
 #include "serdata.hpp"
 #include "demangle.hpp"
+
+using namespace std::literals::chrono_literals;
 
 /* Security must be enabled when compiling and requires cyclone to support QOS property lists */
 #if DDS_HAS_SECURITY && DDS_HAS_PROPERTY_LIST_QOS
@@ -3299,13 +3302,21 @@ extern "C" rmw_ret_t rmw_send_response(
   // workaround: rmw_service_server_is_available should keep returning false until this
   // is a given).
   client_present_t st;
-  while ((st = check_for_response_reader(info->service, reqwrih)) == client_present_t::MAYBE) {
+  std::chrono::system_clock::time_point tnow = std::chrono::system_clock::now();
+  std::chrono::system_clock::time_point tend = tnow + 100ms;
+  while ((st =
+    check_for_response_reader(
+      info->service,
+      reqwrih)) == client_present_t::MAYBE && tnow < tend)
+  {
     dds_sleepfor(DDS_MSECS(10));
+    tnow = std::chrono::system_clock::now();
   }
   switch (st) {
     case client_present_t::FAILURE:
-    case client_present_t::MAYBE:
       break;
+    case client_present_t::MAYBE:
+      return RMW_RET_TIMEOUT;
     case client_present_t::YES:
       return rmw_send_response_request(&info->service, header, ros_response);
     case client_present_t::GONE:
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -16,6 +16,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <algorithm>
+#include <iomanip>
 #include <map>
 #include <set>
 #include <functional>
@@ -244,6 +245,7 @@ struct rmw_context_impl_t
   rmw_dds_common::Context common;
   dds_domainid_t domain_id;
   dds_entity_t ppant;
+  rmw_gid_t ppant_gid;
 
   /* handles for built-in topic readers */
   dds_entity_t rd_participant;
@@ -258,8 +260,12 @@ struct rmw_context_impl_t
   size_t node_count{0};
   std::mutex initialization_mutex;
 
+  /* suffix for GUIDs to construct unique client/service ids
+     (protected by initialization_mutex) */
+  uint32_t client_service_id;
+
   rmw_context_impl_t()
-  : common(), domain_id(UINT32_MAX), ppant(0)
+  : common(), domain_id(UINT32_MAX), ppant(0), client_service_id(0)
   {
     /* destructor relies on these being initialized properly */
     common.thread_is_running.store(false);
@@ -308,10 +314,23 @@ struct CddsSubscription : CddsEntity
   dds_entity_t rdcondh;
 };
 
+struct client_service_id_t
+{
+  // strangely, the writer_guid in a request id is smaller than the rmw_gid_t
+  // cpplint erroneously complains that
+  //
+  //   sizeof((reinterpret_cast<rmw_request_id_t *>(0))->writer_guid)
+  //
+  // is not a constant, and the 16 is a hard-coded magic number in
+  // rmw_request_id_t ...
+  uint8_t data[16];
+};
+
 struct CddsCS
 {
   CddsPublisher * pub;
   CddsSubscription * sub;
+  client_service_id_t id;
 };
 
 struct CddsClient
@@ -489,6 +508,33 @@ static void get_entity_gid(dds_entity_t h, rmw_gid_t & gid)
   convert_guid_to_gid(guid, gid);
 }
 
+static std::map<std::string, std::vector<uint8_t>> parse_user_data(const dds_qos_t * qos)
+{
+  void * ud;
+  size_t udsz;
+  if (!dds_qget_userdata(qos, &ud, &udsz)) {
+    std::map<std::string, std::vector<uint8_t>> emptymap;
+    return emptymap;
+  } else {
+    std::vector<uint8_t> udvec(static_cast<uint8_t *>(ud), static_cast<uint8_t *>(ud) + udsz);
+    dds_free(ud);
+    return rmw::impl::cpp::parse_key_value(udvec);
+  }
+}
+
+static bool get_user_data_key(const dds_qos_t * qos, const std::string key, std::string & value)
+{
+  if (qos != nullptr) {
+    auto map = parse_user_data(qos);
+    auto name_found = map.find(key);
+    if (name_found != map.end()) {
+      value = std::string(name_found->second.begin(), name_found->second.end());
+      return true;
+    }
+  }
+  return false;
+}
+
 static void handle_ParticipantEntitiesInfo(dds_entity_t reader, void * arg)
 {
   static_cast<void>(reader);
@@ -515,18 +561,9 @@ static void handle_DCPSParticipant(dds_entity_t reader, void * arg)
     } else if (si.instance_state != DDS_ALIVE_INSTANCE_STATE) {
       impl->common.graph_cache.remove_participant(gid);
     } else if (si.valid_data) {
-      void * ud;
-      size_t udsz;
-      if (dds_qget_userdata(s->qos, &ud, &udsz)) {
-        std::vector<uint8_t> udvec(static_cast<uint8_t *>(ud), static_cast<uint8_t *>(ud) + udsz);
-        dds_free(ud);
-        auto map = rmw::impl::cpp::parse_key_value(udvec);
-        auto name_found = map.find("enclave");
-        if (name_found != map.end()) {
-          auto enclave =
-            std::string(name_found->second.begin(), name_found->second.end());
-          impl->common.graph_cache.add_participant(gid, enclave);
-        }
+      std::string enclave;
+      if (get_user_data_key(s->qos, "enclave", enclave)) {
+        impl->common.graph_cache.add_participant(gid, enclave);
       }
     }
     dds_return_loan(reader, &raw, 1);
@@ -921,6 +958,7 @@ rmw_context_impl_t::init(rmw_init_options_t * options)
       "rmw_cyclonedds_cpp", "rmw_create_node: failed to create DDS participant");
     return RMW_RET_ERROR;
   }
+  get_entity_gid(this->ppant, this->ppant_gid);
 
   /* Create readers for DDS built-in topics for monitoring discovery */
   if ((this->rd_participant =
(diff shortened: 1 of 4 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3135,6 +3147,16 @@ extern "C" rmw_ret_t rmw_send_response(
   cdds_request_header_t header;
   memcpy(&header.guid, request_header->writer_guid, sizeof(header.guid));
   header.seq = request_header->sequence_number;
+  // if the number of writers matching our request reader equals the number
+  // of readers matching our response writer, we have a pretty decent claim
+  // to have discovered this client's response reader.
+  //
+  // it is pretty horrid ... but blocking the service is the only option if
+  // the client is unable to determine that it has been fully discovered by
+  // the service.
```
hidmic, line 3156: >  it is pretty horrid ...

I'm fine with this as long as we agree that this isn't a solution but a workaround for distros up and including Foxy.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3135,6 +3147,16 @@ extern "C" rmw_ret_t rmw_send_response(
   cdds_request_header_t header;
   memcpy(&header.guid, request_header->writer_guid, sizeof(header.guid));
   header.seq = request_header->sequence_number;
+  // if the number of writers matching our request reader equals the number
+  // of readers matching our response writer, we have a pretty decent claim
+  // to have discovered this client's response reader.
```
hidmic, line 3152: @eboasson nit: consider moving this comment block closer to the `check_for_response_reader()` function.

hidmic: > we have a pretty decent claim

:eyes:

hidmic: Any chance we can tell if remote writer and reader belong to the same server instance? Or at least to the same participant?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3135,6 +3147,16 @@ extern "C" rmw_ret_t rmw_send_response(
...
+  //
+  // it is pretty horrid ... but blocking the service is the only option if
+  // the client is unable to determine that it has been fully discovered by
+  // the service.
+  while (!check_for_response_reader(info->service.sub->enth, info->service.pub->enth)) {
+    dds_sleepfor(DDS_MSECS(10));
+  }
```
hidmic, line 3159: @eboasson what if the client goes away before sending the response? I'm fine with the busy wait, but it should timeout at some point. Unless we can detect that the request writer went away.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3628,7 +3650,6 @@ extern "C" rmw_ret_t rmw_service_server_is_available(
   ret =
     common_context->graph_cache.get_writer_count(sub_topic_name, &number_of_response_publishers);
   if (ret != RMW_RET_OK || 0 == number_of_response_publishers) {
-    // error
```
hidmic, line 3631: @eboasson why this change?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3124,6 +3124,18 @@ static rmw_ret_t rmw_send_response_request(
...
 }
 
+static bool check_for_response_reader(dds_entity_t request_reader, dds_entity_t response_writer)
+{
+  dds_subscription_matched_status sm;
+  dds_publication_matched_status pm;
+  if (dds_get_subscription_matched_status(request_reader, &sm) < 0 ||
+    dds_get_publication_matched_status(response_writer, &pm) < 0)
+  {
+    return RMW_RET_ERROR;
```
hidmic, line 3134: @eboasson 
```suggestion
    return false;
```
?

rmw_cyclonedds_cpp/src/rmw_node.cpp
wjwwood: Wait why is this needed? Shouldn't the requester wait for a bidirectional match before sending a request? If, in the meantime, the response channel goes away I think this should just fail, not block... Blocking without an external option for a timeout is not very nice in my opinion. The user may not even know it is blocking or for how long.

eboasson: The requester is definitely the one that should be doing the waiting. Unfortunately, DDS doesn’t offer a mechanism for waiting for a bidirectional match. Cyclone does guarantee that the writer having matched the reader is sufficient for that reader to receive the data (modulo disconnections and things like that), but that gives you one-way only.

Secondly, the simple counting that is done here (and in other RMW layers) only works if there is but a single server. You really ought to wait until you have a matched pair, but that would additionally require DDS to be aware of these pairs, and it isn’t. We could use, e.g., the `USER_DATA` QoS to make the mapping clear at the RMW level, but if service interoperability is desirable that needs to be agreed in a wider audience.

What has been on my to-do list for Cyclone for some time is to make it possible to wait until all local entities have been discovered by all remote entities, much like “wait for acknowledgements” allows you to do this for a regular writer and its readers. Even then, there are little details, for example, one would have to guarantee that effects of a discovery message node P sent to node Q have occurred prior to P receiving Q’s acknowledgment. That’s not guaranteed anywhere in the spec and also not today in Cyclone because it processes all discovery data asynchronously, but it is an important primitive.

As to this change: I considered a variant (check in `rmw_take_request`) but the same quality of implementation comes out way more complicated there. A good one would put the request aside and retry/prune based on “publication matched” events. It’s feasible (I should still have a proof-of-concept implementation lying around for something similar as far as the waitset goes) but it is significantly more complex. That’d probably be time better spent on fixing the root cause.

Another mapping would use transient local requests and responses, using the request id as key. It is easy enough in Cyclone, but that’d also be quite a big departure from what has been done until now (by any RMW as far as I know).

The only indefinite blocking this does is when there are clients arriving and leaving all the time. If the currently client simply disappears, both counters will end up being the same once the disappearance is detected. But, yeah ...

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: I actually started working out how to do that, and it is definitely possible the determine they are from the same participant. But now that all nodes in a process (well, context) share the same participant that seems insufficiently selective. I couldn’t find a way to tie them to the node (which I think is selective enough) short of using `USER_DATA`. That seemed like it might be unwise. I suppose the graph cache interface could be extended to answer this question.

What we do know here is that we have discovered the client’s writer (else we could never have received its request), and that therefore the client’s writer is accounted for in the “subscription matched” status. Thus, it is pretty likely that the service writer’s match count won’t equal the service’s reader match count unless that client’s reader has been discovered. But there is no guarantee: if another client shows up and its reader happens to be discovered before its writer, you will draw the wrong conclusion.

In the absence of packet loss, that is highly unlikely because it creates the writer first, because under that assumption, the discovery of the writer will precede the discovery of the reader. Indeed, I wouldn’t be surprised if simply swapping the creation of the reader and the writer in the client would make the problem disappear in the test setup.

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: The DDS discovery will discover the disappearance of the client, remove all matches with it and decrement the `current_count`. So I believe that case is covered. (But as I remarked above, there is a problem if you create/delete clients all the time.)

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: The error case was self-evident and the comment was incorrect for the common case (no response publishers discovered yet). But, yeah, perhaps that ought to be a separate commit.

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: Oops! Thanks! It’ll probably never get there (the function can’t fail if the reader/writer exist) but that’s no excuse for messing up the type so badly ...

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: Actually, on continued consideration, I could probably make it work without this hack and without adding a "wait for discovery to complete" operation, but I do need to think through the consequences of the changes it would require in Cyclone. It also wouldn't solve the problem if there are multiple matching services.

So it is not yet a given that it'll happen, and it certainly won't be available immediately.

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: +1 to discriminating error conditions from entity absence.

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: > Indeed, I wouldn’t be surprised if simply swapping the creation of the reader and the writer in the client would make the problem disappear in the test setup.

Interesting. Have you tried it?

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: I see. I'm still a bit wary about indefinite waits. Even if it rarely occurs, to timeout and fail on replying would give calling code a chance to do something about it.

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: I think we can all agree that this isn't the correct fix, but a workaround. And @eboasson has been open about it from the get-go. As I see it, the sole purpose of this patch is to reduce the likelihood of a discovery race to impair service performance. 

Considering the current behavior is objectively worst (once we've guaranteed an indefinite wait cannot happen here) and that we're about to release an LTS distro, I'd be very much inclined to land this and document accordingly. I'm sure @jacobperron and @dirk-thomas have their own opinions as well.

jacobperron: IIUC, it seems like a general design flaw with trying to build ROS services on top of DDS, which plagues all of our RMWs.

A workaround sounds okay for now if it's making the situation better; in this particular case, I would want certainty that the logic is not going to cause an indefinite wait. Maybe that means introducing at timeout.

eboasson:
Apologies for the force push ... the original two commits are still there, but I noticed that the subject line mentioned the wrong function and I didn't want to run the risk of that mistake getting merged.

What is new is:
* fb040c5 which changes the request headers on the wire to use `rmw_request_id_t` (so a 128-bit client identifier + sequence number instead of a 64-bit client identifier). The extra room is then used by
* 4c3b8fa to do a precise check whether a reader/writer pair of a *single* server has been matched (in `rmw_service_server_is_available`) and whether the requesting client's response reader has been matched (in `rmw_send_response`).

The mechanism employed is generating unique client/service identifiers (based on the participant GUID), storing these in the reader/writer `USER_DATA` QoS (as a key/value pair, using `serviceid` and `clientid` as keys) and in the GUID part of the `rmw_request_id_t`, and then using the various operations to interrogate information on the matched endpoints to see if there is a match.

It adds overhead and complexity to the previous proposal but as far as I can tell, it doesn't suffer from ever having to wait indefinitely (it is by design immune to adding/removing other clients, and it still handles the case where the client has disappeared).

But clearly this is still a workaround.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -490,6 +508,33 @@ static void get_entity_gid(dds_entity_t h, rmw_gid_t & gid)
   convert_guid_to_gid(guid, gid);
 }
 
+static std::map<std::string, std::vector<uint8_t>> parse_user_data(const dds_qos_t * qos)
+{
+  void * ud;
+  size_t udsz;
+  if (!dds_qget_userdata(qos, &ud, &udsz)) {
+    std::map<std::string, std::vector<uint8_t>> emptymap;
```
hidmic, line 516: @eboasson nit: I'd think any compiler would be smart enough to copy-elide the return value, but

```c++
std::map<std::string, std::vector<uint8_t>> map;
void * ud;
size_t udsz;
if (dds_qget_userdata(qos, &ud, &udsz)) {
  std::vector<uint8_t> udvec(static_cast<uint8_t *>(ud), static_cast<uint8_t *>(ud) + udsz);
  dds_free(ud);
  map = rmw::impl::cpp::parse_key_value(udvec);
}
return map;
```
would make it simpler and slightly clearer.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3020,10 +3046,79 @@ extern "C" rmw_ret_t rmw_wait(
...
+  dds_free(e->type_name);
+  dds_free(e);
+}
+
+static std::unique_ptr<dds_builtintopic_endpoint_t,
+  std::function<void(dds_builtintopic_endpoint_t *)>> get_matched_subscription_data(
+  dds_entity_t writer, dds_instance_handle_t readerih)
+{
+  std::unique_ptr<dds_builtintopic_endpoint_t, std::function<void(dds_builtintopic_endpoint_t *)>>
+  ep(dds_get_matched_subscription_data(writer, readerih), &free_builtintopic_endpoint);
```
hidmic, line 3083: @eboasson nit: 
```suggestion
  ep(dds_get_matched_subscription_data(writer, readerih), free_builtintopic_endpoint);
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3020,10 +3046,79 @@ extern "C" rmw_ret_t rmw_wait(
...
+  return ep;
+}
+
+static std::unique_ptr<dds_builtintopic_endpoint_t,
+  std::function<void(dds_builtintopic_endpoint_t *)>> get_matched_publication_data(
+  dds_entity_t reader, dds_instance_handle_t writerih)
+{
+  std::unique_ptr<dds_builtintopic_endpoint_t, std::function<void(dds_builtintopic_endpoint_t *)>>
+  ep(dds_get_matched_publication_data(reader, writerih), &free_builtintopic_endpoint);
```
hidmic, line 3092: @eboasson nit: 
```suggestion
  ep(dds_get_matched_publication_data(reader, writerih), free_builtintopic_endpoint);
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3020,10 +3046,79 @@ extern "C" rmw_ret_t rmw_wait(
...
+  return ep;
+}
+
+static const std::string request_id_writer_guid_to_string(const rmw_request_id_t & id)
+{
+  std::ostringstream os;
+  os << std::hex;
+  os << std::setw(2) << static_cast<int>(static_cast<uint8_t>(id.writer_guid[0]));
```
hidmic, line 3100: @eboasson why `static_cast<uint8_t>(...)` ? Are you expecting to truncate it?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3597,6 +3793,43 @@ static rmw_ret_t get_topic_name(dds_entity_t endpoint_handle, std::string & name
...
+  {
+    RMW_SET_ERROR_MSG("rmw_service_server_is_available: failed to get reader/writer matches");
+    return RMW_RET_ERROR;
+  }
+  // first extract all service ids from matched readers
+  std::set<std::string> needles;
+  for (auto rdih : rds) {
+    auto rd = get_matched_subscription_data(client.pub->enth, rdih);
+    std::string serviceid;
+    if (rd.get() && get_user_data_key(rd->qos, "serviceid", serviceid)) {
```
hidmic, line 3810: @eboasson nit: 
```suggestion
    if (rd && get_user_data_key(rd->qos, "serviceid", serviceid)) {
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3597,6 +3793,43 @@ static rmw_ret_t get_topic_name(dds_entity_t endpoint_handle, std::string & name
...
+      needles.insert(serviceid);
+    }
+  }
+  if (needles.size() == 0) {
```
hidmic, line 3814: @eboasson nit: 
```suggestion
  if (needles.empty()) {
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3597,6 +3793,43 @@ static rmw_ret_t get_topic_name(dds_entity_t endpoint_handle, std::string & name
...
+    return RMW_RET_OK;
+  }
+  // then scan the writers to see if there is at least one with a service id in the set
+  for (auto wrih : wrs) {
+    auto wr = get_matched_publication_data(client.sub->enth, wrih);
+    std::string serviceid;
+    if (wr.get() &&
```
hidmic, line 3821: @eboasson nit: 
```suggestion
    if (wr &&
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3597,6 +3793,43 @@ static rmw_ret_t get_topic_name(dds_entity_t endpoint_handle, std::string & name
...
+      get_user_data_key(
+        wr->qos, "serviceid",
+        serviceid) && needles.find(serviceid) != needles.end())
+    {
+      *is_available = true;
```
hidmic, line 3826: @eboasson considering asserting that `is_available != nullptr && !*is_available`, which seems to be the implicit assumption in this function.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3124,6 +3221,56 @@ static rmw_ret_t rmw_send_response_request(
...
+  std::vector<dds_instance_handle_t> rds, wrs;
+  if (get_matched_endpoints(service.pub->enth, dds_get_matched_subscriptions, rds) < 0 ||
+    get_matched_endpoints(service.sub->enth, dds_get_matched_publications, wrs) < 0)
+  {
+    RMW_SET_ERROR_MSG("rmw_send_response: failed to get reader/writer matches");
+    return client_present_t::ERROR;
+  }
+  const std::string needle = request_id_writer_guid_to_string(reqid);
+  // if we have matched this client's reader, all is well
+  for (auto rdih : rds) {
```
hidmic, line 3257: @eboasson nit: 
```suggestion
  for (auto & rdih : rds) {
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3124,6 +3221,56 @@ static rmw_ret_t rmw_send_response_request(
...
+    auto rd = get_matched_subscription_data(service.pub->enth, rdih);
+    if (check_client_service_endpoint(rd.get(), "clientid", needle)) {
+      return client_present_t::YES;
+    }
+  }
+  // if not, we should stop waiting if the writer is no longer there,
+  // as that implies the client no longer exists
+  for (auto wrih : wrs) {
```
hidmic, line 3265: @eboasson nit: 
```suggestion
  for (auto & wrih : wrs) {
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3132,10 +3279,27 @@ extern "C" rmw_ret_t rmw_send_response(
...
+    check_for_response_reader(info->service, *request_header)) == client_present_t::MAYBE)
+  {
+    dds_sleepfor(DDS_MSECS(10));
+  }
+  switch (st) {
+    case client_present_t::ERROR:
+      // break instead of returns makes gcc happy
+      break;
+    case client_present_t::MAYBE:
+      return RMW_RET_TIMEOUT;
```
hidmic, line 3296: @eboasson how could `st` ever be `client_present_t::MAYBE` at this point if that wouldn't allow it to leave the loop in the first place?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -309,10 +314,23 @@ struct CddsSubscription : CddsEntity
...
 
+struct client_service_id_t
+{
+  // strangely, the writer_guid in a request id is smaller than the rmw_gid_t
+  // cpplint erroneously complains that
+  //
+  //   sizeof((reinterpret_cast<rmw_request_id_t *>(0))->writer_guid)
+  //
+  // is not a constant, and the 16 is a hard-coded magic number in
+  // rmw_request_id_t ...
```
ivanpauno, line 325: I don't fully follow the comment, but you can also suppress a cpplint warning by using `problematic line of code;  // NOLINT`.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3020,10 +3046,79 @@ extern "C" rmw_ret_t rmw_wait(
 ///////////                                                                   ///////////
 /////////////////////////////////////////////////////////////////////////////////////////
 
+static rmw_ret_t get_matched_endpoints(
+  dds_entity_t h, dds_return_t (* fn)(
+    dds_entity_t h,
+    dds_instance_handle_t * xs, size_t nxs), std::vector<dds_instance_handle_t> & res)
```
ivanpauno, line 3052: nit: defining the pointer to function before highly increases readability:

```cpp
using get_matched_endpoints_fn_t = dds_return_t (* fn)(
    dds_entity_t h,
    dds_instance_handle_t * xs, size_t nxs);
static rmw_ret_t get_matched_endpoints(
  dds_entity_t h, get_matched_endpoints_fn_t fn, std::vector<dds_instance_handle_t> & res)
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3020,10 +3046,79 @@ extern "C" rmw_ret_t rmw_wait(
...
+
+static void free_builtintopic_endpoint(dds_builtintopic_endpoint_t * e)
+{
+  dds_delete_qos(e->qos);
+  dds_free(e->topic_name);
+  dds_free(e->type_name);
+  dds_free(e);
+}
+
+static std::unique_ptr<dds_builtintopic_endpoint_t,
```
ivanpauno, line 3078: similarly here:

```cpp
using BuiltinTopicEndpoint = std::unique_ptr<dds_builtintopic_endpoint_t,
  std::function<void(dds_builtintopic_endpoint_t *)>>;
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3020,10 +3046,79 @@ extern "C" rmw_ret_t rmw_wait(
...
+  ep(dds_get_matched_publication_data(reader, writerih), &free_builtintopic_endpoint);
+  return ep;
+}
+
+static const std::string request_id_writer_guid_to_string(const rmw_request_id_t & id)
+{
+  std::ostringstream os;
+  os << std::hex;
+  os << std::setw(2) << static_cast<int>(static_cast<uint8_t>(id.writer_guid[0]));
+  for (size_t i = 1; i < sizeof(id.writer_guid) - 1; i++) {
```
ivanpauno, line 3101: should be `< sizeof(id.writer_guid)`?
it seems that the last byte won't be copied if not.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3020,10 +3046,79 @@ extern "C" rmw_ret_t rmw_wait(
...
+  }
+  return os.str();
+}
+
+static const std::string csid_to_string(const client_service_id_t & id)
+{
+  std::ostringstream os;
+  os << std::hex;
+  os << std::setw(2) << static_cast<int>(id.data[0]);
+  for (size_t i = 1; i < sizeof(id.data) - 1; i++) {
```
ivanpauno, line 3112: same about the `- 1`, it doesn't sound correct ...

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3132,10 +3279,27 @@ extern "C" rmw_ret_t rmw_send_response(
   RET_NULL(request_header);
   RET_NULL(ros_response);
   CddsService * info = static_cast<CddsService *>(service->data);
-  cdds_request_header_t header;
-  memcpy(&header.guid, request_header->writer_guid, sizeof(header.guid));
-  header.seq = request_header->sequence_number;
-  return rmw_send_response_request(&info->service, header, ros_response);
+  // Block until the response reader has been matched by the response writer (this is a
+  // workaround: rmw_service_server_is_available should keep returning false until this
+  // is a given).
```
ivanpauno, line 3284: Isn't that possible with the addition of the new `user_data`?
Blocking doesn't sound nice

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3597,6 +3793,43 @@ static rmw_ret_t get_topic_name(dds_entity_t endpoint_handle, std::string & name
...
+  std::vector<dds_instance_handle_t> rds, wrs;
+  if (get_matched_endpoints(client.pub->enth, dds_get_matched_subscriptions, rds) < 0 ||
+    get_matched_endpoints(client.sub->enth, dds_get_matched_publications, wrs) < 0)
+  {
+    RMW_SET_ERROR_MSG("rmw_service_server_is_available: failed to get reader/writer matches");
+    return RMW_RET_ERROR;
+  }
+  // first extract all service ids from matched readers
+  std::set<std::string> needles;
+  for (auto rdih : rds) {
```
ivanpauno, line 3807: nit: `const auto &`

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3597,6 +3793,43 @@ static rmw_ret_t get_topic_name(dds_entity_t endpoint_handle, std::string & name
...
+    std::string serviceid;
+    if (rd.get() && get_user_data_key(rd->qos, "serviceid", serviceid)) {
+      needles.insert(serviceid);
+    }
+  }
+  if (needles.size() == 0) {
+    return RMW_RET_OK;
+  }
+  // then scan the writers to see if there is at least one with a service id in the set
+  for (auto wrih : wrs) {
```
ivanpauno, line 3818: nit: `const auto &`

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: There are actually two comments with bad formatting: one that I put in originally: "strangely, the writer_guid in a request id is smaller than the rmw_gid_t". I find it weird that the writer_guid in a `rmw_request_t` is 16 bytes whereas a `rmw_gid_t` has 24 bytes.

The second bit is about lint. I didn't know about `// NOLINT`, but that's a better solution than a hard-coded constant. Thanks!

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: I must have seen too many C function pointer types ... :) but yes, that would be wise

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: Yes, absolutely. (I changed it from 0 .. n-1 to 1..n, except I didn't change the upper bound.) There are not enough services/clients in tests to catch this one. This would have been a source of nasty bugs ...

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: The DDS spec doesn't say a word about what a subscription/publication match event actually guarantees and Cyclone DDS (along with most, perhaps even all, DDS implementations) treats it as a local decision, generating the event when the reader has discovered the writer. It is the most intuitive interpretation of the spec.

However, that doesn't mean it yields the most intuitive behaviour and the absence of a clear specification means there is room to make the "subscription matched" event more useful generating it only once there's proof that the writer has discovered the reader. But doing that depends in part on the behaviour of the writer implementation, and so interoperability at DDS level needs to be thought through.

I am considering doing all that in Cyclone, but there's no way I can get that released soon enough. My 0.6.0 release is already slipping because of the release process' dependency on very busy volunteers ...

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: True, and it would be better to combine it with the `ERROR` case. I usually go out of my way to treat `enum`s as enumerated types and cover all the cases in switches over them. But C & C++ semantics don't guarantee that one can assign only defined values to an object of an enumerated type, and what then happens is compiler dependent.

Clang doesn't warn if you cover all cases of an enum, presumably on the assumption that you treat the enum as an enum (or perhaps it proves that no other values ever get assigned to it). Gcc warns, presumably because technically other values are legal too and, again presumably, it doesn't do the analysis to prove that no other values ever get assigned to it. As the comment suggests, there used to be a return in line 3294. With that, the `MAYBE` case makes a bit more sense. Now, it just looks silly.

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: Yeah, at some point I was confused and thought that that extra information could help to solve the problem, but it can't.

I think that [this comment](https://github.com/ros2/rmw_cyclonedds/pull/187#discussion_r428660615) should be addressed, to avoid the remote chance of the wait taking forever.

I also don't quite like the idea of introducing blocking behavior, but if other people approve I'm fine with the workaround for now.

eboasson:
@hidmic and @ivanpauno thanks for reviewing (and catching a few bugs!). I'm fine with all the nitpicks. I'll sort them out tomorrow, but I thought it might make sense to quickly respond to the comments given the time zone differences and the urgency.

@hidmic, regarding your two overall questions:

> I wonder though, is this change backwards compatible (i.e. it can work with a peer participant running without this patch, with the known potential races)?

It isn't: firstly, the wire representation of the request id changed from the quick hack of yore to the kind-of sensible `rmw_request_id`. While that change is convenient, it is not strictly necessary: the alternative is to convert the `publication_handle` of the request in `rmw_take_request` to the client id, and then associate the `rmw_request_id_t` with the client id internally. Then `rmw_send_response` can use the request id to lookup the GUID, and so on. It is messy ... and it does assume that all requests result in exactly one response.

Secondly, it now only matches services/clients that have these identifiers. Making it backwards compatible requires treating an "unidentified" reader and writer pair as sufficient in `rmw_service_server_is_available` and never waiting in `rmw_send_response`.

> Also, won't this change compromise cross-vendor communication?

That doesn't work anyway ... different wire representations, use of vendor-specific tricks ...

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -490,6 +508,33 @@ static void get_entity_gid(dds_entity_t h, rmw_gid_t & gid)
   convert_guid_to_gid(guid, gid);
 }
 
+static std::map<std::string, std::vector<uint8_t>> parse_user_data(const dds_qos_t * qos)
```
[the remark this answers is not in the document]
eboasson: `parse_key_value` from `rmw` uses `std::map` so I think it is better to leave this as-is

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: That cast is there to avoid sign extension (`id.writer_guid` is an array of `int8_t` ...)

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: given that `dds_instance_handle_t` is just a 64-bit integer, I don't understand why using a reference would be an improvement

eboasson:
I believe the first of the two commits addresses all the comments regarding small details (unless noted otherwise — if I am mistaken, I'll be happy to change them after all, as I don't know the C++ idiom very well).

The second addresses backwards compatibility. It also happens to make the code a bit simpler, too. The changes in that commit outside `rmw_node.cpp` are a simple reverting of fb040c5.

There is still the matter of the desirability of blocking in `rmw_send_response`. I don't particularly like this workaround, but having it ready at least puts us in a situation where we can decide what to do:
* do as in dashing/eloquent (i.e., just do 0cf065d)
* do as in dashing/eloquent for matching but do add the client/service identifiers (head, except for the blocking)
* use this workaround
* if desired, add a timeout and expect services to handle that case (that's easy enough)

Note that I don't think adding a timeout make sense: the duration would be completely arbitrary and the service implementations would then have be modified to deal with it.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3133,9 +3284,34 @@ extern "C" rmw_ret_t rmw_send_response(
...
+    "request header size assumptions not met");
+  memcpy(
+    static_cast<void *>(&header.guid), static_cast<const void *>(request_header->writer_guid),
+    sizeof(header.guid));
+  memcpy(
+    static_cast<void *>(&reqwrih),
+    static_cast<const void *>(request_header->writer_guid + sizeof(header.guid)), sizeof(reqwrih));
   header.seq = request_header->sequence_number;
-  return rmw_send_response_request(&info->service, header, ros_response);
+  // Block until the response reader has been matched by the response writer (this is a
```
hidmic, line 3298: @eboasson mind to add a TODO for a proper fix? I've opened #191 to track that work.

wjwwood:
> @wjwwood @ivanpauno @jacobperron I'd like your approval as well before merging anything.

Based on the discussions we've had about this and the corresponding Fast-RTPS pull request, I'm good with it. I'll have another look, but the code changes also look good.

hidmic:
@eboasson I think Windows CI compilation issues stem from `client_present_t::ERROR` enum value clashing with the ERROR macro that comes with `windows.h`.

jacobperron:
Given that we're scheduled to release this week, I'd be more comfortable holding this PR for the first patch release to give us more time for testing. This means holding until the end of this week.

eboasson:
I think this covers all the remarks. It has a conflict in `rmw_node.cpp` because the merged #190 uses the exact same code as dashing/eloquent, whereas this one had equivalent code in its place. If everyone agrees it is ok (which is not necessarily the same as it being ok to merge it straightaway, because of the Foxy release), I'll fix the conflict and squash the changes.

hidmic:
@eboasson +1, feel free to rebase.

hidmic:
Once Windows' CI is back, I think this is ok to go.

hidmic:
Alright, going in.

ros-discourse:

https://discourse.ros.org/t/new-packages-for-foxy-fitzroy-2020-07-23/15570/2

ros-discourse:

https://discourse.ros.org/t/new-packages-and-patch-release-for-ros-2-foxy-fitzroy-2020-08-07/15818/1

PR #195, opened 2020-06-19
Add space between 'ROS' and '2'

christophebedard:
For the proper use of "ROS 2" :grin:

I modified more than just the README, hope you don't mind.

Also, if someone could fix the repo description, that would be nice

> **ROS2** RMW layer for Eclipse Cyclone DDS

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,4 +1,4 @@
-name: ROS2 CI
+name: ROS 2 CI
 on: [push, pull_request]
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


christophebedard:
> I wonder, though: why not a non-breaking space?

For simplicity, maybe. Or perhaps it just wasn't considered. I agree that it might be worth mentioning/suggesting that to OR, but for code/text like this, I don't think we'd want to use a non-breaking space.

eboasson:
Looks like I forgot to merge it ...

PR #198, opened 2020-06-22
Avoid leaking init options in context.

hidmic:
Spin off from #196.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1125,19 +1125,24 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t 
 
   context->instance_id = options->instance_id;
   context->implementation_identifier = eclipse_cyclonedds_identifier;
-
-  context->impl = new(std::nothrow) rmw_context_impl_t();
-  if (nullptr == context->impl) {
-    return RMW_RET_BAD_ALLOC;
-  }
+  context->impl = nullptr;
 
   ret = rmw_init_options_copy(options, &context->options);
   if (RMW_RET_OK != ret) {
-    if (RMW_RET_OK != rmw_init_options_fini(&context->options)) {
-      RMW_SAFE_FWRITE_TO_STDERR(
-        "'rmw_init_options_fini' failed while being executed due to '"
-        RCUTILS_STRINGIFY(__function__) "' failing.\n");
-    }
+    goto fail;
+  }
+
+  context->impl = new(std::nothrow) rmw_context_impl_t();
+  if (nullptr == context->impl) {
+    ret = RMW_RET_BAD_ALLOC;
+    goto fail;
+  }
+  return ret;
+fail:
+  if (RMW_RET_OK != rmw_init_options_fini(&context->options)) {
+    RMW_SAFE_FWRITE_TO_STDERR(
+      "'rmw_init_options_fini' failed while being executed due to '"
+      RCUTILS_STRINGIFY(__function__) "' failing.\n");
   }
   return ret;
 }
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1125,19 +1125,21 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t
 
   context->instance_id = options->instance_id;
   context->implementation_identifier = eclipse_cyclonedds_identifier;
-  context->impl = nullptr;
 
-  if ((ret = rmw_init_options_copy(options, &context->options)) != RMW_RET_OK) {
-    return ret;
-  }
```
ivanpauno, line 1132: I don't think that changing the order was needed

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1125,19 +1125,21 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t
...
 
-  context->impl = impl;
-  return RMW_RET_OK;
+  ret = rmw_init_options_copy(options, &context->options);
+  if (RMW_RET_OK != ret) {
+    if (RMW_RET_OK != rmw_init_options_fini(&context->options)) {
+      RMW_SAFE_FWRITE_TO_STDERR(
+        "'rmw_init_options_fini' failed while being executed due to '"
+        RCUTILS_STRINGIFY(__function__) "' failing.\n");
+    }
```
ivanpauno, line 1140: leaking `context->impl` in this case?
When returning an error everything should be cleaned up

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: It made easier, but it's not necessarily more correct. Changed.

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: Good point.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1127,17 +1127,24 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t
...
+  if (nullptr == context->impl) {
+    ret = RMW_RET_BAD_ALLOC;
+    goto fail;
   }
-
-  context->impl = impl;
-  return RMW_RET_OK;
+  return ret;
+fail:
+  if (RMW_RET_OK != rmw_init_options_fini(&context->options)) {
```
eboasson, line 1142: Is it safe to call `fini` if `init` failed? Just checking ...

ivanpauno: good point, I think it's not.
@hidmic maybe just returning an error in line 1132.

hidmic: Well, arguably it should not and either `rmw_init_options_init()` or `rmw_init_options_copy()` should guarantee output options remain uninitialized. Reality is that such guarantee is not documented in `rmw` API and most implementations do not really check. `rmw_cyclonedds_cpp` isn't the exception. Thus the extra `fini`, which judging by the code isn't unsafe (though again, not necessarily correct nor ideal).

Alternatively, we can establish that guarantee and change implementations to abide to it. Thoughts?

ivanpauno: > Alternatively, we can establish that guarantee and change implementations to abide to it. Thoughts?

:+1:
If I call `*_init()` or `*_copy()` and the function fails, I expect it to cleanup all the resources it created (and not having to do it manually).
In the case of `*_copy(src, dst)`, I also expect the function to no modify `*dst` if it failed to do the copy.

It's the more usual pattern in C code IMO.

hidmic: See https://github.com/ros2/rmw/pull/244.

hidmic:
This has been superseded by #202.

PR #200, opened 2020-06-23
Ensure compliant init options API implementations.

hidmic:
Complying with https://github.com/ros2/rmw/pull/244.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -494,11 +494,9 @@ extern "C" rmw_ret_t rmw_init_options_fini(rmw_init_options_t * init_options)
   rcutils_allocator_t * allocator = &init_options->allocator;
   RCUTILS_CHECK_ALLOCATOR(allocator, return RMW_RET_INVALID_ARGUMENT);
 
+  allocator->deallocate(init_options->enclave, allocator->state);
   rmw_ret_t ret = rmw_security_options_fini(&init_options->security_options, allocator);
-  if (RMW_RET_OK == ret) {
-    allocator->deallocate(init_options->enclave, allocator->state);
-    *init_options = rmw_get_zero_initialized_init_options();
-  }
+  *init_options = rmw_get_zero_initialized_init_options();
   return ret;
 }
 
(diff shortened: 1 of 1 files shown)


hidmic:
CI up to `rcl` and `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11262/)

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -457,40 +461,45 @@ extern "C" rmw_ret_t rmw_init_options_copy(const rmw_init_options_t * src, rmw_i
...
-  allocator.deallocate(init_options->enclave, allocator.state);
-  rmw_security_options_fini(&init_options->security_options, &allocator);
-  *init_options = rmw_get_zero_initialized_init_options();
-  return RMW_RET_OK;
+  rcutils_allocator_t * allocator = &init_options->allocator;
+  RCUTILS_CHECK_ALLOCATOR(allocator, return RMW_RET_INVALID_ARGUMENT);
+
+  rmw_ret_t ret = rmw_security_options_fini(&init_options->security_options, allocator);
+  if (RMW_RET_OK == ret) {
+    allocator->deallocate(init_options->enclave, allocator->state);
```
ivanpauno, line 499: see https://github.com/ros2/rmw_fastrtps/pull/399#discussion_r444441126

hidmic:
> Ignoring the gap between theory and practice, finalization failures are only possible if the environment was changed in invalid ways (e.g., DRAM failure, application deliberately freeing something it didn't own, etc.). In my view, that means "finalization never fails" (and therefore aborts on error) is a generally preferable over one where the state becomes undefined on error.

Yeah, there's that path too. But I think for finalization APIs, the `rmw_ret_t` vs. `void` return type ship has sailed already. @wjwwood for comments.

wjwwood:
Yeah we discussed this before but essentially we saw the return codes as a way to more easily assert how it failed in tests, and because the caller can always choose to abort anyways.

hidmic:
One last sanity CI run after API contract updates:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11281/)

PR #202, opened 2020-06-24
Ensure compliant init/shutdown API implementations.

hidmic:
To comply with https://github.com/ros2/rmw/pull/243.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -50,7 +50,7 @@
 
 #include "fallthrough_macro.hpp"
 #include "Serialization.hpp"
-#include "rmw/impl/cpp/atexit.hpp"
+#include "rcpputils/scope_exit.hpp"
 #include "rmw/impl/cpp/macros.hpp"
 #include "rmw/impl/cpp/key_value.hpp"
 
@@ -1140,10 +1140,10 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t 
     return RMW_RET_INVALID_ARGUMENT;
   }
 
-  rmw::impl::cpp::atexit cleanup{[context]() {
+  auto cleanup = rcpputils::make_scope_exit([context]() {
       delete context->impl;
       *context = rmw_get_zero_initialized_context();
-    }};
+    });
 
   context->instance_id = options->instance_id;
   context->implementation_identifier = eclipse_cyclonedds_identifier;
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cassert>
+#include <cstring>
 #include <mutex>
 #include <unordered_map>
 #include <unordered_set>
@@ -1140,11 +1142,10 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t 
     return RMW_RET_INVALID_ARGUMENT;
   }
 
-  auto cleanup = rcpputils::make_scope_exit(
-    [context]() {
-      delete context->impl;
-      *context = rmw_get_zero_initialized_context();
-    });
+  const rmw_context_t zero_context = rmw_get_zero_initialized_context();
+  assert(0 == std::memcmp(context, &zero_context, sizeof(rmw_context_t)));
+  auto restore_context = rcpputils::make_scope_exit(
+    [context, &zero_context]() {*context = zero_context;});
 
   context->instance_id = options->instance_id;
   context->implementation_identifier = eclipse_cyclonedds_identifier;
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1117,68 +1118,86 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t
...
+    return RMW_RET_INVALID_ARGUMENT;
+  }
 
   if (options->domain_id >= UINT32_MAX && options->domain_id != RMW_DEFAULT_DOMAIN_ID) {
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_cyclonedds_cpp", "rmw_create_node: domain id out of range");
     return RMW_RET_INVALID_ARGUMENT;
   }
 
+  rmw::impl::cpp::atexit cleanup{[context]() {
```
eboasson, line 1143: This cleanup routine is written on the assumption that nothing that may require cleaning up will happen following the allocation of `context->impl`, which is true now. I am a little concerned that this model increases the risk of failure to cleanup correctly when additional resource claims are added in the initialization code at a later stage.

I don't perceive that "increase of risk" (if indeed there is any) to be large. To me, the exact way in which this arranges the cleanup doesn't really add anything of value, but it's not like I am against it. Clearly, you have a different sense of the value or you wouldn't have put in the effort. On balance, that makes me in favour of merging it, and I am raising this merely because it could be a perspective you hadn't taken into account yet.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1117,68 +1118,86 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t
...
   context->instance_id = options->instance_id;
   context->implementation_identifier = eclipse_cyclonedds_identifier;
-  context->impl = nullptr;
 
-  if ((ret = rmw_init_options_copy(options, &context->options)) != RMW_RET_OK) {
-    return ret;
+  context->impl = new (std::nothrow) rmw_context_impl_t();
+  if (nullptr == context->impl) {
+    RMW_SET_ERROR_MSG("failed to allocate context impl");
+    return RMW_RET_BAD_ALLOC;
```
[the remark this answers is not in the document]
hidmic: It is a no-op, per https://en.cppreference.com/w/cpp/language/delete:

```
If expression evaluates to a null pointer value, no destructors are called, and the deallocation function may or may not be called (it's unspecified), but the default deallocation functions are guaranteed to do nothing when passed a null pointer.
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: > To me, the exact way in which this arranges the cleanup doesn't really add anything of value, but it's not like I am against it. 

The sole point of using RAII for cleanup here is to de-duplicate cleanup code, with a preference for stack unwinding in lieu of `goto` statements.

rmw_cyclonedds_cpp/src/rmw_node.cpp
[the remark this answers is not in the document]
eboasson: Thanks!

rmw_cyclonedds_cpp/src/rmw_node.cpp
eboasson: I like the RAII bit very much (one of the things that is so much better in C++ than in C ...), but to me it would then be more logical to do set it up immediately after the allocation happened. Anyway, as I said, I'm ok with it, especially after your reply.

hidmic: > but to me it would then be more logical to do set it up immediately after the allocation happened.

[Per API contract](https://github.com/ros2/rmw/pull/243/files#diff-9fe3987dc902dde2f416ab8fa78d0ed4R66) I must ensure that the context remains zero initialized if initialization fails not due to invalid use. So I have to have that block there to zero re-initialize it, and figured a no-op `delete` didn't make any harm. But yeah, we could split those two statements into separate `rcpputils::scope_exit` instances.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1117,68 +1118,87 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t
...
+    return RMW_RET_INVALID_ARGUMENT;
+  }
 
   if (options->domain_id >= UINT32_MAX && options->domain_id != RMW_DEFAULT_DOMAIN_ID) {
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_cyclonedds_cpp", "rmw_create_node: domain id out of range");
     return RMW_RET_INVALID_ARGUMENT;
   }
 
+  auto cleanup = rcpputils::make_scope_exit(
```
ivanpauno, line 1143: see https://github.com/ros2/rmw_connext/pull/432#pullrequestreview-438615709

hidmic: See https://github.com/ros2/rmw_connext/pull/432/commits/c1eb006f475586bd64392eb5f85a469d40648762

hidmic:
CI up to `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11320/)

hidmic:
Going in, thanks for all the reviews!

PR #208, opened 2020-07-17
Set context actual domain id

ivanpauno:
See https://github.com/ros2/rcl/pull/718.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -287,7 +287,7 @@ struct rmw_context_impl_t
   // Initializes the participant, if it wasn't done already.
   // node_count is increased
   rmw_ret_t
-  init(rmw_init_options_t * options);
+  init(rmw_init_options_t * options, size_t domain_id);
 
   // Destroys the participant, when node_count reaches 0.
   rmw_ret_t
@@ -926,7 +926,7 @@ rmw_ret_t configure_qos_for_security(
 }
 
 rmw_ret_t
-rmw_context_impl_t::init(rmw_init_options_t * options)
+rmw_context_impl_t::init(rmw_init_options_t * options, size_t domain_id)
 {
   std::lock_guard<std::mutex> guard(initialization_mutex);
   if (0u != this->node_count) {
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1154,6 +1154,8 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t
 
   context->instance_id = options->instance_id;
   context->implementation_identifier = eclipse_cyclonedds_identifier;
+  context->actual_domain_id =
+    RMW_DEFAULT_DOMAIN_ID != options->domain_id ? options->domain_id : 0u;
```
hidmic, line 1158: @ivanpauno [this line](https://github.com/ros2/rmw_cyclonedds/pull/208/files#diff-d03c99cafce317e1a93de46e787e0235R944) has to change too.

ivanpauno: @hidmic can you clarify which line you're referring to?
The hyperlink redirects to this same comment.

ivanpauno: Ok, I think I guessed which line was: https://github.com/ros2/rmw_cyclonedds/pull/208/commits/a8b45bfd2ccf0639bf769506422e3e38b63dcec3

hidmic: Uh, that's so weird. It's almost like if Github's line hyperlinks within a PR diff are broken. Anyhow, you guessed right :)

ivanpauno: > It's almost like if Github's line hyperlinks within a PR diff are broken

Yeah, links within a PR diff are broken in my experience after the review is sent.
Opening the file and getting the permalink always works correctly.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1147,13 +1145,14 @@ extern "C" rmw_ret_t rmw_init(const rmw_init_options_t * options, rmw_context_t
     return RMW_RET_INVALID_ARGUMENT;
   }
 
-  const rmw_context_t zero_context = rmw_get_zero_initialized_context();
-  assert(0 == std::memcmp(context, &zero_context, sizeof(rmw_context_t)));
```
hidmic, line 1151: @ivanpauno unrelated?

ivanpauno: See https://github.com/ros2/rmw_connext/pull/443#discussion_r458314262

PR #210, opened 2020-07-24
Ensure compliant publisher API

hidmic:
Precisely what the title says. Connected to ros2/rmw#252.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1588,7 +1588,6 @@ static dds_qos_t * create_readwrite_qos(
   dds_qset_writer_data_lifecycle(qos, false); /* disable autodispose */
   switch (qos_policies->history) {
     case RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT:
-    case RMW_QOS_POLICY_HISTORY_UNKNOWN:
     case RMW_QOS_POLICY_HISTORY_KEEP_LAST:
       if (qos_policies->depth == RMW_QOS_POLICY_DEPTH_SYSTEM_DEFAULT) {
         dds_qset_history(qos, DDS_HISTORY_KEEP_LAST, 1);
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2076,12 +2134,32 @@ extern "C" rmw_ret_t rmw_destroy_publisher(rmw_node_t * node, rmw_publisher_t *
...
+    if (RMW_RET_OK != publish_ret) {
+      error_state = *rmw_get_error_state();
+      ret = publish_ret;
+      rmw_reset_error();
     }
   }
-  return destroy_publisher(publisher);
+
+  rmw_ret_t inner_ret = destroy_publisher(publisher);
+  if (RMW_RET_OK != inner_ret) {
```
hidmic, line 2147: I don't like having to do this to ensure the first error that occurs is the one that pops up and that no subsequent errors goes unnoticed.

If we had a way to tell `rcutils` that we're in error handling or finalization/cleanup mode so that it doesn't complain about error states being overwritten (and prints the new one to `stderr`, or aborts the program, or whatever), this code would simpler (and a bit more neat). What do reviewers think?

ivanpauno: The current logic isn't super nice, but isn't wrong too.
Maybe directly logging all destruction errors and setting a generic error state at the end is a better idea.

hidmic: > The current logic isn't super nice, but isn't wrong either.

No, it's not incorrect, but it's... messy.

> Maybe directly logging all destruction errors and setting a generic error state at the end is a better idea.

We could do that, but that doesn't cover errors during cleanup. Besides, it's important to know what the actual error was to do something about it (or debug it), nevermind leaks. At any rate, I don't mind doing this. I just figured it'd much easier if `rcutils` took care of redirecting all `RCUTILS_SET_ERROR_MSG()` for a given scope. @clalancette @wjwwood @jacobperron for feedback too.

ivanpauno: > We could do that, but that doesn't cover errors during cleanup

Do you mean the cleanup that a function does when it fails?
In that case it would be good to set an error message about what failed, and log all cleanup failures.
So logging is fine in that case too,

> I just figured it'd much easier if rcutils took care of redirecting all RCUTILS_SET_ERROR_MSG() for a given scope

How?

hidmic: For instance, having a flag in TLS that client code can toggle. For C++ implementations we may be able to use RAII for easier control of the error handling scope.

ivanpauno: > For instance, having a flag in TLS that client code can toggle

That's ok, but I would prefer just having a simple policy.
IMO, directly logging all destruction errors is fine.

hidmic: Ok, that's fair. As I said elsewhere, I wonder about the usefulness of a return code if we log everything.

ivanpauno: The return code would still be useful, as the caller can know that sth failed (it doesn't matter if we return the first error code or the last one).
The error string won't be useful if we log everything, yes.

gbiggs: > I wonder about the usefulness of a return code if we log everything

Isn't the purpose of the return code so that the calling function knows something went wrong, and optionally if there is enough information in the return code to then make an intelligent decision about what to do about it? The logging is more for after-the-fact debugging so having everything logged doesn't seem bad to me.

Having an error string would also still be useful for when logging is turned down, I think.

eboasson: Why not simply pass an RMW error state object around inside the implementation, with a setter that implements the behaviour of returning the first error and logging the rest? That's basically the pattern you have here with `ret` and `inner_ret`, but it'd remove the messiness from the code.

hidmic: > and optionally if there is enough information in the return code to then make an intelligent decision about what to do about it?

IMHO that's the whole point of having different, documented return codes, vs. a bool. 

> Why not simply pass an RMW error state object around inside the implementation, with a setter that implements the behaviour of returning the first error and logging the rest?

Yeah, I was trying not to (re)build this behavior for each RMW implementation, but to make it widely available. I'll probably push this as-is until we've reached consensus on how to do this nicely.

hidmic:
CI up `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11630/)

hidmic:
Ok, CI's green and reviewers are happy. Going in!

hidmic:
Confirmed locally that this patch introduces no _new_ leaks. `valgrind` does point out that we may be leaking through `create_message_typesupport`. We can follow-up in a separate PR.

PR #214, opened 2020-07-31
Ensure compliant subscription API.

hidmic:
Depends on https://github.com/ros2/rmw/pull/256.


hidmic:
CI up to `test_rmw_implementation` and `rcl`, against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11736/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/6817/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9574/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/11676/)

hidmic:
>  I just realised that this prioritizes the error code from publishing the ros discovery topic (a side effect) over the error code from actually destroying the subscription (the primary operation) if both error, and that might be somewhat counterintuitive.

Hmm, that's a very good point. And I'm inclined to change it.

PR #217, opened 2020-08-10
Fix bad conditional in rmw_serialize().

hidmic:
A tiny fix, to make sure `ret` actually gets `rmw_serialized_message_resize()` return value.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1383,13 +1383,13 @@ extern "C" rmw_ret_t rmw_serialize(
   const rosidl_message_type_support_t * type_support,
   rmw_serialized_message_t * serialized_message)
 {
-  rmw_ret_t ret;
   try {
     auto writer = rmw_cyclonedds_cpp::make_cdr_writer(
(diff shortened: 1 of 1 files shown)


rotu:
This is certainly clearer, but I believe it was correct before too. `if(x = f() != y)` will *always* call `f` and assign the value to `x`, no matter how the comparison turns out.

rotu:
@hidmic YOU WERE RIGHT. `==` has higher operator precedence than `=` and so ret would get a `bool` value, not the intended `rmw_ret_t`.

hidmic:
Indeed. I stumbled upon this while working on https://github.com/ros2/rmw_implementation/pull/118.

hidmic:
I don't count my own anymore.

hidmic:
Piggybacking this fix for CI up to `test_rmw_implementation` and `rmw`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11885/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6930/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9682/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11822/)

PR #221, opened 2020-08-20
include rcutils/macros for stringify macro

rotu:


rotu:
This is the wrong solution. Closing.

PR #222, opened 2020-08-20
Bump CI version

rotu:
Bring CI up to date and add Ubuntu 20.04, the current target platform for ROS Rolling.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -5,8 +5,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-          rosdistro: [master]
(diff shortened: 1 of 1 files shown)


eboasson (approved):
LGTM, thanks for figuring out how/what to update. The macOS and Windows failures are a bit weird considering that the same sources are considered fine on Linux, so I’m not even sure it is an issue with the CI config or with the source. I’d say this is a welcome improvement.

rotu:
@eboasson The macOS and Windows failures seem due (at least in part) to a problem in the underlying CI action: https://github.com/ros-tooling/action-ros-ci/issues/312

PR #223, opened 2020-08-20
Ensure compliant matched pub/sub count API.

hidmic:
Connected to https://github.com/ros2/rmw/pull/262.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -2047,16 +2047,22 @@ extern "C" rmw_ret_t rmw_publisher_count_matched_subscriptions(
   const rmw_publisher_t * publisher,
   size_t * subscription_count)
 {
-  RET_NULL(publisher);
-  RET_WRONG_IMPLID(publisher);
+  RMW_CHECK_ARGUMENT_FOR_NULL(publisher, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    publisher,
+    publisher->implementation_identifier,
+    eclipse_cyclonedds_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RMW_CHECK_ARGUMENT_FOR_NULL(subscription_count, RMW_RET_INVALID_ARGUMENT);
+
   auto pub = static_cast<CddsPublisher *>(publisher->data);
   dds_publication_matched_status_t status;
   if (dds_get_publication_matched_status(pub->enth, &status) < 0) {
     return RMW_RET_ERROR;
-  } else {
-    *subscription_count = status.current_count;
-    return RMW_RET_OK;
   }
+
+  *subscription_count = status.current_count;
+  return RMW_RET_OK;
 }
 
 rmw_ret_t rmw_publisher_assert_liveliness(const rmw_publisher_t * publisher)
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2042,7 +2042,13 @@ extern "C" rmw_ret_t rmw_publisher_count_matched_subscriptions(
   const rmw_publisher_t * publisher,
   size_t * subscription_count)
 {
-  RET_WRONG_IMPLID(publisher);
```
clalancette, line 2045: While this is different from the other RMW implementations, I'm thinking we should probably keep using `RET_WRONG_IMPLID` and change the return value of the macro to `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`.

That is, this rmw uses `RET_WRONG_IMPLID` in a lot of places, and now we'll have different return values at different entry points.

What do you think?

hidmic: I agree with changing `RET_WRONG_IMPLID` to return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, but I'm somewhat inclined towards standardizing how such errors are handled. Why having two macros to do the same thing?

clalancette: > I agree with changing `RET_WRONG_IMPLID` to return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, but I'm somewhat inclined towards standardizing how such errors are handled. Why having two macros to do the same thing?

I agree, but I think it is more important how the API reacts than the internal implementation details.  I'm still worried that we are only changing part of the API to return a different value.  With that in mind, my suggestion here is that this PR updates `RET_WRONG_IMPLID` to return `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`, but leaves this call to `RET_WRONG_IMPLID` in place.  Then in a follow-up we can replace all uses of `RET_WRONG_IMPLID` with `RMW_CHECK_TYPE_IDENTIFIERS_MATCH`.  Thoughts about that?

hidmic: I'd rather keep this PR small and in scope (i.e. I'd rather not touch the return code of every function that uses `RET_WRONG_IMPLID` here). 

What if we put this one on hold, phase out `RET_WRONG_IMPLID` in a separate PR, and then rebase this one on top?

eboasson (approved):
LGTM. I agree with the conclusion that cleaning up the `RET_WRONG_IMPLID` in a separate PR makes sense. Thanks @hidmic for all the work you are doing to clean things up a bit!

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: Alright, ended up changing `RET_WRONG_IMPLID` only. See https://github.com/ros2/rmw_cyclonedds/pull/226. It was the simplest change, as the implementation isn't consistent about what to return when an argument is `nullptr` either. 

I'll standardize it over time as I keep adding tests, but I want to avoid blanket PRs now.

hidmic:
Alright, PR rebased. PTAL!

hidmic:
CI up to `test_rmw_implementation`, against all Tier 1 RMW implementations:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12070/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7095/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9844/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12001/)

hidmic:
Alright, going in!

PR #227, opened 2020-08-27
Fix context impl cleanup.

hidmic:
Precisely what the title says. Otherwise, SIGSEGV ensues if a subsequent (re)initialization fails halfway through.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1097,17 +1097,22 @@ rmw_context_impl_t::clean_up()
   common.graph_cache.clear_on_change_callback();
   if (common.graph_guard_condition) {
     destroy_guard_condition(common.graph_guard_condition);
+    common.graph_guard_condition = nullptr;
   }
   if (common.pub) {
     destroy_publisher(common.pub);
+    common.pub = nullptr;
   }
   if (common.sub) {
     destroy_subscription(common.sub);
+    common.sub = nullptr;
   }
   if (ppant > 0 && dds_delete(ppant) < 0) {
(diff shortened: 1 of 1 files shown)


ivanpauno:
> Otherwise, SIGSEGV ensues if a subsequent (re)initialization fails halfway through.

Do we provide that guarantee?

Just to double check, this is what I understand from your comment:
```
rmw_context_t context = rmw_create_context(...);
rmw_destroy_context(context); // this one fails
rmw_destroy_context(context); // retrying after failing segfaults
```

hidmic:
> Just to double check, this is what I understand from your comment:

I meant subsequent context _impl_ (perhaps that's the bit that's missing) (re)initialization failure **after** a preceding (often implicit) cleanup may result in SIGSEGV. That can happen if, within `rmw_create_node`, [`rmw_context_impl_t::init()`](https://github.com/ros2/rmw_cyclonedds/blob/416c01d45dba80bd713d38fd4cc6360efb7448f5/rmw_cyclonedds_cpp/src/rmw_node.cpp#L937) fails halfway through twice in a row.

ivanpauno:
> I meant subsequent context impl (perhaps that's the bit that's missing) (re)initialization failure after a preceding (often implicit) cleanup may result in SIGSEGV. That can happen if, within rmw_create_node, rmw_context_impl_t::init() fails halfway through twice in a row.

I see, it's super weird though.

hidmic:
CI up to `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12124/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7121/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9873/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12029/)

PR #229, opened 2020-08-29
Specify default cmake config type

rotu:


rotu:
I thought this might fix the cppcheck issue but it didn't. closing.

PR #228, opened 2020-08-28
Use correct functions to resize and get an item, avoiding memory leaks in typesupport code

iuhilnehc-ynos:
This PR is related to https://github.com/ros2/rmw_cyclonedds/issues/219

NOTE: I used multiple commits in a PR because it'll easy for me to revert some of them if they're unnecessary.

Signed-off-by: Chen.Lihui <lihui.chen@sony.com>

PR #230, opened 2020-08-30
Remove public declarations

rotu:
This package is only intended to be accessed through `rmw_implementation`. As such none of the headers in this package were intended to be part of the public interface. The link interface of this package is that provided by `rmw_implementation`.


eboasson (approved):
Good point @rotu . I'd like to avoid requiring unnecessary rework on #228, which looks about ready to go in. So let's wait for that one before merging this one.

rotu:
@eboasson Any objection to merging this in now?

eboasson:
Hi @rotu, if everything checks out (GitHub says there's a conflict now, hence my including some weasel words 😄), I'm all for merging it in.

PR #232, opened 2020-08-30
Run cppcheck only on linux

rotu:


rotu:
@piraka9011 @cottsay any idea why this is failing? I feel like it should be super obvious but I’m not seeing it. https://github.com/ros2/rmw_cyclonedds/runs/1047174668#step:5:6627

piraka9011:
Maybe you need to add the packages to the `CMakeLists.txt` as well?

rotu:
> Maybe you need to add the packages to the `CMakeLists.txt` as well?
> Although it looks like it's [already there](https://github.com/ros2/rmw_cyclonedds/blob/e04b4510db520ef7a23c592519c523d9de68573a/rmw_cyclonedds_cpp/CMakeLists.txt#L280).

Yeah already added.

> Maybe `ament_cmake_ros` doesn't pull in all the linters?

Maybe. But it looks like I added all that is required according to https://github.com/ament/ament_lint/blob/master/ament_lint_auto/doc/index.rst

It seems like:

1. ament_lint_auto isn’t properly pulling in ament_cppcheck dependencies on linux https://github.com/ros2/rmw_cyclonedds/pull/232/checks?check_run_id=1047174882#step:5:6984

2. ament_lint_auto *is* pulling in ament_cppcheck on Mac and Win despite me not telling it to https://github.com/ros2/rmw_cyclonedds/pull/232/checks?check_run_id=1047174919#step:5:6239

PR #233, opened 2020-09-02
Change wrong use of %ld to print std::size_t to %zu

svengit:
Using %ld to print a std::size_t leads to compiler warnings and will lead
to wrong behavior whenever the specific std::size_t is bigger than LONG_MAX

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -2572,7 +2572,7 @@ static rmw_ret_t rmw_take_seq(
 
   if (count > (std::numeric_limits<uint32_t>::max)()) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
-      "Cannot take %ld samples at once, limit is %d",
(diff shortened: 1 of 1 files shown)


eboasson (commented):
Good point, thanks!

I am a bit surprised that it doesn't warn about the `%d` as well. I'd expect the type of `std::numeric_limits<uint32_t>::max)()` to be `uint32_t` while `%d` expects an `int`, and those
are not necessarily compatible. In C the correct way to deal with that is to replace `%d"` with `"%" PRIu32` (from `inttypes.h`), but I am not sure about these macros in C++.

Perhaps someone who knows the details of how these formatting strings are to be used properly in C++ can comment?

svengit:
I assume you do the same in C++. As far as I know if you really want to burden yourself with format strings in C++ you just use the C headers/functions, because the "C++" way to do it is ostreams or std::format (finally)

eboasson:
With apologies for the very slow response. #253 takes care of the `%d` detail mentioned in the review.

eboasson (approved):

PR #234, opened 2020-09-03
Don't run CI for documentation-only changes

rotu:
Make Joe happier.


joespeed:
@rotu happy Joe is a less annoying Joe🤣

PR #244, opened 2020-09-21
Update gid API return codes.

hidmic:
To be compliant with the API spec. Connected to https://github.com/ros2/rmw/pull/274.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -2033,11 +2033,14 @@ extern "C" rmw_publisher_t * rmw_create_publisher(
 
 extern "C" rmw_ret_t rmw_get_gid_for_publisher(const rmw_publisher_t * publisher, rmw_gid_t * gid)
 {
-  RET_NULL(publisher);
-  RET_WRONG_IMPLID(publisher);
-  RET_NULL(gid);
+  RMW_CHECK_ARGUMENT_FOR_NULL(publisher, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    publisher,
+    publisher->implementation_identifier,
+    eclipse_cyclonedds_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RMW_CHECK_ARGUMENT_FOR_NULL(gid, RMW_RET_INVALID_ARGUMENT);
   auto pub = static_cast<const CddsPublisher *>(publisher->data);
-  RET_NULL(pub);
   gid->implementation_identifier = eclipse_cyclonedds_identifier;
   memset(gid->data, 0, sizeof(gid->data));
   assert(sizeof(pub->pubiid) <= sizeof(gid->data));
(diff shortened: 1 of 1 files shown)


hidmic:
CI up to `test_rmw_implementation`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12383/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7352/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10095/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12296/)

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2033,11 +2033,14 @@ extern "C" rmw_publisher_t * rmw_create_publisher(
...
-  RET_NULL(gid);
+  RMW_CHECK_ARGUMENT_FOR_NULL(publisher, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    publisher,
+    publisher->implementation_identifier,
+    eclipse_cyclonedds_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RMW_CHECK_ARGUMENT_FOR_NULL(gid, RMW_RET_INVALID_ARGUMENT);
   auto pub = static_cast<const CddsPublisher *>(publisher->data);
-  RET_NULL(pub);
```
Lobotuerk, line 2040: So, you don't need to check if publisher->data was casteable?

hidmic: To `static_cast` a `void *` gives no guarantees as to whether that pointer actually refers to such object type e.g. `publisher->data` could be equal to `0xFFFFFFFF`, and it'd still succeed. 

To check for nullity is insufficient at best, harmful at worst. Insufficient because it's the canonical invalid pointer but not the only one -- and we cannot detect any other invalid pointer. Harmful because it deals with the situation as if it was an expected runtime error. It's not. If `publisher->data` is invalid but the publisher belongs to this implementation, that means that either the caller broke API preconditions, the implementation itself has logical errors, or memory has been corrupted. In any case, it's UB.

Lobotuerk: So, checking if `publisher->data` is not null makes no sense either?

hidmic: Correct.

Lobotuerk: Same as https://github.com/ros2/rmw_cyclonedds/pull/244#discussion_r492741744

Lobotuerk (approved):
LGTM

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2033,11 +2033,14 @@ extern "C" rmw_publisher_t * rmw_create_publisher(
...
-  RET_NULL(publisher);
-  RET_WRONG_IMPLID(publisher);
-  RET_NULL(gid);
+  RMW_CHECK_ARGUMENT_FOR_NULL(publisher, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    publisher,
+    publisher->implementation_identifier,
+    eclipse_cyclonedds_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RMW_CHECK_ARGUMENT_FOR_NULL(gid, RMW_RET_INVALID_ARGUMENT);
```
ivanpauno, line 2042: does the order in the docs matter?
e.g.: https://github.com/ros2/rmw/blob/a680b34da1b5be45d946327ffb3f19386e504af1/rmw/include/rmw/rmw.h#L2110-L2111

if you pass a `rmw_publisher_t` with invalid identifier and a null `gid`, which error should be returned?

I'm fine if that's undefined, I don't think it matters, but I wanted to double check.

ivanpauno (approved):

ivanpauno:
> CI up to test_rmw_implementation:

Consider running CI up to rcl

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: So far, we've left that bit undefined across the entire API. It seemed like an unnecessary constrain.

hidmic:
CI up `test_rmw_implementation` and `rcl`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12401/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7369/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10112/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12314/)

hidmic:
Alright, finally green. Going in!

PR #247, opened 2020-09-23
Update service/client construction/destruction API return codes.

hidmic:
Connected to https://github.com/ros2/rmw/pull/276. Precisely what the title says.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -3757,10 +3757,32 @@ static rmw_ret_t rmw_init_cs(
   const char * service_name, const rmw_qos_profile_t * qos_policies,
   bool is_service)
 {
-  RET_NULL(node);
-  RET_WRONG_IMPLID(node);
-  RET_NULL_OR_EMPTYSTR(service_name);
-  RET_NULL(qos_policies);
+  RMW_CHECK_ARGUMENT_FOR_NULL(node, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    node,
+    node->implementation_identifier,
+    eclipse_cyclonedds_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RMW_CHECK_ARGUMENT_FOR_NULL(type_supports, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_ARGUMENT_FOR_NULL(service_name, RMW_RET_INVALID_ARGUMENT);
+  if (0 == strlen(service_name)) {
+    RMW_SET_ERROR_MSG("service_name argument is an empty string");
+    return RMW_RET_INVALID_ARGUMENT;
+  }
+  RMW_CHECK_ARGUMENT_FOR_NULL(qos_policies, RMW_RET_INVALID_ARGUMENT);
+  if (!qos_policies->avoid_ros_namespace_conventions) {
+    int validation_result = RMW_TOPIC_VALID;
+    rmw_ret_t ret = rmw_validate_full_topic_name(service_name, &validation_result, nullptr);
+    if (RMW_RET_OK != ret) {
+      return ret;
+    }
+    if (RMW_TOPIC_VALID != validation_result) {
+      const char * reason = rmw_full_topic_name_validation_result_string(validation_result);
+      RMW_SET_ERROR_MSG_WITH_FORMAT_STRING("service_name argument is invalid: %s", reason);
+      return RMW_RET_INVALID_ARGUMENT;
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


hidmic:
CI up to `test_rmw_implementation` and `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12435/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7402/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10145/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12348/)

Lobotuerk (approved):
LGTM

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3823,10 +3845,13 @@ static rmw_ret_t rmw_init_cs(
     RMW_SET_ERROR_MSG("failed to create topic");
     goto fail_subtopic;
   }
+  // before proceeding to outright ignore given QoS policies, sanity check them
   dds_qos_t * qos;
-  if ((qos = dds_create_qos()) == nullptr) {
+  if ((qos = create_readwrite_qos(qos_policies, false)) == nullptr) {
     goto fail_qos;
   }
+  dds_reset_qos(qos);
```
ivanpauno, line 3853: can you explain this?

eboasson: I also don't understand this, but then I also don't understand what the QoS policies are doing here in the first place:
* best-effort, keep-last or a lifespan are fundamentally broken because the request or the response may then be dropped silently;
* service invocations are (by definition) aperiodic data, so deadline doesn't apply, and, it would seem, neither does liveliness;
* transient-local would be an interesting choice for dealing with the discovery latencies (and can safely be done with cyclone by leaving the "durability service history" at "keep-last 1", but the discovery latencies have been dealt with in another way already, and so doesn't make sense either.

Allowing `avoid_ros_namespace_conventions` might be useful, but I think one should really only use that for interfacing with existing non-ROS 2 systems. It seems unlikely that those systems would happen to implement the service invocation protocol.

Arguably that leaves only `ignore_local_publications`, which could be interpreted as a request to ignore services in the same node, or participant, or, indeed, process. But I'm not sure what the interpretation of it is supposed to be and it doesn't seem to be supported by all RMW implementations anyway. So I doubt it would be wise to interpret it.

And so I'd argue none of the QoS policies are meaningful, and if you want to check the validity, you should check whether they match the above, or are set to "system default" ...

ivanpauno: > I also don't understand this, but then I also don't understand what the QoS policies are doing here in the first place:

Yes, we should probably have separated service/topic qos profile definitions.
Service QoS are mostly unusable right now.

hidmic: Indeed I've never seen service QoS policies being used in any meaningful way out there.

However, I won't argue here with the decisions made by each implementation. This change is simply ensuring `rmw_cyclonedds_cpp` does not accept invalid QoS profiles, as [the API mandates](https://github.com/ros2/rmw/pull/276/files#diff-137eb23aab0499b25f47cc6e60812dbfR1762).

hidmic:
CI up to `test_rmw_implementation` and `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12448/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7415/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10158/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12363/)

rmw_cyclonedds_cpp/src/rmw_node.cpp
ivanpauno: I imagine that this `create_readwrite_qos`/`dds_reset_qos` combo is the same than calling `dds_create_qos`.
If that's the case, lgtm.

eboasson: It is the same.

hidmic: Oh, yes, it is !

hidmic:
@eboasson @ivanpauno anything else?

ivanpauno (approved):
LGTM

Checking the validity of a qos profile that is being ignored feels weird to me.
Except that, having consistent error handling across the different rmw implementations is a great improvement.

hidmic:
> Checking the validity of a qos profile that is being ignored feels weird to me.

Yeah, it is odd. I don't want to start a discussion here and now, but this is only Tier 1 implementation that chose to ignore QoS settings when the API suggests (and now explicitly says) otherwise. Perhaps we owe ourselves a discussion about this, but right now we need a (seemingly) consistent API.

hidmic:
Thanks for the review @ivanpauno !

PR #248, opened 2020-09-24
Fix array `get_function` semantics

ivanpauno:
See https://github.com/ros2/rosidl/pull/531.

### DIFF
--- a/rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
+++ b/rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
@@ -56,58 +56,6 @@ SPECIALIZE_GENERIC_C_SEQUENCE(uint32, uint32_t)
 SPECIALIZE_GENERIC_C_SEQUENCE(int64, int64_t)
 SPECIALIZE_GENERIC_C_SEQUENCE(uint64, uint64_t)
 
-typedef struct rosidl_runtime_c__void__Sequence
-{
-  void * data;
-  /// The number of valid items in data
-  size_t size;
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
```diff
@@ -56,58 +56,6 @@ SPECIALIZE_GENERIC_C_SEQUENCE(uint32, uint32_t)
...
-    sequence->data = nullptr;
-    sequence->size = 0;
-    sequence->capacity = 0;
-  } else {
-    // ensure that data, size, and capacity values are consistent
-    assert(0 == sequence->size);
-    assert(0 == sequence->capacity);
-  }
-}
-
```
dirk-thomas, line 110: This change doesn't seem to be related to the topic of this PR?

dirk-thomas (approved):
Beside the one inline comment this LGTM

rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
ivanpauno: That one is a leftover of https://github.com/ros2/rmw_cyclonedds/pull/228 and it's not the main goal of this PR.
This is also kind of a follow-up of https://github.com/ros2/rmw_cyclonedds/pull/228, but I can break the change if two PRs if needed.

dirk-thomas: I don't mind either or. Your call. Just want to make sure that is the case.

PR #267, opened 2020-11-20
Handle potential divide by 0

brawner:
Since `get_cdr_alignof_primitive` can potentially return 0 if `v.type_kind()` is not a primitive type, clang static analysis flags this `if` branch as having potentially undefined behavior.

I wasn't entirely sure what to do here. I could use an `assert(cdr_alignof != 0)` since this is situation is not supposed to occur, but returning false also seemed like a decent solution.

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/rmw_cyclonedds_cpp/src/Serialization.cpp
+++ b/rmw_cyclonedds_cpp/src/Serialization.cpp
@@ -357,9 +357,10 @@ protected:
   {
     align %= max_align;
 
-    // Value of 0 implies it is not a primitive
+    // Value of 0 implies it is not a primitive, which should not happen and is checked elsewhere
     const size_t cdr_alignof = get_cdr_alignof_primitive(v.type_kind());
-    if (cdr_alignof == 0 || align % cdr_alignof != 0) {
+    assert(0 != cdr_alignof);
+    if (align % cdr_alignof != 0) {
       return false;
     }
     return v.sizeof_type() == get_cdr_size_of_primitive(v.type_kind());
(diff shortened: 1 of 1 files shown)


clalancette (approved):
Seems reasonable to me.  I'd like to get a second opinion from @ivanpauno or @eboasson though.

brawner:
Testing `--packages-select rmw_cyclonedds_cpp`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13147/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8078/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10856/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13143/)

eboasson:
My preference would be to use an assertion, simply because returning `false` makes it seem like it is an expected condition where this is something that is not supposed to happen by design. That said, I'm not opposed to the proposed change.

brawner:
I'm happy using an assert. Updated and retesting

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13156/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8087/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10871/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13161/)

brawner:
cmake warnings on windows are a known issue and pull requests to deal with them have been opened.

eboasson (approved):
LGTM

Issue #269, opened 2020-11-25
Lots of memory leaks in minimal publisher example


sloretz:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - Source (foxy)
- Version or commit hash:
  - c8b34e3ddb65017acdcb3c1fc412aaa9d9a1dd1b
- DDS implementation:
  - CycloneDDS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

Create a package called `pub_mem_leak`.

```
pub_mem_leak/
├── CMakeLists.txt
├── package.xml
└── pub_mem_leak.cpp
```

CMakeLists.txt

```cmake
cmake_minimum_required(VERSION 3.5)
project(pub_mem_leak)

# Default to C++14
if(NOT CMAKE_CXX_STANDARD)
  set(CMAKE_CXX_STANDARD 14)
endif()

if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  add_compile_options(-Wall -Wextra -Wpedantic)
endif()

find_package(ament_cmake REQUIRED)
find_package(rclcpp REQUIRED)
find_package(std_msgs REQUIRED)

add_executable(pub_mem_leak pub_mem_leak.cpp)
target_link_libraries(pub_mem_leak
  rclcpp::rclcpp
  ${std_msgs_TARGETS}
)
target_compile_options(pub_mem_leak PUBLIC $<$<CXX_COMPILER_ID:GNU>:-fsanitize=leak>)
target_link_libraries(pub_mem_leak $<$<CXX_COMPILER_ID:GNU>:-fsanitize=leak>)

install(TARGETS
  pub_mem_leak
  DESTINATION lib/${PROJECT_NAME}
)

ament_package()
```

package.xml

```xml
<?xml version="1.0"?>
<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
<package format="2">
  <name>pub_mem_leak</name>
  <version>0.1.0</version>
  <description>See if creating/deleting publishers leaks memory</description>
  <maintainer email="sloretz@openrobotics.org">Shane Loretz</maintainer>
  <license>Apache License 2.0</license>

  <buildtool_depend>ament_cmake</buildtool_depend>

  <depend>rclcpp</depend>
  <depend>std_msgs</depend>

  <export>
[...]

PR #225, opened 2020-08-24
Fix memory leak that type support not deleted.

iuhilnehc-ynos:
Related to ros2/rcl#721

<details><summary>memory leak log</summary>
<p>

```shell
==23528== 92 (40 direct, 52 indirect) bytes in 1 blocks are definitely lost in loss record 38 of 72
==23528==    at 0x4C3017F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==23528==    by 0x5D9668B: create_message_type_support(void const*, char const*) (serdata.cpp:84)
==23528==    by 0x5D379CF: create_cdds_publisher(int, int, rosidl_message_type_support_t const*, char const*, rmw_qos_profile_t const*) (rmw_node.cpp:1839)
==23528==    by 0x5D37E25: create_publisher(int, int, rosidl_message_type_support_t const*, char const*, rmw_qos_profile_t const*, rmw_publisher_options_t const*) (rmw_node.cpp:1910)
==23528==    by 0x5D34221: rmw_context_impl_t::init(rmw_init_options_t*) (rmw_node.cpp:1028)
==23528==    by 0x5D357CC: rmw_create_node (rmw_node.cpp:1258)
==23528==    by 0x4E59265: rcl_node_init (node.c:263)
==23528==    by 0x161872: TestNodeFixture__rmw_cyclonedds_cpp_test_rcl_node_accessors_Test::TestBody() (test_node.cpp:107)
==23528==    by 0x1D242F: void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2447)
==23528==    by 0x1CC1B4: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2483)
==23528==    by 0x1AA5A1: testing::Test::Run() (gtest.cc:2522)
==23528==    by 0x1AAF36: testing::TestInfo::Run() (gtest.cc:2703)
==23528== 
==23528== 92 (40 direct, 52 indirect) bytes in 1 blocks are definitely lost in loss record 39 of 72
==23528==    at 0x4C3017F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==23528==    by 0x5D9668B: create_message_type_support(void const*, char const*) (serdata.cpp:84)
[...]

PR #270, opened 2020-11-26
Use type erased unique_ptr for type support classes to fix memory leaks

sloretz:
This reduces the number of leaks reported in #269. It uses a type-erased `unique_ptr<void,...>` instead of a `void *` for the type_support classes to make sure they get deleted. The amount of leaked data in the example in #269 reduces from 2517 bytes in 57 allocations without this PR to 900 bytes leaked in 21 allocations with this PR.

Note, I tested this on the `foxy` branch, but I've opened the PR on `master` because it cherry-picked cleanly. Assuming it's merged, it should be backported to Foxy.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -80,6 +80,7 @@
 #include "serdes.hpp"
 #include "serdata.hpp"
 #include "demangle.hpp"
+#include "unique_void.hpp"
 
 using namespace std::literals::chrono_literals;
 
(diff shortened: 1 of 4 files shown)


eboasson (approved):
LGTM!

iuhilnehc-ynos:
Do you think the following source code :confused:  should be removed after using `unique_void_ptr`?
https://github.com/ros2/rmw_cyclonedds/blob/f214d3cbf6984600a2f4072aaa7abf25773641f1/rmw_cyclonedds_cpp/src/serdata.cpp#L438-L445

eboasson:
@sloretz oh dear, I had this nagging doubt in my mind but the changes did look good to me ... @iuhilnehc-ynos, now I know why I had that doubt ...

Please consider my LGTM to be a statement about aesthetics!

jacobperron:
Bumping to Foxy patch 5

sloretz:
It looks like the leaks that would have been fixed by this PR are already fixed on master as of bb15182f from #225.

> Do you think the following source code confused should be removed after using unique_void_ptr?

I removed the delete calls in edd7698. That said, the `unique_void_ptr` doesn't seem to have a practical benefit other than moving a little bit of the code away from raw pointer `new`/`delete` to RAII. I can run CI if you'd like to see it landed, but I would recommend closing this PR instead.

iuhilnehc-ynos:
using `unique_void_ptr` seems more elegant.
IMO, we can revert https://github.com/ros2/rmw_cyclonedds/commit/bb15182fe107efdc54ccbfe49d31812543aa92dc, and then put `unique_void_ptr` into `rcpputils`, finally apply unique_void_ptr into this PR.
What do you think?

jacobperron:
@sloretz Any plans to continue with this PR or should we close it?

sloretz:
Unfortunately I don't think I'll have time to look at this again any time soon. I'll close it.

PR #271, opened 2020-12-15
Handle typesupport errors on fetch.

hidmic:
Follow-up after https://github.com/ros2/rosidl_typesupport/pull/99. Missing error handling code caused test failures in downstream packages (e.g. https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_release/1415/). This patch addresses them.

Full CI (just in case):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13273/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8205/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10993/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13304/)

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1568,13 +1568,22 @@ static const rosidl_message_type_support_t * get_typesupport(
   {
     return ts;
   } else {
+    rcutils_error_string_t prev_error_string = rcutils_get_error_string();
+    rcutils_reset_error();
     if ((ts =
       get_message_typesupport_handle(
         type_supports, rosidl_typesupport_introspection_cpp::typesupport_identifier)) != nullptr)
     {
       return ts;
     } else {
-      RMW_SET_ERROR_MSG("type support not from this implementation");
+      rcutils_error_string_t error_string = rcutils_get_error_string();
+      rcutils_reset_error();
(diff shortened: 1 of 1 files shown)


clalancette (approved):
Looks good to me.  See my [review comment](https://github.com/ros2/rmw_fastrtps/pull/495#pullrequestreview-552610624) on rmw_fastrtps for an explanation of why this makes sense.

eboasson:
It looks to me like this fix was insufficient to resolve all the test failures that check the error string because of the work done in `make_message_value_type` and `make_request_response_value_types`. See, e.g., https://ci.ros2.org/job/ci_linux/13296/#showFailuresLink

I really think the underlying issue is that `get_typesupport_handle_function` in `rosidl_typesupport_cpp` sets an error message even though the function is being used to poll for the presence of some type support, and so a lookup failure should simply not be treated as an error at that point.

One way to fix this would be to not set a global error string in `get_typesupport_handle_function` and instead let the callers do it once they know what the error means. That would also eliminate the messing about with error strings in handling typesupports in the RMW implementations, which is an improvement in itself.

I think the nicest option would be to enhance the rosidl_typesupport interface so that one can specify the allowed "suffixes", then dispatch in the RMW implementation based on the one that was found. That sounds like it is more work in the typesupport code, but it would remove the need to fix all the RMW layers.

@hidmic @clalancette what do you think?

PR #272, opened 2020-12-16
Compatibility with Cyclone DDS type discovery

eboasson:
Cyclone DDS is about to get type discovery in eclipse-cyclonedds/cyclonedds#559 and the evolving interface for custom message representation/serialisation will change a little bit as a consequence. That PR does provide ABI compatibility, so an old `rmw_cyclonedds_cpp` binary will still work fine (if slightly slower), eliminating one risk in upgrading binary packages. (Source compatibility could have been provided but I decided against that because the benefit is just too marginal, and it isn't an official stable API yet anyway.)

The first commit in this PR updates the interface and uses the C preprocessor to also support building against a version of Cyclone DDS that hasn't got the type discovery merged in yet. It is fairly big because it consistently updates all places where it really should be using "type" rather than the "topic" that was a quirk of the old interface we really wanted to eliminate.

The second commit in this PR drops some conditionally compiled parts for supporting Cyclone versions of (about) a year ago. I don't think it is realistic that anyone would want to take the RMW layer and build it against a version of Cyclone from January 2020.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -2724,8 +2724,8 @@ static rmw_ret_t rmw_take_ser_int(
   CddsSubscription * sub = static_cast<CddsSubscription *>(subscription->data);
   RET_NULL(sub);
   dds_sample_info_t info;
-  struct ddsi_serdata * dcmn;
-  while (dds_takecdr(sub->enth, &dcmn, 1, &info, DDS_ANY_STATE) == 1) {
+  struct ddsi_serdata * d;
+  while (dds_takecdr(sub->enth, &d, 1, &info, DDS_ANY_STATE) == 1) {
     if (info.valid_data) {
       if (message_info) {
         message_info->publisher_gid.implementation_identifier = eclipse_cyclonedds_identifier;
@@ -2735,20 +2735,19 @@ static rmw_ret_t rmw_take_ser_int(
           message_info->publisher_gid.data, &info.publication_handle,
           sizeof(info.publication_handle));
       }
-      auto d = static_cast<serdata_rmw *>(dcmn);
-      /* FIXME: what about the header - should be included or not? */
-      if (rmw_serialized_message_resize(serialized_message, d->size()) != RMW_RET_OK) {
-        ddsi_serdata_unref(dcmn);
+      size_t size = ddsi_serdata_size(d);
+      if (rmw_serialized_message_resize(serialized_message, size) != RMW_RET_OK) {
+        ddsi_serdata_unref(d);
         *taken = false;
         return RMW_RET_ERROR;
       }
-      memcpy(serialized_message->buffer, d->data(), d->size());
-      serialized_message->buffer_length = d->size();
-      ddsi_serdata_unref(dcmn);
+      ddsi_serdata_to_ser(d, 0, size, serialized_message->buffer);
+      serialized_message->buffer_length = size;
+      ddsi_serdata_unref(d);
       *taken = true;
       return RMW_RET_OK;
     }
-    ddsi_serdata_unref(dcmn);
+    ddsi_serdata_unref(d);
   }
   *taken = false;
   return RMW_RET_OK;
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -76,7 +76,6 @@
 #include "namespace_prefix.hpp"
 
 #include "dds/dds.h"
-#include "dds/ddsi/ddsi_sertopic.h"
 #include "serdes.hpp"
 #include "serdata.hpp"
 #include "demangle.hpp"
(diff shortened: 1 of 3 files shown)


eboasson:
@ivanpauno would you be willing to review this PR? It is really a triviality, but it relates to https://github.com/eclipse-cyclonedds/cyclonedds/pull/559 which is not quite so trivial. We're now almost at the point that I feel significant changes can be introduced without risking complications in changing the default RMW, so if this PR is ok then I will start with carefully merging things one step at a time.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1867,12 +1875,13 @@ static CddsPublisher * create_cdds_publisher(
...
   std::string fqtopic_name = make_fqtopic(ROS_TOPIC_PREFIX, topic_name, "", qos_policies);
 
-  auto sertopic = create_sertopic(
+  auto sertype = create_sertype(
     fqtopic_name.c_str(), type_support->typesupport_identifier,
     create_message_type_support(type_support->data, type_support->typesupport_identifier), false,
     rmw_cyclonedds_cpp::make_message_value_type(type_supports));
-  struct ddsi_sertopic * stact;
-  topic = create_topic(dds_ppant, sertopic, &stact);
+  rcutils_reset_error();
```
ivanpauno, line 1882: why `rcutils_reset_error()` here?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2243,11 +2252,12 @@ static CddsSubscription * create_cdds_subscription(
 
   std::string fqtopic_name = make_fqtopic(ROS_TOPIC_PREFIX, topic_name, "", qos_policies);
 
-  auto sertopic = create_sertopic(
+  auto sertype = create_sertype(
     fqtopic_name.c_str(), type_support->typesupport_identifier,
     create_message_type_support(type_support->data, type_support->typesupport_identifier), false,
     rmw_cyclonedds_cpp::make_message_value_type(type_supports));
-  topic = create_topic(dds_ppant, sertopic);
+  rcutils_reset_error();
```
ivanpauno, line 2259: same question here

ivanpauno (approved):
The changes look good to me!

I'm not sure about some of the added `rcutils_reset_error()`, I would expect that function to be called immediately after an error occurred and not in other places.

eboasson:
> The changes look good to me!

Thanks for reviewing and catching:

> I'm not sure about some of the added `rcutils_reset_error()`, I would expect that function to be called immediately after an error occurred and not in other places.

which a simple workaround for the large number of tests failing on unexpected "typesupport is not from this implementation" errors in checking error/exception messages with 8bce86b2770e6ea4d4713e70ea48f13070472da7 not yet in place. They were accidentally included in the commit ...

eboasson:
CI rclcpp & test_communication of this PR rebased on current master:

Linux [Build Status](https://ci.ros2.org/job/ci_linux/)
Linux aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/)
OS X [Build Status](https://ci.ros2.org/job/ci_osx/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/)

(I hope I got the badges right!)

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3828,6 +3835,7 @@ static rmw_ret_t rmw_init_cs(
   if (is_service) {
     std::tie(sub_msg_ts, pub_msg_ts) =
       rmw_cyclonedds_cpp::make_request_response_value_types(type_supports);
+    rcutils_reset_error();
```
ivanpauno, line 3838:
```suggestion
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3839,6 +3847,7 @@ static rmw_ret_t rmw_init_cs(
   } else {
     std::tie(pub_msg_ts, sub_msg_ts) =
       rmw_cyclonedds_cpp::make_request_response_value_types(type_supports);
+    rcutils_reset_error();
```
ivanpauno, line 3850:
```suggestion
```

ivanpauno (approved):

eboasson:
Re-run of the same but with a fix in for the compiler warning on Windows

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/13429/)
* Linux aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/8344/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/11151/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/13486)

I think it is fair to say that the macOS test failure is unrelated to the change (from the test output, it looks like the kind of test that could be flaky to begin with). @ivanpauno do you agree?

eboasson:
Re-run of the same but with a fix in for the compiler warning on Windows

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/13429/)
* Linux aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/8344/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/11151/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/13486)

I think the macOS test failure is unrelated. @ivanpauno do you agree?

ivanpauno:
> I think the macOS test failure is unrelated. @ivanpauno do you agree?

Yes, that also happens on nightlies sometimes https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/2219/#showFailuresLink

PR #274, opened 2021-01-04
Reset the error on failing to find the C typesupport

clalancette:
The individual commits have more information, but essentially make sure to reset the error in between calls to probing for typesupport.  This should fix the bulk of the test failures we are seeing in https://github.com/ros2/rmw/pull/293 .

@eboasson Note that this is different than what we had discussed earlier.  After talking to some more people, changing the contract of [get_typesupport_handle_function](https://github.com/ros2/rosidl_typesupport/blob/64dd1be918b759f0a2d0c731ebe38ccbbe4601b4/rosidl_typesupport_cpp/src/type_support_dispatch.hpp#L37) is a little more dangerous than we want to attempt right now .  So this is the more targeted, if less ideal, solution.  I'll likely still open an issue to follow-up and change that contract, since the current contract seems a bit ugly.

CI for this change is in https://github.com/ros2/rmw/pull/293 .

### DIFF
--- a/rmw_cyclonedds_cpp/src/TypeSupport2.cpp
+++ b/rmw_cyclonedds_cpp/src/TypeSupport2.cpp
@@ -18,6 +18,9 @@
 #include <utility>
 #include <vector>
 
+#include "rosidl_runtime_c/message_type_support_struct.h"
+#include "rosidl_runtime_c/service_type_support_struct.h"
+
 namespace rmw_cyclonedds_cpp
 {
 class ROSIDLC_StructValueType : public StructValueType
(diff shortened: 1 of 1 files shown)


eboasson (approved):
Thanks @clalancette! It looks good to me and it definitely does the trick (https://github.com/ros2/rmw/pull/293 confirmed it).

clalancette:
> Thanks @clalancette! It looks good to me and it definitely does the trick ([ros2/rmw#293](https://github.com/ros2/rmw/pull/293) confirmed it).

Thanks for the review.  I'm going to go ahead and make this a real PR and merge it.  We can work on the rest of the failing tests over in ros2/rmw#293 .

PR #275, opened 2021-01-06
Use init-on-first-use for global state

eboasson:
Otherwise global destructor ordering can result in the global state being
destroyed before the last use.  Not cleaning up does cause a small memory leak
at shutdown, but it is limited to an empty map of domains and an empty
unordered set of waitsets.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -186,7 +186,11 @@ struct Cdds
   {}
 };
 
-static Cdds gcdds;
+static Cdds& gcdds()
+{
+  static Cdds * x = new Cdds();
+  return *x;
+}
 
 struct CddsEntity
 {
@@ -735,10 +739,10 @@ static rmw_ret_t discovery_thread_stop(rmw_dds_common::Context & common_context)
 static bool check_create_domain(dds_domainid_t did, rmw_localhost_only_t localhost_only_option)
 {
   const bool localhost_only = (localhost_only_option == RMW_LOCALHOST_ONLY_ENABLED);
-  std::lock_guard<std::mutex> lock(gcdds.domains_lock);
+  std::lock_guard<std::mutex> lock(gcdds().domains_lock);
   /* return true: n_nodes incremented, localhost_only set correctly, domain exists
      "      false: n_nodes unchanged, domain left intact if it already existed */
-  CddsDomain & dom = gcdds.domains[did];
+  CddsDomain & dom = gcdds().domains[did];
   if (dom.refcount != 0) {
     /* Localhost setting must match */
     if (localhost_only == dom.localhost_only) {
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -186,7 +186,11 @@ struct Cdds
   {}
 };
 
-static Cdds gcdds;
+static Cdds& gcdds()
```
clalancette, line 189: Uncrustify is cranky about this.

```suggestion
static Cdds & gcdds()
```

clalancette (commented):
The code looks OK to me.  There's one style error inline I've pointed out.

I also think we should add a comment above the `gcdds()` function explaining why we are doing this and the tradeoff of never calling the destructor.  That will help future readers understand.

Finally, a CI run with this in place, both with and without https://github.com/ros2/rmw/pull/293, would be good to ensure that the problem there is solved.

clalancette (approved):
Looks good to me.  I'm going to go ahead and run full CI on this both for the current state of the world, and in combination with https://github.com/ros2/rmw/pull/293 to make sure things are happy.

clalancette:
CI with just this PR in place:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13365/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8292/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11094/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13416/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/13417/)

clalancette:
CI with this PR and https://github.com/ros2/rmw/pull/293:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13366/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8293/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11095/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13418/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13419/)

clalancette:
All right, looking at the CI results with just this PR in place:

* Linux - 2 failing tests, but that should now be fixed by https://github.com/ros2/rcl/pull/883
* Linux-aarch64 - Green
* macOS - 4 failing tests.  2 of them should now be fixed by https://github.com/ros2/rcl/pull/883 .  Two of them are flaky tests; they can be seen failing in https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/2211/#showFailuresLink, for instance.
* Windows - 13 failing tests.  2 of them should now be fixed by https://github.com/ros2/rcl/pull/883 .  The remaining 11 can be seen in the nightlies as well, i.e. https://ci.ros2.org/view/nightly/job/nightly_win_rel/1808/#showFailuresLink
* Windows Debug - 4 failing tests.  All 4 are in the nightly: https://ci.ros2.org/view/nightly/job/nightly_win_deb/1860/#showFailuresLink

So as far as I can tell this PR is ready to go.  I'm going to go ahead and merge, thanks @eboasson !

PR #278, opened 2021-01-21
Fix use-after-free in error handling bug

### LINKED ISSUE #277: 👩‍🌾 Regression on test_publisher__rmw_cyclonedds_cpp
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Focal
- Installation type:
  - source
- Version or commit hash:
  - `master/ros2.repos`
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue

The [test_publisher](https://github.com/ros2/rmw_implementation/blob/5a9424fdfd93ac2c3260bfa3e857108e1ce12077/test_rmw_implementation/test/test_publisher.cpp) test from `test_rmw_implementation` has regressed for CycloneDDS on nightly CI.

According to the test history for different platforms:

* [linux_debug](https://ci.ros2.org/job/nightly_linux_debug/1814/testReport/(root)/projectroot/test_publisher__rmw_cyclonedds_cpp/history/)
*  [linux_aarch64_debug](https://ci.ros2.org/job/nightly_linux-aarch64_debug/1485/testReport/(root)/projectroot/test_publisher__rmw_cyclonedds_cpp/history/)
* [osx-debug](https://ci.ros2.org/job/nightly_osx_debug/1897/testReport/(root)/projectroot/test_publisher__rmw_cyclonedds_cpp/history/)

The test started failing on January 19th, and was passing consistently before then.

#### Expected behavior

Test passes 100% of times.

#### Actual behavior

Test has been failing 100% of times since January 19th.

#### Additional information

Failure output:

```
[----------] 7 tests from TestPublisher__rmw_cyclonedds_cpp
[ RUN      ] TestPublisher__rmw_cyclonedds_cpp.create_and_destroy
[       OK ] TestPublisher__rmw_cyclonedds_cpp.create_and_destroy (8 ms)
[ RUN      ] TestPublisher__rmw_cyclonedds_cpp.create_with_internal_errors
/home/jenkins-agent/workspace/nightly_linux_debug/ws/src/ros2/rcutils/src/allocator.c:34 Injecting fault and returning ((void *)0)
/home/jenkins-agent/workspace/nightly_linux_debug/ws/src/ros2/rcutils/src/allocator.c:34 Injecting fault and returning ((void *)0)
[       OK ] TestPublisher__rmw_cyclonedds_cpp.create_with_internal_errors (7 ms)
[ RUN      ] TestPublisher__rmw_cyclonedds_cpp.create_and_destroy_native
[       OK ] TestPublisher__rmw_cyclonedds_cpp.create_and_destroy_native (4 ms)
[ RUN      ] TestPublisher__rmw_cyclonedds_cpp.create_with_bad_arguments
test_publisher__rmw_cyclonedds_cpp: /home/jenkins-agent/workspace/nightly_linux_debug/ws/src/eclipse-cyclonedds/cyclonedds/src/core/ddsi/src/ddsi_sertype.c:127: ddsi_sertype_unref: Assertion `(flags_refc & DDSI_SERTYPE_REFC_MASK) >= 1' failed.
-- run_test.py: return code -6
-- run_test.py: generate result file '/home/jenkins-agent/workspace/nightly_linux_debug/ws/build/test_rmw_implementation/test_results/test_rmw_implementation/test_publisher__rmw_cyclonedds_cpp.gtest.xml' with failed test
-- run_test.py: verify result file '/home/jenkins-agent/workspace/nightly_linux_debug/ws/build/test_rmw_implementation/test_results/test_rmw_implementation/test_publisher__rmw_cyclonedds_cpp.gtest.xml'

```

Here's the [assert line](https://github.com/eclipse-cyclonedds/cyclonedds/blob/b84c035ce0292cd76358475f069da4aec1540bb2/src/core/ddsi/src/ddsi_sertype.c#L127) in the source code.

---

https://github.com/osrf/buildfarmer/issues/145

eboasson:
The reference to the sertype is not counted and shouldn't be released.
At this point, and in this test anyway, the memory has been freed just
before (by `dds_delete(topic)`) and this probably explains why it hasn't
caused any breakage before.  The assert that failed now is new in
Cyclone.

There has been an ancient interface for creating custom types that did
require it and that was used by `rmw_cyclonedds_cpp` in the past.  The
rewrite of the RMW layer to use one DDS participant per context
(commit 8a1d3fc092d1d4841c9f5c39e7da0e3b26886add) also dropped support
for that ancient interface.  Instead of removing the #if/unref/#endif,
that commit removed the #if/#endif but left the unref call in place.

This fixes #277

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1929,7 +1929,6 @@ fail_writer:
   dds_delete_qos(qos);
 fail_qos:
   dds_delete(topic);
-  ddsi_sertype_unref(stact);
 fail_topic:
   delete pub;
   return nullptr;


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1929,7 +1929,6 @@ static CddsPublisher * create_cdds_publisher(
   dds_delete_qos(qos);
 fail_qos:
   dds_delete(topic);
-  ddsi_sertype_unref(stact);
```
clalancette, line 1932: There's a few things in this function that are unclear to me:

1.  Why do we need the "stact" variable at all?  As far as I can tell, if it is provided to `create_topic`, it just saves a copy of the `sertype` pointer.  But we already have that pointer (we are passing it into `create_topic` to begin with).  So why do we need it?

2.  If `create_topic` fails for some reason, I feel like we are leaking the `sertype` variable; `create_sertype` allocates heap space for it, but nothing frees it in the error case.

3.  Is `dds_delete` guaranteed to free `sertype` in the case where `create_topic` succeeds, but `create_readwrite_qos` fails?  (I'm guess this is the case, given the nature of this bug, but I'm just checking).

eboasson: > So why do we need it?

There is a bit of trickery going on: one constructs a `sertype` and passes it to `create_topic` but it isn't a given that this is the one that will actually be used. Instead, `dds_create_topic_sertype` checks whether it has an equivalent one already, and if it has one, uses that instead. It always consumes the reference, which means that in the latter case it drops the `sertype` (because it was just created and there is no other reference).

The pointer it returns is a "raw" pointer, so there is no unref'ing of that (unless you increment the refcount yourself). Its lifetime is *at least* that of the topic object and of any readers/writers that you created with that topic object. The latter bit is why calling `dds_delete(topic)` in the successful path of creating a publisher is fine: it all remains valid because the writer still exists.

(You could increment the refcount prior to calling `dds_create_topic_sertype`, that would keep it alive, but it would mean you might be using a different `sertype` than the DDS writer is using. That in turn means some relatively expensive conversions are required when you try to publish a serialised message. So it makes more sense to do this.)

> If create_topic fails for some reason [...]

That case is handled inside `create_topic` in `rmw_node.cpp`: if `dds_create_topic_sertype` fails, `create_topic` immediately drops the `sertype` it just created. If it succeeds, it either retains the `sertype` or drops it to use an already existing one instead. In the former case, it is freed immediately (because there aren't any other references around); in the latter, it will be freed whenever the last reference disappears.

> Is dds_delete guaranteed to free sertype in the case where create_topic succeeds, but create_readwrite_qos fails?

Yes, at that point `topic` owns the reference. (Well, in the simplified version in which there are no other threads creating topics using an equivalent `sertype` in parallel anyway ...)

ivanpauno (approved):
LGTM!

If I understand correctly `dds_create_topic_sertype` takes ownership of the passed `sertype` if and only if the function doesn't fail.
`create_topic` [handles the error case](https://github.com/ros2/rmw_cyclonedds/blob/45179db533fc01e47c9904ca8aa42f60766806fc/rmw_cyclonedds_cpp/src/rmw_node.cpp#L1508-L1510).
If a function after `create_topic` fails, there's no need to free `stact` because that's a borrow of the internal `sertype` owned by `cyclonedds` library "global state".
After all topics that share the same `sertype` are deleted (the `dds_delete(topic);` line), the `sertype` will actually be freed too.

If my rephrasing still makes sense, I think this is fine.

eboasson:
> If my rephrasing still makes sense, I think this is fine.

It does! ☺️

clalancette:
> Before merging it: @clalancette have your concerns been allayed?

Yes.  Thanks for the explanations from both of you.

Issue #279, opened 2021-02-01
abort() in test_subscription__rmw_cyclonedds_cpp


clalancette:
We're seeing occasional failures in [test_subscription__rmw_cyclonedds_cpp::take_sequence](https://github.com/ros2/rmw_implementation/blob/master/test_rmw_implementation/test/test_subscription.cpp#L526): https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_repeated/1491/testReport/junit/(root)/projectroot/test_subscription__rmw_cyclonedds_cpp/

I finally managed to capture a stack trace, and it looks like this:

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x0000ffff81ca9d68 in __GI_abort () at abort.c:79
#2  0x0000ffff81af782c in sertype_rmw_realloc_samples (ptrs=0xaaaaca346c10, d=0xaaaaca027250, old=0x0, oldcount=0, count=1)
    at /home/ubuntu/ros2_ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/serdata.cpp:458
#3  0x0000ffff818c7290 in ddsi_sertype_realloc_samples (ptrs=0xaaaaca346c10, tp=0xaaaaca027250, old=0x0, oldcount=0, count=1)
    at /home/ubuntu/ros2_ws/src/eclipse-cyclonedds/cyclonedds/src/security/api/../../core/ddsi/include/dds/ddsi/ddsi_sertype.h:160
#4  0x0000ffff8195959c in dds_read_impl (take=true, reader_or_condition=682686299, buf=0xaaaaca346c10, bufsz=1, maxs=1, si=0xaaaaca3446b0, mask=128, hand=0, 
    lock=true, only_reader=false) at /home/ubuntu/ros2_ws/src/eclipse-cyclonedds/cyclonedds/src/core/ddsc/src/dds_read.c:93
#5  0x0000ffff8195a0d8 in dds_take (rd_or_cnd=682686299, buf=0xaaaaca346c10, si=0xaaaaca3446b0, bufsz=1, maxs=1)
    at /home/ubuntu/ros2_ws/src/eclipse-cyclonedds/cyclonedds/src/core/ddsc/src/dds_read.c:354
[...]

PR #280, opened 2021-02-03
Fix dds_take buffer argument in rmw_take_sequence

eboasson:
This fixes the crash in #279 

It doesn't address the other observations made there, such as the need for improving Cyclone DDS' read/take interface.

I also note that there are *no tests* in ROS 2 for the "take sequence" operation where it actually returns some data. I've hacked it into the rmw implementation subscription tests to check my hypothesis on the crash and the behaviour of my fix, but that hack isn't suitable for public viewing. Perhaps it would make sense to consider adding proper tests in a proper location.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -2656,8 +2656,12 @@ static rmw_ret_t rmw_take_seq(
   RET_NULL(sub);
 
   std::vector<dds_sample_info_t> infos(count);
+  std::vector<void *> ptrs(count);
   auto maxsamples = static_cast<uint32_t>(count);
-  auto ret = dds_take(sub->enth, message_sequence->data, infos.data(), count, maxsamples);
+  for (size_t i = 0; i < count; i++) {
+    ptrs[i] = &message_sequence->data[i];
+  }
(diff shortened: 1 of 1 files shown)


clalancette (approved):
This seems reasonable to me and should fix the crash.  I'll open an issue over at https://github.com/ros2/rmw_implementation to implement a proper test for this case.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13597/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8480/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11311/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13654/)

clalancette:
@eboasson Where are we with this PR?  Do we still need it?

eboasson:
@clalancette I think not: from the changes in this PR I suspect I misread the `rmw_take_sequence` interface and the sequence of messages passed in actually conforms to what Cyclone DDS expects (an initialised array of pointers to messages), and https://github.com/ros2/rmw_implementation/pull/175 seems to clarify that.

That makes this one incorrect and the problem it tried to address a non-existent problem.

Thanks for drawing attention to it!

PR #288, opened 2021-02-19
Take and return new RMW_DURATION_INFINITE correctly

emersonknapp:
Depends on https://github.com/ros2/rmw/pull/301
Alternative to https://github.com/ros2/rmw_cyclonedds/pull/287

Meets the new API expectations introduced in https://github.com/ros2/rmw/pull/301 - to accept and return `RMW_DURATION_INFINITE` for cross-implementation consistency.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1635,6 +1635,12 @@ static std::string make_fqtopic(
   return make_fqtopic(prefix, topic_name, suffix, qos_policies->avoid_ros_namespace_conventions);
 }
 
+static bool is_rmw_duration_default(rmw_time_t duration)
+{
+  return rmw_time_equal(duration, RMW_QOS_DURATION_UNSPECIFIED) ||
+         rmw_time_equal(duration, RMW_DURATION_INFINITE);
+}
+
 static dds_qos_t * create_readwrite_qos(
   const rmw_qos_profile_t * qos_policies,
   bool ignore_local_publications)
@@ -1700,20 +1706,17 @@ static dds_qos_t * create_readwrite_qos(
     default:
       rmw_cyclonedds_cpp::unreachable();
   }
-  if (qos_policies->lifespan.sec > 0 || qos_policies->lifespan.nsec > 0) {
-    dds_qset_lifespan(qos, DDS_SECS(qos_policies->lifespan.sec) + qos_policies->lifespan.nsec);
+  if (!is_rmw_duration_default(qos_policies->lifespan)) {
+    dds_qset_lifespan(qos, rmw_time_total_nsec(qos_policies->lifespan));
   }
-  if (qos_policies->deadline.sec > 0 || qos_policies->deadline.nsec > 0) {
-    dds_qset_deadline(qos, DDS_SECS(qos_policies->deadline.sec) + qos_policies->deadline.nsec);
+  if (!is_rmw_duration_default(qos_policies->deadline)) {
+    dds_qset_deadline(qos, rmw_time_total_nsec(qos_policies->deadline));
   }
 
-  if (qos_policies->liveliness_lease_duration.sec == 0 &&
-    qos_policies->liveliness_lease_duration.nsec == 0)
-  {
+  if (is_rmw_duration_default(qos_policies->liveliness_lease_duration)) {
     ldur = DDS_INFINITY;
   } else {
-    ldur = DDS_SECS(qos_policies->liveliness_lease_duration.sec) +
-      qos_policies->liveliness_lease_duration.nsec;
+    ldur = rmw_time_total_nsec(qos_policies->liveliness_lease_duration);
   }
   switch (qos_policies->liveliness) {
     case RMW_QOS_POLICY_LIVELINESS_SYSTEM_DEFAULT:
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1634,6 +1634,12 @@ static std::string make_fqtopic(
   return make_fqtopic(prefix, topic_name, suffix, qos_policies->avoid_ros_namespace_conventions);
 }
 
+static bool is_rmw_duration_default(rmw_time_t duration)
+{
+  return rmw_time_equal(duration, RMW_QOS_DURATION_UNSPECIFIED) ||
+         rmw_time_equal(duration, RMW_DURATION_INFINITE);
```
eboasson, line 1640: I find it surprising that ∞ is treated as default as well. It does mean that `rmw_time_total_nsec` can be used instead of a Cyclone-specific conversion function mapping `RMW_DURATION_INFINITE` to `DDS_INFINITY`, but to me it is a bit of a stain on an otherwise clean proposal 🙂

ivanpauno: I agree with this comment.
Isn't possible to have another function to compare the input value to "RMW_DURATION_INFINITE" and in that case set `DDS_INFINITY`?

emersonknapp: Sure, that seems fine to me. I wasn't sure if setting e.g. Deadline to DDS_INFINITY was the same thing as not setting it at all. 

I noticed that in the current implementation, an input of 0 for `liveliness.lease_duration` is explicitly translated to `DDS_INFINITY` - but for Deadline and Lifespan, this same input is ignored, so the value is not set at all.

I would like for users providing `RMW_DURATION_INFINITE` to Deadline+Lifespan to get the same behavior as they do when passing 0 now. Does setting Deadline/Lifespan to `DDS_INFINITY` have the same behavior as not setting it?

eboasson: There is a small difference between setting it to ∞ and leaving it unset in where the value comes from: in Cyclone's C API, any QoS not set explicitly is inherited from the topic for those settings that also apply to the topic. The topics here do not override any defaults, so the default value of ∞ applies either way.

Ignoring a lease duration of 0 would work, too, for the same reasons. I suppose the difference is that deadline and lifespan can be set separately, but liveliness kind and lease duration can only be set together. So while one could ignore (automatic, 0), but any other liveliness setting requires setting something. And so one might as well do it always.

I can imagine that setting these defaults in all cases would probably avoid some head scratching in the future.

ivanpauno: In other rmw implementations (I'm not sure if here), the "default" can be modified externally in a QoS profile file, so it might be different to INFINITY.

> I can imagine that setting these defaults in all cases would probably avoid some head scratching in the future.

:+1:

emersonknapp: This makes sense, thanks! I will update the RMW API proposal to `0 == unspecified`, and `RMW_DURATION_INFINITE == DDS_INFINITY` - then update this PR to match.

emersonknapp: Ok - updated to treat the cases differently.

eboasson (approved):
LGTM @emersonknapp!

ivanpauno (approved):

emersonknapp:
@eboasson @ivanpauno friendly ping - the dependent RMW PR has been merged, so this is ready to go. I notice that the actions here are not working, and I'm not sure I want to mess around with your specific CI setup - so what is the next step?

ivanpauno:
@emersonknapp can you run CI for this?

for the Rpr checker to pass you would need someone releasing rmw to rolling.
The other jobs are failing because of unrelated reasons (this [line](https://github.com/ros2/rmw_cyclonedds/blob/91aed377a0a2e39664d0eae401ee49e9c419540a/.github/resources/local.repos#L16), I think master can be used now).

clalancette:
I think we should aim for at least the Rpr job to go green (along with the CI job).  I can do an rmw release a little later.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/6c22d23c3fac5b8ced3605986aa33a46/raw/fb33f2320709a25cfb79267321a16e57dc0f4eb9/ros2.repos
BUILD args: --packages-up-to rmw_cyclonedds_cpp rcl
TEST args: --packages-select rmw_cyclonedds_cpp rcl
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7855
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13915/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8763/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11621/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14003/)

clalancette:
FYI, I did do an `rmw` release.  Given how low-level it is, it will take a few hours to finish building "everything", but once that is done the PR job should be able to compile.

ivanpauno:
@ros-pull-request-builder retest this please

PR #290, opened 2021-02-26
Add install instruction for Foxy

JEnoch:


README.md
```diff
@@ -6,7 +6,10 @@ This package lets [*ROS 2*](https://index.ros.org/doc/ros2) use [*Eclipse Cyclon
 Cyclone DDS is ready to use. It seeks to give the fastest, easiest, and most robust ROS 2 experience. Let the Cyclone blow you away!
 
 1. Install:
-
+   ```
+   apt install ros-foxy-rmw-cyclonedds-cpp
```
jacobperron, line 10: Rather than adding to this list over time, I think we should either instruct the user to use the codename for the ROS distro they are using or update this README to reflect the branch (e.g. the `master` branch is for `rolling`).

For the first option, this could look like:

```
# Set the codename for the ROS distribution you are using (e.g. galactic)
export ROS_DISTRO=galactic

apt install ros-${ROS_DISTRO}-rmw-cyclonedds-cpp
```

jacobperron (commented):
Eloquent and Dashing are both EOL, so we should probably remove those.

README.md
clalancette: > # Set the codename for the ROS distribution you are using (e.g. galactic)
> export ROS_DISTRO=galactic

I don't think we should instruct users to explicitly set `ROS_DISTRO` in their environment, as that potentially clashes with what `source /opt/ros/<distro>/setup.bash` does.

clalancette:
No response from the original contributor in almost a year, so I'm going to close this out.  Feel free to reopen and respond to the comments from the review.

PR #291, opened 2021-03-11
Fix github workflows

ivanpauno:
Workflows started to fail after https://github.com/ament/ament_lint/pull/268 was merged, e.g.: https://github.com/ros2/rmw_cyclonedds/pull/288.

This should fix it :crossed_fingers:

### DIFF
--- a/.github/resources/local.repos
+++ b/.github/resources/local.repos
@@ -9,8 +9,3 @@ repositories:
 
   ros2/rosidl_typesupport_fastrtps/COLCON_IGNORE: *empty_repo
   ros2/rmw_fastrtps/COLCON_IGNORE: *empty_repo
-
-  ament/ament_lint:
-    type: git
-    url: https://github.com/ament/ament_lint.git
-    version: pull/268/merge
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,7 +31,7 @@ jobs:
       run: git config --global --add remote.origin.fetch '+refs/pull/*:refs/remotes/origin/pull/*'
     - name: Build and test ROS
       id: ros_ci
-      uses: ros-tooling/action-ros-ci@0.0.19
+      uses: ros-tooling/action-ros-ci@v0.1
       with:
         package-name: >
           rmw_cyclonedds_cpp


emersonknapp (approved):

emersonknapp:
I might recommend that you upgrade to `action-ros-ci@v0.1` instead of `0.0.19` - which has some important fixes

ivanpauno:
@emersonknapp do you know what could be ground on the Windows jobs?

This is already greener than before, so I think that the windows issue can be addressed afterwards.

emersonknapp:
No, I'm not sure exactly what's wrong. I have not been running windows jobs with the actions recently. I agree that it is a strict improvement and should be merged.

christophebedard:
It seems like this step (removing the duplicate package if it's part of the `ros2.repos` file) doesn't work on Windows: https://github.com/ros-tooling/action-ros-ci/blob/64858bf3c148043e873c2985d7c9fed60914061c/src/action-ros-ci.ts#L338-L347

clalancette:
Also, I canceled the two builds on ubuntu-20.04.  They were both blocked waiting for someone to agree to the RTI license.  Somehow the ubuntu-18.04 builder gets around this (either by not installing Connext, or setting an environment variable to skip the prompt), but the 20.04 don't seem to be setup the same way.

emersonknapp:
> Also, I canceled the two builds on ubuntu-20.04. They were both blocked waiting for someone to agree to the RTI license. Somehow the ubuntu-18.04 builder gets around this (either by not installing Connext, or setting an environment variable to skip the prompt), but the 20.04 don't seem to be setup the same way.

That's weird, this should have been fixed in https://github.com/ros-tooling/action-ros-ci/pull/523, which is released to the v0.1 tag

clalancette:
> > 
> 
> That's weird, this should have been fixed in [ros-tooling/action-ros-ci#523](https://github.com/ros-tooling/action-ros-ci/pull/523), which is released to the v0.1 tag

I see what you mean in terms of the source repository for action-ros-ci.  However, https://github.com/ros-tooling/action-ros-ci/pull/523 doesn't seem to be achieving the desired effect.  It looks like `rti-connext-5.3.1` is being dragged in as a recursive dependency of something else.  I haven't looked closely, but maybe `rosdep --skip-keys` only looks at immediate key dependencies, and not recursive ones?  I'm not sure.

eboasson (approved):
I always consider this as a bit of black magic that miraculously landed in the repo thanks to @rotu and had no idea that resuscitating it would be so straightforward. Thanks @ivanpauno !

It seems to me that hanging tests could be bit of an issue (I assume a hanging run on one PR might cause it not to run for another PR) but otherwise I all in favour of the changes. It going green for Linux is a major improvement over everything all red, all the time.

PR #305, opened 2021-04-20
Fix the history depth for KEEP_ALL.

clalancette:
The comment in the code explains it in a lot more detail, but
essentially this avoids a signed/unsigned mismatch in higher
level code.  Since the depth is meaningless in KEEP_ALL anyway,
this shouldn't have any deleterious effects elsewhere.

Originally from Erik Boasson (@eboasson).

This should fix https://github.com/ros2/ros2cli/issues/603

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1849,7 +1849,13 @@ static bool dds_qos_to_rmw_qos(const dds_qos_t * dds_qos, rmw_qos_profile_t * qo
         break;
       case DDS_HISTORY_KEEP_ALL:
         qos_policies->history = RMW_QOS_POLICY_HISTORY_KEEP_ALL;
-        qos_policies->depth = (uint32_t) depth;
+        // When using a policy of KEEP_ALL, the depth is meaningless.
+        // CycloneDDS reports this as -1, but the rmw_qos_profile_t structure
+        // expects an unsigned number.  Casting -1 to unsigned would yield
+        // a value of 2^32 - 1, but unfortunately our XML-RPC connection
+        // (used for the command-line tools) doesn't understand anything
+        // larger than 2^31 - 1.  Just set the depth to 0 here instead.
+        qos_policies->depth = 0;
         break;
       default:
         rmw_cyclonedds_cpp::unreachable();


clalancette:
Full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14415/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9188/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12088/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14512/)

ivanpauno (approved):
LGTM!

(ci failures seem unrelated)

jacobperron (approved):
Looks okay to me. But we should probably ticket a separate issue in ros2cli regarding handling 32-bit values. IIUC, the tool will fail if we have a KEEP_LAST history with a very large depth.

clalancette:
> (ci failures seem unrelated)

Yeah, the cpplint stuff is coming from rosbag2 and is fixed in https://github.com/ros2/rosbag2/pull/744

clalancette:
> Looks okay to me. But we should probably ticket a separate issue in ros2cli regarding handling 32-bit values. IIUC, the tool will fail if we have a KEEP_LAST history with a very large depth.

Yeah, it occurred to me too.  I think a depth of > 2 billion is unreasonable, but it still shouldn't fail.  I'll open a follow-up issue; I think we'll want to switch away from XML-RPC for this, as it just doesn't seem up to the job.

eboasson:
> > Looks okay to me. But we should probably ticket a separate issue in ros2cli regarding handling 32-bit values. IIUC, the tool will fail if we have a KEEP_LAST history with a very large depth.
> 
> Yeah, it occurred to me too. I think a depth of > 2 billion is unreasonable, but it still shouldn't fail. I'll open a follow-up issue; I think we'll want to switch away from XML-RPC for this, as it just doesn't seem up to the job.

In general, yes; but as the DDS spec has a signed 32-bit integer for the depth (don't know why, I suspect Java) and as 2 billion really is a pretty unreasonable value, I think it is highly unlikely there is any code out there that tries to use a depth ≥2^31. I think you could also quite reasonably limit the depth in ROS 2 to 2^31-1 without causing any problems.

clalancette:
> In general, yes; but as the DDS spec has a signed 32-bit integer for the depth (don't know why, I suspect Java) and as 2 billion really is a pretty unreasonable value, I think it is highly unlikely there is any code out there that tries to use a depth ≥2^31. I think you could also quite reasonably limit the depth in ROS 2 to 2^31-1 without causing any problems.

That's fair.  I'm going to reproduce this comment over in https://github.com/ros2/ros2cli/issues/629, so when we consider fixing this in the future we have this datapoint.  Thanks for the comment.

msmcconnell:
Is there any chance this can be backported to Foxy? Or if someone is aware of how to get access to the QoS settings of a topic when using cyclone that would work for me as well.

clalancette:
> Is there any chance this can be backported to Foxy? Or if someone is aware of how to get access to the connections and QoS settings of a topic when using cyclone that would work for me as well.

Sure, this is an easy backport.

@Mergifyio backport foxy

clalancette:
@Mergifyio backport foxy

msmcconnell:
> Sure, this is an easy backport.

Thanks!

PR #306, opened 2021-04-22
Update usage of dds_data_allocator APIs

sumanth-nirmal:
This MR updates the usage of the `dds_data_allocator` APIs to be in sync with Cyclone DDS

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -2307,7 +2307,7 @@ static rmw_ret_t borrow_loaned_message_int(
   // if the publisher can loan
   if (cdds_publisher->is_loaning_available) {
     // initialize the data allocator
-    dds_data_allocator_init(cdds_publisher->enth, &cdds_publisher->data_allocator);
+    dds_data_allocator_init(cdds_publisher->enth, &cdds_publisher->data_allocator, false);
     // allocate memory for message + header
     auto sample_size = rmw_cyclonedds_cpp::get_message_size(type_support);
     auto chunk_size = DETERMINE_ICEORYX_CHUNK_SIZE(sample_size);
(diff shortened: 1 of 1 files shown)


eboasson (approved):
Thanks @sumanth-nirmal.

@clalancette, @ivanpauno this PR is one of the two options of undoing a mistake on my side earlier today: I merged a PR in Cyclone that I shouldn't have merged yet. I belatedly realised that while you're building from tags for the release process, rolling ros2.repos is currently using HEAD from the `iceoryx` branch of Cyclone. So as things are *right now*, a build of ROS 2 based on https://github.com/ros2/ros2/blob/master/ros2.repos fails.

I see two options: one is to merge this PR in the next few hours; the other is for me to temporarily revert the change on cyclonedds. I obviously prefer the former, but if you prefer the latter, I'll do that.

ivanpauno:
> I see two options: one is to merge this PR in the next few hours; the other is for me to temporarily revert the change on cyclonedds. I obviously prefer the former, but if you prefer the latter, I'll do that.

Could you revert that commit for the moment?
We haven't created galactic branches yet and it would be better to not break API at this point.

If the PR in cyclonedds was fixing a bug and there's not other way to fix it that breaking API then we could consider to keep the commit, but that doesn't seem to be the case.

Thanks!

clalancette:
Another alternative is to pin the ros2.repos to 0.8.0beta4 tag for cyclonedds.  That will make the released packages match the from-source builds.

As @ivanpauno, we don't really want to change API at this point in Galactic, so I'm for either reverting the commits in cyclonedds or pinning the version used in ros2.repos.

eboasson:
@clalancette @ivanpauno I'll revert it for the timing being. I don't think the argument that API changes are undesirable at this point in time is really applicable (this isn't an API change in terms of ROS 2) but I do see the complexity of trying to release so many moving parts at the same time and that this sort of churn is something you need to avoid.

@sumanth-nirmal we'll find a way that avoids any complications 🙂

emersonknapp:
Can you please revert immediately? This is effectively blocking all Galactic development

eboasson:
https://github.com/eclipse-cyclonedds/cyclonedds/pull/772

I foresee no complications in the review process as I already had a quick conversation with @k0ekk0ek on the subject

eboasson:
It's been reverted.

clalancette:
@eboasson Thanks.  Should we close this PR now, given that it no longer matches the API in cyclonedds proper?

ivanpauno:
Thanks for the fast response @eboasson !!!

> Should we close this PR now, given that it no longer matches the API in cyclonedds proper?

The change can be reintroduced after the `galactic` branch is created, so I think we can keep this open.

eboasson:
@sumanth-nirmal with https://github.com/eclipse-cyclonedds/cyclonedds/commit/8f5d8b6d24a97e646ec7be0d1636074d90ee92fd this PR is no longer relevant, right?

PR #307, opened 2021-04-25
Fix Publishing when SHM is enabled

### LINKED ISSUE #301: Subscriber receives empty sample when using shared memory and no loaned messages
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source
- Version or commit hash:
  - 0.22.1
- DDS implementation:
  - Cyclone 0.8.0 
  - rclcpp

#### Steps to reproduce issue
Enable the shared memory option. Publisher is sending a fixed-size message with the regular publish API (no borrowed message)

#### Expected behavior
Subscriber in another process receives the sample with the sent payload

#### Actual behavior
Subscriber receives an empty sample 

#### Additional information
It seems that the message size (iox_size) is not properly set for the topic. So Cyclone does a memcpy with size 0

sumanth-nirmal:
This PR should fix #301


eboasson:
Closed via #309

PR #308, opened 2021-04-25
Fix serialized path when SHM is enabled

### LINKED ISSUE #302: How to deal with serialized messages and iceoryx shared memory transport?
#### Feature description
It seems that the combination of publishing and taking serialized messages and a zero-copy shared memory transport is currently not supported. This should also be possible

#### Implementation considerations
1. Option A
- Sending the serialized message via the Cyclone shared memory transport
- The subscription needs to know whether a serialized or non-serialized message was received
- The subscription just can pass the received serialized message when  `rmw_take_serialized_message()` is used
- The subscription has to do a de-serialization if the normal `rmw_take()` is used
- Cyclone DDS has to know the size of the serialized message when write() is called (maybe different from the topic size)
- One copy when sending the message as we have to copy it into the shared memory
- No copy when the subscription takes a serialized message
- One copy when the subscription takes the non-serialized message as we have to de-serialize it

2. Option B  
- The rmw layer always passes non-serialized messages to Cyclone
- One copy with a de-serialization step for sending the message. This is done when copying the message into the shared memory and is needed anyway as there is no combination of loaned and serialized messages, right?
- No copy when the subscription takes the non-serialized message
- One copy when the subscription takes a serialized message as we have to serialize it again

sumanth-nirmal:
This should fix #302

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -354,6 +354,7 @@ struct CddsSubscription : CddsEntity
 {
   rmw_gid_t gid;
   dds_entity_t rdcondh;
+  rosidl_message_type_support_t type_supports;
 #ifdef DDS_HAS_SHM
   dds_data_allocator_t data_allocator;
 #endif  // DDS_HAS_SHM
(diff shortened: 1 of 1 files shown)


eboasson:
Closed via #309

PR #310, opened 2021-04-25
Remove ubuntu-18.04 job from CI config matrix and ignore rmw_connextdds

christophebedard:
There's no real reason to have it. It's not officially supported for Rolling.

Also, ignore `rmw_connextdds` in CI (just like `rmw_connext`). Note that if you want to keep the `ubuntu-18.04` job, this might fix it.

Finally, I bumped `action-ros-ci` to `v0.2`. This fixes the issue for Windows, although there seems to be another issue. Still an improvement though!

### DIFF
--- a/.github/resources/local.repos
+++ b/.github/resources/local.repos
@@ -5,7 +5,6 @@ definitions:
 
 repositories:
   ros2/rosidl_typesupport_connext/COLCON_IGNORE: *empty_repo
-  ros2/rmw_connext/COLCON_IGNORE: *empty_repo
   ros2/rmw_connextdds/COLCON_IGNORE: *empty_repo
 
   ros2/rosidl_typesupport_fastrtps/COLCON_IGNORE: *empty_repo
--- a/.github/resources/local.repos
+++ b/.github/resources/local.repos
@@ -4,7 +4,6 @@ definitions:
     url: data:application/zip;base64,UEsFBgAAAAAAAAAAAAAAAAAAAAAAAA==
 
 repositories:
-  ros2/rosidl_typesupport_connext/COLCON_IGNORE: *empty_repo
   ros2/rmw_connextdds/COLCON_IGNORE: *empty_repo
 
   ros2/rosidl_typesupport_fastrtps/COLCON_IGNORE: *empty_repo
--- a/.github/resources/local.repos
+++ b/.github/resources/local.repos
@@ -6,6 +6,7 @@ definitions:
 repositories:
   ros2/rosidl_typesupport_connext/COLCON_IGNORE: *empty_repo
   ros2/rmw_connext/COLCON_IGNORE: *empty_repo
+  ros2/rmw_connextdds/COLCON_IGNORE: *empty_repo
 
   ros2/rosidl_typesupport_fastrtps/COLCON_IGNORE: *empty_repo
   ros2/rmw_fastrtps/COLCON_IGNORE: *empty_repo
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,7 +31,7 @@ jobs:
       run: git config --global --add remote.origin.fetch '+refs/pull/*:refs/remotes/origin/pull/*'
     - name: Build and test ROS
       id: ros_ci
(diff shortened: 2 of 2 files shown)


christophebedard:
for some reason, connext just doesn't want to be ignored :sweat_smile: I used a workaround for now

eboasson (approved):
LGTM, but I am don't consider myself sufficiently knowledgeable on this particular subject to consider my own approval sufficient. In any case, there's no point in having platforms in the list that are no longer officially supported.

> for some reason, connext just doesn't want to be ignored

typical ...

.github/resources/local.repos
```diff
@@ -6,6 +6,7 @@ definitions:
 repositories:
   ros2/rosidl_typesupport_connext/COLCON_IGNORE: *empty_repo
   ros2/rmw_connext/COLCON_IGNORE: *empty_repo
```
clalancette, line 8: While you are in here, you may as well remove the `rmw_connext` line; we'll no longer check it out by default.

.github/workflows/CI.yml
```diff
@@ -27,11 +27,14 @@ jobs:
       run: sed -e 's/azure.archive.ubuntu.com/us.archive.ubuntu.com/g' -e t -e d /etc/apt/sources.list | sudo tee /etc/apt/sources.list.d/nonazure.list
     - name: Acquire ROS dependencies
       uses: ros-tooling/setup-ros@master
+      with:
+        # Install to avoid getting stuck on the license agreement prompt
+        install-connext: true
```
clalancette, line 32: It really shouldn't be necessary if you aren't building Connext.  I'd be interested to see what errors you got.  Nevertheless, this isn't blocking.

.github/resources/local.repos
christophebedard: oh, right, removed in 7f7784bdff6be1ceccec2291b7b39bd8b6050381

clalancette (approved):
Windows is still failing, but it doesn't look to be because of this PR.

.github/workflows/CI.yml
christophebedard: I've had this problem before, but I fixed it in the action by adding `--skip-keys rti-connext-dds-5.3.1`. However, in this case, `rosdep install` won't skip `rti-connext-dds-5.3.1`, and it gets stuck on the license agreement prompt, since there's no way (that I know of) to accept the license when installing through `rosdep` (the `RTI_NC_LICENSE_ACCEPTED` env var doesn't work).

I had a hunch and tried removing the `COLCON_IGNORE` entries for `rmw_connext*`/`rosidl_typesupport_connext` (and removing the workaround), and `rosdep install` does correctly skip `rti-connext-dds-5.3.1`: https://github.com/christophebedard/rmw_cyclonedds/pull/4/checks?check_run_id=2438697165.

The `COLCON_IGNORE` trick makes `colcon` ignore the source packages here when running `colcon list --paths-only` before passing those paths to `rosdep install --from-paths`: https://github.com/ros-tooling/action-ros-ci/blob/af55a0e0a72a3343a161caa55e93b7731bf36bdc/src/action-ros-ci.ts#L195. `rosdep` then sees a dependency on connext (because of `rmw_implementation`), and installs it, ignoring the `--skip-keys`. Maybe I don't actually understand how `rosdep` works.

I'm now trying to make `colcon build` ignore a list of packages through the `colcon-defaults` input for `action-ros-ci`. I can open another PR for that though if you want.

christophebedard: > I'm now trying to make `colcon build` ignore a list of packages through the `colcon-defaults` input for `action-ros-ci`.

it's not as straightforward as I thought it would be, so I'll just try again later and open a follow-up PR if it does end up working. For now this will do!

.github/resources/local.repos
christophebedard: I also removed the line for `rosidl_typesupport_connext` since that was also removed from the `ros2.repos` file: 0197ba6a285c0dfca3d29fcfa5baee1bf4c723a7

christophebedard:
@clalancette @eboasson this is ready to be merged

PR #313, opened 2021-05-06
Handle allocation errors during message deserialization

hidmic:
Precisely what the title says. This should address the test failures in nightlies (e.g. [here](https://ci.ros2.org/view/nightly/job/nightly_linux_debug/lastCompletedBuild/testReport/(root)/rcl_action/test_action_communication__rmw_cyclonedds_cpp_gtest_missing_result/)).

CI up to `rmw_cyclonedds_cpp` and `rcl_action`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14579/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9334/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12243/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14706/)

This isn't a great fix though. I found error handling to be somewhat inconsistent in these source files. Sometimes exceptions are thrown, sometimes an `rmw` error is set and a falsy boolean is returned, sometimes errors are ignored. I couldn't find a pattern.

### DIFF
--- a/rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
+++ b/rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
@@ -75,6 +75,34 @@ align_int_(size_t __align, T __int) noexcept
   return (__int - 1u + __align) & ~(__align - 1);
 }
 
+template<typename MembersType>
+void resize_field(
+  const MembersType * member,
+  void * field,
+  size_t size)
+{
+  if (!member->resize_function) {
+    throw std::runtime_error("unexpected error: resize function is null");
+  }
+
+  member->resize_function(field, size);
+}
+
+template<>
+inline void resize_field<rosidl_typesupport_introspection_c__MessageMember>(
+  const rosidl_typesupport_introspection_c__MessageMember * member,
+  void * field,
+  size_t size)
+{
+  if (!member->resize_function) {
+    throw std::runtime_error("unexpected error: resize function is null");
+  }
+
+  if (!member->resize_function(field, size)) {
+    throw std::runtime_error("unable to resize field");
+  }
+}
+
 template<typename T>
 void deserialize_field(
   const rosidl_typesupport_introspection_cpp::MessageMember * member,
@@ -125,7 +153,7 @@ inline void deserialize_field<std::wstring>(
       size = static_cast<uint32_t>(member->array_size_);
     } else {
       deser >> size;
-      member->resize_function(field, size);
+      resize_field(member, field, size);
     }
     for (size_t i = 0; i < size; ++i) {
       void * element = member->get_function(field, i);
@@ -150,7 +178,9 @@ void deserialize_field(
     auto & data = *reinterpret_cast<typename GenericCSequence<T>::type *>(field);
     int32_t dsize = 0;
     deser >> dsize;
-    GenericCSequence<T>::init(&data, dsize);
+    if (!GenericCSequence<T>::init(&data, dsize)) {
+      throw std::runtime_error("unable initialize generic sequence");
+    }
     deser.deserializeA(reinterpret_cast<T *>(data.data), dsize);
   }
 }
@@ -295,11 +325,7 @@ bool TypeSupport<MembersType>::deserializeROSmessage(
               array_size = member->array_size_;
             } else {
               array_size = deser.deserialize_len(1);
-              if (!member->resize_function) {
-                RMW_SET_ERROR_MSG("unexpected error: resize function is null");
-                return false;
-              }
-              member->resize_function(field, array_size);
+              resize_field(member, field, array_size);
             }
 
             if (array_size != 0 && !member->get_function) {


rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
```diff
@@ -75,6 +75,34 @@ align_int_(size_t __align, T __int) noexcept
...
+void resize_field(
+  const MembersType * member,
+  void * field,
+  size_t size)
+{
+  if (!member->resize_function) {
+    throw std::runtime_error("unexpected error: resize function is null");
+  }
+
+  member->resize_function(field, size);
```
ivanpauno, line 88: so this code is only used for cpp introspection typesupport, right?

rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
```diff
@@ -75,6 +75,34 @@ align_int_(size_t __align, T __int) noexcept
...
 }
 
+template<typename MembersType>
+void resize_field(
+  const MembersType * member,
+  void * field,
+  size_t size)
+{
+  if (!member->resize_function) {
+    throw std::runtime_error("unexpected error: resize function is null");
```
ivanpauno, line 85: if we're throwing here, we have to make sure to catch this exception everywhere else, as we cannot throw an exception through the cpp/c "boundary"

rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
hidmic: Correct. In retrospect, I could use overloads instead of template specialization.

rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
hidmic: I agree with that. Every invocation of `deserializeROSmessage` in a C API implementation is in a try-catch block (see [here](https://github.com/ros2/rmw_cyclonedds/blob/8857f2ddead5bfc84d64207bc72bf75c1928a5cd/rmw_cyclonedds_cpp/src/rmw_node.cpp#L1459) and [here](https://github.com/ros2/rmw_cyclonedds/blob/8857f2ddead5bfc84d64207bc72bf75c1928a5cd/rmw_cyclonedds_cpp/src/rmw_node.cpp#L1459)), besides checking for falsy return values. Any new code that uses these APIs has to be careful about it.

rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
ivanpauno: > Correct. In retrospect, I could use overloads instead of template specialization.

:+1: that sounds better

rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
hidmic: Those two snippets were the same :sweat_smile:. But you're right. One isn't.

hidmic: See https://github.com/ros2/rmw_cyclonedds/pull/313/commits/694364f3fea275b7c99bef04f5de250f55606608.

rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
hidmic: See https://github.com/ros2/rmw_cyclonedds/pull/313/commits/6a5b158025a264b1e397765d185c393ce310614c.

hidmic:
Re-running CI up to `rcl_action` and `rmw_cyclonedds_cpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14580/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9335/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12244/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14707/)

rmw_cyclonedds_cpp/src/TypeSupport_impl.hpp
ivanpauno: > Those two snippets were the same

hahaha, I checked both like if they were different :joy:

ivanpauno (approved):
LGTM!

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1474,6 +1474,9 @@ extern "C" rmw_ret_t rmw_deserialize(
   } catch (rmw_cyclonedds_cpp::Exception & e) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING("rmw_serialize: %s", e.what());
     ok = false;
+  } catch (std::runtime_error & e) {
```
clalancette, line 1477: Nit:

```suggestion
  } catch (const std::runtime_error & e) {
```

clalancette (approved):
I left one nit.  It's not blocking, so I'll approve anyway.  I'll let you decide if it is better to give the compiler a hint here, or to be more consistent with the existing code.

rmw_cyclonedds_cpp/src/rmw_node.cpp
hidmic: I think I'll pick consistency. It'd be good to flesh out this details though (along with error handling normalization). FYI @eboasson.

hidmic:
Github Actions seem to be broken for unrelated reasons. Merging.

PR #315, opened 2021-06-02
Compatibility with older versions of Cyclone DDS

eboasson:
The dds_data_allocator_init_heap function was added to Cyclone
specifically to cover some use cases in ROS 2 in a backwards-compatible
way.  It now turns out that building an updated Cyclone DDS and RMW
layer in a workspace overlaid on a binary installation fails because the
newly built libraries end up behind the binary installation in the
include and linker paths.  This happens because the binary installation
has all packages in the same prefix (the "colcon build --merge-install"
kind) and some of the other dependencies adding the binary install
directories ahead of the newly built ones in the include paths.  I don't
think this can be fixed in colcon or CMake, but it can be worked around
by avoiding the new definitions.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1163,11 +1163,20 @@ static void * init_and_alloc_sample(
 {
   // initialise the data allocator
   if (alloc_on_heap) {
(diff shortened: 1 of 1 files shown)


clalancette:
@eboasson Do we still need this PR?

eboasson:
@clalancette I think not. Closing it.

PR #317, opened 2021-06-03
Change index.ros.org -> docs.ros.org links.

clalancette:


eboasson (approved):
LGTM, it does what it says on tin and the new URLs work. I actually like the new URLs better than the old ones, too 🙂

clalancette:
Thanks.  I'm not actually going to run CI here, as it won't tell us anything.  Going ahead and merging, thanks for the review!

PR #320, opened 2021-06-09
Fix typesupport bug.

EvanLiUAV:
Currently when I was tried to use [rules_ros2](https://github.com/mvukov/rules_ros2) in my bazel project. we found an issue as below  [No msg shown in listener logging.](https://github.com/mvukov/rules_ros2/issues/1)
The talker output:

### DIFF
--- a/rmw_cyclonedds_cpp/src/serdata.cpp
+++ b/rmw_cyclonedds_cpp/src/serdata.cpp
@@ -51,15 +51,16 @@ using ResponseTypeSupport_cpp = rmw_cyclonedds_cpp::ResponseTypeSupport<
   rosidl_typesupport_introspection_cpp::ServiceMembers,
   rosidl_typesupport_introspection_cpp::MessageMembers>;
 
-static bool using_introspection_c_typesupport(const char * typesupport_identifier)
-{
-  return typesupport_identifier == rosidl_typesupport_introspection_c__identifier;
+static bool using_introspection_c_typesupport(const char* typesupport_identifier) {
+  return !std::string(typesupport_identifier)
+      .compare(rosidl_typesupport_introspection_c__identifier);
 }
 
-static bool using_introspection_cpp_typesupport(const char * typesupport_identifier)
-{
-  return typesupport_identifier == rosidl_typesupport_introspection_cpp::typesupport_identifier;
+static bool using_introspection_cpp_typesupport(const char* typesupport_identifier) {
+  return !std::string(typesupport_identifier)
+      .compare(rosidl_typesupport_introspection_cpp::typesupport_identifier);
 }
+
 
 void * create_message_type_support(
   const void * untyped_members,


eboasson:
@ivanpauno I find this an interesting bug. I know there are ways in which the addresses of the two strings being compared can end up being different (though I don't know exactly why it is happening with this bazel build) and I think the proposed solution is a sensible one.

The only downside I can think of is that the check happens also when (de)serializing or freeing a message, but I doubt it would cause a noticeable slowdown. I think the information could be cached in the `ddsi_sertype` to avoid having to do the string compare often.

Another issue is that I scanned the sources and it looks like the dynamic versions of the fastrtps and connext RMW layers do the same thing, so those would presumably have to be changed as well.

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -51,16 +51,17 @@ using ResponseTypeSupport_cpp = rmw_cyclonedds_cpp::ResponseTypeSupport<
   rosidl_typesupport_introspection_cpp::ServiceMembers,
   rosidl_typesupport_introspection_cpp::MessageMembers>;
 
-static bool using_introspection_c_typesupport(const char * typesupport_identifier)
-{
-  return typesupport_identifier == rosidl_typesupport_introspection_c__identifier;
+static bool using_introspection_c_typesupport(const char* typesupport_identifier) {
+  return !std::string(typesupport_identifier)
```
ivanpauno, line 55: can we use `std::strcmp()`?
To avoid allocating a `std::string` unnecessarely.

ivanpauno (approved):
>  I think the information could be cached in the ddsi_sertype to avoid having to do the string compare often.

Yes, that sounds like a good idea.
Though I don't mind to introduce this fix first.

> Another issue is that I scanned the sources and it looks like the dynamic versions of the fastrtps and connext RMW layers do the same thing, so those would presumably have to be changed as well.

@EvanLiUAV could you open a similar issue in [rmw_fastrtps](https://github.com/ros2/rmw_fastrtps) and [rmw_connextdds](https://github.com/ros2/rmw_connextdds)?

MiguelCompany:
> Another issue is that I scanned the sources and it looks like the dynamic versions of the fastrtps and connext RMW layers do the same thing, so those would presumably have to be changed as well.

In the case of `rmw_fastrtps_dynamic_cpp`, we are keeping the `const char*` pointer returned by `get_message_typesupport_handle`, which should be the pointer exported by the introspection type support package ([this one](https://github.com/ros2/rosidl/blob/155ba4f237cb483000a1261f76fa3a0c6b67f24f/rosidl_typesupport_introspection_cpp/include/rosidl_typesupport_introspection_cpp/identifier.hpp#L24) for instance).

I think the pointer should not change (at least in the case of `rmw_fastrtps_dynamic_cpp`)

ivanpauno:
> > Another issue is that I scanned the sources and it looks like the dynamic versions of the fastrtps and connext RMW layers do the same thing, so those would presumably have to be changed as well.
> 
> In the case of `rmw_fastrtps_dynamic_cpp`, we are keeping the `const char*` pointer returned by `get_message_typesupport_handle`, which should be the pointer exported by the introspection type support package ([this one](https://github.com/ros2/rosidl/blob/155ba4f237cb483000a1261f76fa3a0c6b67f24f/rosidl_typesupport_introspection_cpp/include/rosidl_typesupport_introspection_cpp/identifier.hpp#L24) for instance).
> 
> I think the pointer should not change (at least in the case of `rmw_fastrtps_dynamic_cpp`)

The same happens here.
I'm not sure what @EvanLiUAV is exactly doing, but there might be some linker issues were the global doesn't exactly have the same address.

IMO doing a pointer comparison is unnecessarily fragile, comparing the strings seem more reasonable and can be cached when the subscription is created if there's any performance concern.

PR #322, opened 2021-06-22
Added shared memory support usage description

eboasson:
"Backport" from master.


clalancette (approved):

clalancette:
(since this one is documentation, I would say it is safe enough to just merge into Galactic without additional reviews.  But FYI @cottsay).

PR #343, opened 2021-10-12
Add client/service QoS getters

mauropasse:
Signed-off-by: Mauro Passerino <mpasserino@irobot.com>

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -5095,86 +5095,70 @@ extern "C" rmw_ret_t rmw_qos_profile_check_compatible(
     publisher_profile, subscription_profile, compatibility, reason, reason_size);
 }
 
-extern "C" rmw_ret_t rmw_client_get_actual_qos(
-  const rmw_client_t * client,
-  rmw_qos_profile_t * qos)
+extern "C" rmw_ret_t rmw_client_request_publisher_get_actual_qos(
+    const rmw_client_t * client,
+    rmw_qos_profile_t * qos)
 {
   RMW_CHECK_ARGUMENT_FOR_NULL(client, RMW_RET_INVALID_ARGUMENT);
   RMW_CHECK_ARGUMENT_FOR_NULL(qos, RMW_RET_INVALID_ARGUMENT);
 
-  auto info = static_cast<CddsClient *>(client->data);
+  auto cli = static_cast<CddsClient *>(client->data);
 
-  rmw_qos_profile_t pub_qos = rmw_qos_profile_default;
-  rmw_qos_profile_t sub_qos = rmw_qos_profile_default;
-
-  // Check if the QoS of the client's pub/sub match
-  if (get_readwrite_qos(info->client.pub->enth, &pub_qos)) {
-    if (get_readwrite_qos(info->client.sub->enth, &sub_qos)) {
-      if (pub_qos.history != sub_qos.history ||
-        pub_qos.depth != sub_qos.depth ||
-        pub_qos.reliability != sub_qos.reliability ||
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -5098,3 +5098,87 @@ extern "C" rmw_ret_t rmw_qos_profile_check_compatible(
...
+        pub_qos.deadline.sec == sub_qos.deadline.sec &&
+        pub_qos.deadline.nsec == sub_qos.deadline.nsec &&
+        pub_qos.lifespan.sec == sub_qos.lifespan.sec &&
+        pub_qos.lifespan.nsec == sub_qos.lifespan.nsec &&
+        pub_qos.liveliness_lease_duration.sec == sub_qos.liveliness_lease_duration.sec &&
+        pub_qos.liveliness_lease_duration.nsec == sub_qos.liveliness_lease_duration.nsec)
+      {
+        // The client has a single QoS, we can set it either as the qos
+        // of the publisher or subscriber, as they match.
+        get_readwrite_qos(info->client.pub->enth, qos);
```
eboasson, line 5131: Nitpick: you already have a copy of the QoS settings in `pub_qos` and `sub_qos`, so would it not make more sense to copy either of those instead?

There is also the (remote) possibility that this call to `get_readwrite_qos` fails: if some other thread calls `rmw_destroy_client` between line 5117 and this line, it might. Not calling it avoids having to consider the possibility and whether it should be taken seriously.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -5098,3 +5098,87 @@ extern "C" rmw_ret_t rmw_qos_profile_check_compatible(
...
+
+  auto info = static_cast<CddsService *>(service->data);
+
+  rmw_qos_profile_t pub_qos = rmw_qos_profile_default;
+  rmw_qos_profile_t sub_qos = rmw_qos_profile_default;
+
+  // Check if the QoS of the service's pub/sub match
+  if (get_readwrite_qos(info->service.pub->enth, &pub_qos)) {
+    if (get_readwrite_qos(info->service.sub->enth, &sub_qos)) {
+      if (pub_qos.history == sub_qos.history &&
```
eboasson, line 5159: And another nitpick: would it not make more sense to use a single implementation for testing RMW QoS objects for equality? I'd think it most elegant if `operator==` were to be defined on `rmw_qos_profile_t` and then use that, but I am not sure if that's realistic, considering that the RMW interface is a C one, even if all the code that uses `rmw.h` appears to be C++ code.

eboasson (commented):
Thanks @mauropasse, it looks good but there are two details where I'd like to ask you to reconsider the implementation.

rmw_cyclonedds_cpp/src/rmw_node.cpp
mauropasse: Right, I'm copying the QoS settings now (last commit 1acdd5c)

eboasson:
@mauropasse, @ivanpauno I figured I should wait with merging until all the pieces of the puzzle are in place, but I also know one cannot finish a puzzle without laying a first piece. If no coordination is needed, perhaps it would be best to just merge this one?

mauropasse:
After some discussion (see https://github.com/ros2/rmw_connextdds/pull/67#discussion_r738447092) I updated all PRs to retrieve independently the QoS for Client request publisher / response subscription and Services request subscription / response publisher.

ivanpauno (approved):
lgtm

ivanpauno:
CI https://github.com/ros2/rmw_implementation/pull/196#issuecomment-956442206

PR #357, opened 2021-12-11
Export modern CMake targets

sloretz:
Part of ament/ament_cmake#365

This makes rmw_cyclonedds_cpp export modern CMake targets. Currently it only makes old-style standard variables available.

PR #360, opened 2021-12-15
Export only rmw::rmw to downstream targets

sloretz:
Required for ros2/rmw_implementation#201 which fixes a rolling CI job failure caused by ros2/rmw_cyclonedds#357

Since this is an rmw implementation, it's implementing the `rmw` API. I think `rmw::rmw` is the only target that needs to be marked `PUBLIC`. The rest downstream targets don't need to know about, and exporting them causes failures when the targets are from packages that don't get `find_package()`'d, notably with [`tracetools::tracetools`.](http://ci.ros2.org/job/ci_linux-aarch64/10464/).

### DIFF
--- a/rmw_cyclonedds_cpp/CMakeLists.txt
+++ b/rmw_cyclonedds_cpp/CMakeLists.txt
@@ -70,11 +70,12 @@ target_link_libraries(rmw_cyclonedds_cpp PRIVATE
   CycloneDDS::ddsc)
 
 target_link_libraries(rmw_cyclonedds_cpp PUBLIC
+  rmw::rmw)
+target_link_libraries(rmw_cyclonedds_cpp PRIVATE
   rcutils::rcutils
   rcpputils::rcpputils
   rosidl_typesupport_introspection_c::rosidl_typesupport_introspection_c
   rosidl_typesupport_introspection_cpp::rosidl_typesupport_introspection_cpp
-  rmw::rmw
   ${rmw_dds_common_TARGETS}
   rosidl_runtime_c::rosidl_runtime_c
   tracetools::tracetools)


sloretz:
CI - build and test everything
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15788/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10466/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13450/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16045/)

rmw_cyclonedds_cpp/CMakeLists.txt
```diff
@@ -70,11 +70,12 @@ target_link_libraries(rmw_cyclonedds_cpp PRIVATE
...
 
 target_link_libraries(rmw_cyclonedds_cpp PUBLIC
+  rmw::rmw)
+target_link_libraries(rmw_cyclonedds_cpp PRIVATE
   rcutils::rcutils
   rcpputils::rcpputils
   rosidl_typesupport_introspection_c::rosidl_typesupport_introspection_c
   rosidl_typesupport_introspection_cpp::rosidl_typesupport_introspection_cpp
-  rmw::rmw
   ${rmw_dds_common_TARGETS}
```
clalancette, line 79: While we are in here, should we change this to `rmw_dds_common::rmw_dds_common`?

clalancette (approved):
I left a nit inline, but I'll approve anyway.  You can decide whether to change it or not.

rmw_cyclonedds_cpp/CMakeLists.txt
sloretz: I'd like to do something like that, but that package also exports rosidl generated typesupport targets. Right now there isn't a great way to depend on those other than using `<package name>_TARGETS`

https://github.com/ros2/rmw_dds_common/blob/42d86b736a7292ae0ab926f818afc28e3d64aebd/rmw_dds_common/CMakeLists.txt#L26-L31

sloretz:
CI LGTM because windows rosbag2 test failures are already in the nightly: https://ci.ros2.org/view/nightly/job/nightly_win_rel/2149/

PR #363, opened 2022-01-07
Fix cpplint errors

jacobperron:
Relates to https://github.com/ament/ament_lint/pull/324

PR #365, opened 2022-01-10
Fix freeing of loaned chunks 

### LINKED ISSUE #364: Shared memory chunks are lost when the subscriber callback is delayed
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - From source with this [repos file](https://github.com/ros2/ros2/blob/master/ros2.repos)
- Version or commit hash:
  - ROS2 Rolling
  - `rmw_cyclonedds` version - 1.1.1 cb129b5859e4c1793757d71f521540ae67a4fb20
- DDS implementation:
  - `rmw_cyclonedds_cpp`
  - `rclcpp`

#### Steps to reproduce issue

1. Set up a pub/sub example with all the required SHM constraints
2. Make the subscriber callback significantly time-consuming
3. After some iterations, the shared memory chunks pile up and eventually leading to"out of chunks" error.

#### Actual behavior

When the subscriber callback is expensive and takes significant time to complete, the shared memory chunks are lost

#### Expected behavior

The shared memory chunks shouldn't be lost in scenarios where the subscriber callbacks take significant time to execute and the chunks should be freed properly.

#### Additional information

The probable reasoning for why this happens is as follows, when shared memory is enabled, the flow for executing the subscription is first the loaned message is taken, then the callback is executed with the taken loaned message, then once the callback execution is completed the loaned message is returned back to the middleware. However, if the callback is expensive and takes significant time, then in the mean time there are samples also received in the reader history cache, and based on the history depth the samples in the reader history cache will be lost without being processed by the executor. Ideally, when the sample is removed from the reader history cache, the corresponding loan for this sample should also be returned to the middleware, which is not handled correctly in `serdata_rmw_free`. 

The essence is in `serdata_rmw_free` the loan is not returned to the middleware if there is an outstanding loan.

sumanth-nirmal:
Fixes #364

### DIFF
--- a/rmw_cyclonedds_cpp/CMakeLists.txt
+++ b/rmw_cyclonedds_cpp/CMakeLists.txt
@@ -33,7 +33,16 @@ find_package(tracetools REQUIRED)
 
 #find_package(cyclonedds_cmake_module REQUIRED)
 find_package(CycloneDDS QUIET CONFIG)
-if(NOT CycloneDDS_FOUND)
+if(CycloneDDS_FOUND)
+  # Support for shared memory in RMW depends on support for shared memory being compiled in
+  # Cyclone DDS and iceoryx_binding_c being available
+  get_target_property(cyclonedds_has_shm CycloneDDS::ddsc SHM_SUPPORT_IS_AVAILABLE)
+  if(cyclonedds_has_shm)
+    find_package(iceoryx_binding_c REQUIRED)
+  else()
+    message(STATUS "Cyclone DDS is NOT compiled with support for shared memory")
+  endif()
+else()
   message(WARNING "Could not find Eclipse Cyclone DDS - skipping '${PROJECT_NAME}'")
   ament_package()
   return()
@@ -68,6 +77,10 @@ add_library(rmw_cyclonedds_cpp
 
 target_link_libraries(rmw_cyclonedds_cpp PRIVATE
   CycloneDDS::ddsc)
+
+if(cyclonedds_has_shm)
+  target_link_libraries(rmw_cyclonedds_cpp PRIVATE iceoryx_binding_c::iceoryx_binding_c)
+endif()
 
 target_link_libraries(rmw_cyclonedds_cpp PUBLIC
   rmw::rmw)
--- a/rmw_cyclonedds_cpp/CMakeLists.txt
+++ b/rmw_cyclonedds_cpp/CMakeLists.txt
@@ -36,8 +36,8 @@ find_package(CycloneDDS QUIET CONFIG)
 if(CycloneDDS_FOUND)
   # Support for shared memory in RMW depends on support for shared memory being compiled in
   # Cyclone DDS and iceoryx_binding_c being available
-  get_target_property(cyclonedds_has_shm CycloneDDS::ddsc SHM_SUPPORT_IS_AVAILABLE)
-  if(cyclonedds_has_shm)
+  get_target_property(_cyclonedds_has_shm CycloneDDS::ddsc SHM_SUPPORT_IS_AVAILABLE)
+  if(_cyclonedds_has_shm)
     find_package(iceoryx_binding_c REQUIRED)
   else()
     message(STATUS "Cyclone DDS is NOT compiled with support for shared memory")
@@ -78,7 +78,7 @@ add_library(rmw_cyclonedds_cpp
 target_link_libraries(rmw_cyclonedds_cpp PRIVATE
   CycloneDDS::ddsc)
 
-if(cyclonedds_has_shm)
+if(_cyclonedds_has_shm)
   target_link_libraries(rmw_cyclonedds_cpp PRIVATE iceoryx_binding_c::iceoryx_binding_c)
 endif()
 
--- a/rmw_cyclonedds_cpp/src/serdata.cpp
+++ b/rmw_cyclonedds_cpp/src/serdata.cpp
@@ -180,28 +180,33 @@ static struct ddsi_serdata * serdata_rmw_from_ser(
   enum ddsi_serdata_kind kind,
   const struct nn_rdata * fragchain, size_t size)
 {
-  auto d = std::make_unique<serdata_rmw>(type, kind);
-  uint32_t off = 0;
-  assert(fragchain->min == 0);
-  assert(fragchain->maxp1 >= off);    /* CDR header must be in first fragment */
-  d->resize(size);
+  try {
+    auto d = std::make_unique<serdata_rmw>(type, kind);
+    uint32_t off = 0;
+    assert(fragchain->min == 0);
+    assert(fragchain->maxp1 >= off);    /* CDR header must be in first fragment */
+    d->resize(size);
 
-  auto cursor = d->data();
-  while (fragchain) {
-    if (fragchain->maxp1 > off) {
-      /* only copy if this fragment adds data */
-      const unsigned char * payload =
-        NN_RMSG_PAYLOADOFF(fragchain->rmsg, NN_RDATA_PAYLOAD_OFF(fragchain));
-      auto src = payload + off - fragchain->min;
-      auto n_bytes = fragchain->maxp1 - off;
-      memcpy(cursor, src, n_bytes);
-      cursor = byte_offset(cursor, n_bytes);
-      off = fragchain->maxp1;
-      assert(off <= size);
+    auto cursor = d->data();
+    while (fragchain) {
+      if (fragchain->maxp1 > off) {
+        /* only copy if this fragment adds data */
+        const unsigned char * payload =
+          NN_RMSG_PAYLOADOFF(fragchain->rmsg, NN_RDATA_PAYLOAD_OFF(fragchain));
+        auto src = payload + off - fragchain->min;
+        auto n_bytes = fragchain->maxp1 - off;
+        memcpy(cursor, src, n_bytes);
+        cursor = byte_offset(cursor, n_bytes);
+        off = fragchain->maxp1;
+        assert(off <= size);
+      }
+      fragchain = fragchain->nextfrag;
     }
-    fragchain = fragchain->nextfrag;
+    return d.release();
+  } catch (std::exception & e) {
+    RMW_SET_ERROR_MSG(e.what());
+    return nullptr;
   }
-  return d.release();
 }
 
 static struct ddsi_serdata * serdata_rmw_from_ser_iov(
@@ -210,15 +215,20 @@ static struct ddsi_serdata * serdata_rmw_from_ser_iov(
   ddsrt_msg_iovlen_t niov, const ddsrt_iovec_t * iov,
   size_t size)
 {
-  auto d = std::make_unique<serdata_rmw>(type, kind);
-  d->resize(size);
+  try {
+    auto d = std::make_unique<serdata_rmw>(type, kind);
+    d->resize(size);
 
-  auto cursor = d->data();
-  for (ddsrt_msg_iovlen_t i = 0; i < niov; i++) {
-    memcpy(cursor, iov[i].iov_base, iov[i].iov_len);
-    cursor = byte_offset(cursor, iov[i].iov_len);
+    auto cursor = d->data();
+    for (ddsrt_msg_iovlen_t i = 0; i < niov; i++) {
+      memcpy(cursor, iov[i].iov_base, iov[i].iov_len);
+      cursor = byte_offset(cursor, iov[i].iov_len);
+    }
+    return d.release();
+  } catch (std::exception & e) {
+    RMW_SET_ERROR_MSG(e.what());
+    return nullptr;
   }
-  return d.release();
 }
 
 static struct ddsi_serdata * serdata_rmw_from_keyhash(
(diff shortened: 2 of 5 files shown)


sumanth-nirmal:
@eboasson @MatthiasKillat Can you please review the changes?

Currently, I am depending on iceoryx directly to free the chunk in `serdata_free` under the assumption that we can remove this dependency once we update RMW to use the latest Cyclone tag. Please let me know if it doesn't make sense.

rmw_cyclonedds_cpp/CMakeLists.txt
```diff
@@ -33,7 +33,16 @@ find_package(tracetools REQUIRED)
 
 #find_package(cyclonedds_cmake_module REQUIRED)
 find_package(CycloneDDS QUIET CONFIG)
-if(NOT CycloneDDS_FOUND)
+if(CycloneDDS_FOUND)
+  # Support for shared memory in RMW depends on support for shared memory being compiled in
+  # Cyclone DDS and iceoryx_binding_c being available
+  get_target_property(cyclonedds_has_shm CycloneDDS::ddsc SHM_SUPPORT_IS_AVAILABLE)
+  if(cyclonedds_has_shm)
+    find_package(iceoryx_binding_c REQUIRED)
```
clalancette, line 41: This will likely not work on the buildfarm as-is.  If you want to do this, you'll have to add a package.xml dependency to iceoryx.

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -244,10 +251,10 @@ static struct ddsi_serdata * serdata_rmw_from_iox(
   const struct ddsi_sertype * typecmn,
   enum  ddsi_serdata_kind kind, void * sub, void * iox_buffer)
 {
-  static_cast<void>(sub);  // unused
   const struct sertype_rmw * type = static_cast<const struct sertype_rmw *>(typecmn);
   auto d = std::make_unique<serdata_rmw>(type, kind);
   d->iox_chunk = iox_buffer;
+  d->iox_subscriber = sub;
   return d.release();
```
MatthiasKillat, line 258: Why create a unique pointer if we release it to return a raw pointer after two instructions that cannot throw (i.e. there is no benefit from RAII here)? We could create a raw pointer instead which would be slightly more efficient.

MatthiasKillat (commented):
Logic appears correct, there is some potential for minor efficiency gains (to be discussed).

eboasson (commented):
I agree with @MatthiasKillat's and @clalancette's comments, but also have one of my own that doesn't fit in a comment attached to a line in the diff. The issue is that I run into trouble mentally tracking lifetimes of things.

We construct an `rmw_serdata` with a chunk attached in the sending and in the receiving path, and because of how Iceoryx is integrated, the two are entirely separate.

For the publishing, the serdata is constructed by `publish_loaned_int` or `rmw_publish_serialized_message`. This case obviously has no `iox_subscriber` attached and ownership is transferred to Iceoryx. This eventually ends up in `serdata_rmw_free` (because all serdata's end up there) but it should leave any `iox_chunk` alone.

As a precautionary measure, `dds_write` and friends clear `iox_chunk` on handing it over to Iceoryx. `serdata_rmw_free` is invoked when its refcount has dropped to 0, so `iox_chunk` will be a null pointer by the time it ends up in `serdata_rmw_free`.

In short, the publishing case is fine, and with without this PR and @MatthiasKillat is correct that the `iox_subscriber` test in there is, strictly speaking, superfluous if that "precautionary measure" is elevated to a "contractual behaviour". Checking `iox_subscriber` avoids that and also makes it quite clear that this doesn't apply to the publishing case.

For the subscribing case, the serdata is always constructed by `serdata_rmw_from_iox`. It is dropped when:
* It is pushed out of the RHC (triggering this PR). The only function called in this case is `serdata_rmw_free` and it follows that in this case, this PR is correct in releasing the Iceoryx chunk there.
* It is `dds_take`n out of the RHC, in which case it is preceded by `serdata_rmw_to_sample` which in turn always deserializes. This doesn't affect the chunk. `dds_take` is called by the `rmw_take` family except `rmw_take_serialized` and `rmw_take_loaned_message`. I suspect this PR also fixes a leak there, but I may have overlooked something.
* When `rmw_take_serialized` drops the last reference. This one frees the `iox_chunk` but with this PR I think it should no longer do so.
* When `rmw_take_loaned_message` takes ownership of the chunk. There are two cases supported:
  * Where the serdata contains a sample. This case just returns a pointer and it is `rmw_return_loaned_message_from_subscription` that is responsible for freeing it;
  * Where the serdata contains a serialized sample. In this case the "loaned" sample in reality is a newly allocated thing. In this case the Iceoryx chunk could be freed much earlier.
  In both cases (and like above, I may have overlooked something) I don't see a call to `serdata_unref`, which suggests there may be a memory leak there. In the second case, it may involve shared memory as well.

  In either case, with this PR, I would expect a double-free if `serdata_rmw_free` were called, because `return_loaned_message` frees the Iceoryx chunk but I see no reason why `serdata_rmw_free` would not try to free it again.

The changes in this PR are good, I just think it is incomplete and that completing it will reduce the complexity in `rmw_node.cpp`, because it will no longer need to deal with actually releasing chunks of shared memory *except in one case*: that where `take_loaned_message` actually returned a pointer to shared memory and `return_loaned_message` must free that memory.

Ideally, one would track the serdata and not do any shared memory management in `return_loaned_message`, but that is impossible because of the RMW interface (unless one caches a set of outstanding loans in the subscriber). What should work is clearing `iox_chunk` and `iox_subscriber` in `rmw_take_loaned_message` when it returns a pointer to shared memory.

rmw_cyclonedds_cpp/CMakeLists.txt
sumanth-nirmal: Thanks for pointing this out, I added `iceoryx_binding_c` dependency to `Package.xml`

rmw_cyclonedds_cpp/CMakeLists.txt
```diff
@@ -33,7 +33,16 @@ find_package(tracetools REQUIRED)
 
 #find_package(cyclonedds_cmake_module REQUIRED)
 find_package(CycloneDDS QUIET CONFIG)
-if(NOT CycloneDDS_FOUND)
+if(CycloneDDS_FOUND)
+  # Support for shared memory in RMW depends on support for shared memory being compiled in
+  # Cyclone DDS and iceoryx_binding_c being available
+  get_target_property(cyclonedds_has_shm CycloneDDS::ddsc SHM_SUPPORT_IS_AVAILABLE)
```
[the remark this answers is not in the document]
sumanth-nirmal: I changed this to use the common convention used across the codebase in ROS 2

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -164,7 +164,14 @@ static uint32_t serdata_rmw_size(const struct ddsi_serdata * dcmn)
 
 static void serdata_rmw_free(struct ddsi_serdata * dcmn)
 {
-  auto * d = static_cast<const serdata_rmw *>(dcmn);
+  auto * d = static_cast<serdata_rmw *>(dcmn);
+
+#ifdef DDS_HAS_SHM
+  if(d->iox_chunk && d->iox_subscriber) {
```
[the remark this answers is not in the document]
sumanth-nirmal: I would like to have this check to be sure that the loaned chunk is returned back to iceoryx only in the subscriber case

sumanth-nirmal:
@eboasson Thanks for the comment, I did a code walkthrough for memory audit in the shared memory case and fixed a couple of memory leaks. I tried to summarize below the data path with allocations to identify if we missed anything

#### Publishing

##### `rmw_publish_serialized_message`

1. Construct the serdata from the serialized message (**serdata allocation**)
2. With `dds_data_allocator` we loan the chunk from iceoryx (**Chunk allocation**)
3. Deserialize the sample to the loaned chunk (this additional deserialization will be fixed when we update rmw to use the latest Cyclone)
4. Update the serdata with the loaned chunk
5. Hand over the serdata to the `dds_writecdr()`
    1. The loaned chunk is handover to the iceoryx  (**Chunk free**)
    2. The chunk pointer is reset to null in serdata
    3. The serdata ref count is decremented
        1. Since the ref count will be 0, the serdata will be deleted in `serdata_rmw_free` (**serdata free**)
        
- Error scenario
    1. When deserialize fails, serdata refcount is decremented, which will eventually free the serdata in `serdata_rmw_free` (**serdata free**)

##### `rmw_publish_loaned_message`

1. Construct the serdata with the placement new operator (**serdata allocation**)
2. Update the serdata with the loaned chunk (the chunk is loaned with `rmw_borrow_loaned_message`) (**Chunk allocation**)
3. Hand over the serdata to the `dds_writecdr()`
    1. The loaned chunk is handover to the iceoryx  (**Chunk free**)
    2. The chunk pointer is reset to null in serdata
    3. The serdata ref count is decremented
        1. Since the ref count will be 0, the serdata will be deleted in `serdata_rmw_free` (**serdata free**)
 
- Error scenario
   1. When `dds_writecdr()` fails, serdata refcount is decremented, which will eventually free the serdata in `serdata_rmw_free` (**serdata free**)

----

#### Subscription

##### `rmw_take_serialized_message/with_info`

1. Get serdata with `dds_takecdr()` (**serdata allocation**)
2. The serdata has valid `iox_chunk` (updated in `serdata_rmw_from_iox`) (**Chunk allocation**)
3.  Serialize the sample from the loaned chunk (this additional serialization will be fixed when we update rmw to use the latest Cyclone)
4. The serdata ref count is decremented (the ref count will be 0 after this)
    1. The `iox_chunk` will be released to the iox_subscriber in `serdata_rmw_free` (**Chunk free**) 
    2. The serdata will be freed in `serdata_rmw_free` (**serdata free**)
   
-  Error scenario
    1. When serialize fails, serdata refcount is decremented, the ref count will drop to 0 which will end in `serdata_rmw_free` 
        1. The `iox_chunk` will be released to the iox_subscriber in `serdata_rmw_free` (**Chunk free**) 
        2. The serdata will be freed in `serdata_rmw_free` (**serdata free**)
     2. When the received sample is not valid
       1. We decrement the ref count of the serdata (the ref count will be 0 after this)
           1. With SHM, the iox_chunk will be released to the iox_subscriber (**Chunk free**), and the serdata will be freed in `serdata_rmw_free` (**serdata free**)
           2. without SHM, the serdata is freed in `serdata_rmw_free` (**serdata free**)
    
##### `rmw_take_loaned_message/with_info`

1. Get serdata with `dds_takecdr()` (**serdata allocation**)
2. If the serdata has valid `iox_chunk` (updated in `serdata_rmw_from_iox`) (**Chunk allocation**)
    1. Give the reference to the chunk to the user
    2. Initialize the data_alloc with this subscription
    3. The chunk will be freed in `rmw_return_loaned_message_from_subscription()` (**Chunk free**)
    4. The `iox_chunk` pointer is reset to `null` and serdata ref count is decremented (the ref count will be 0 after this)
       1.  The serdata will be deleted in `serdata_rmw_free` (**serdata free**) \
       (the loaned chunk will not be released here as the `iox_chunk` pointer is null)
3. If the serdata doesn't have the chunk (meaning SHM is disabled on the publisher but enabled on subscription)
    1. Allocate memory for the sample on the heap (**Sample allocation**)
    2. deserialize the sample to this allocated memory 
    3. Give the reference to this allocated memory to the user
    4. The allocated memory will be freed in `rmw_return_loaned_message_from_subscription()` (**Sample free**)
    5. The serdata ref count is decremented (the ref count will be 0 after this)
       1.  The serdata will be deleted in `serdata_rmw_free` (**serdata free**)
   
- Error scenario
    1. When the received sample is not valid
       1. We decrement the ref count of the serdata (the ref count will be 0 after this)
           1. With SHM, the iox_chunk will be released to the iox_subscriber (**Chunk free**), and the serdata will be freed in `serdata_rmw_free` (**serdata free**)
           2. without SHM, the serdata is freed in `serdata_rmw_free` (**serdata free**)
----

I think with this MR, all the previously unhandled memory leaks with SHM should have been fixed (unless I missed anything)!

rmw_cyclonedds_cpp/src/serdata.cpp
sumanth-nirmal: This pattern is used in multiple places in the code base, I don't want to change this in the scope of this MR, but I have wrapped the creation of new serdata in a try-catch block to catch `std::bad_alloc`

eboasson (approved):
LGTM, it think it covers every path now. CI:

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/16063/)
* Linux aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/10716/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/16387/)

*

sumanth-nirmal:
@eboasson I see that linters fail in the CI jobs here https://github.com/ros2/rmw_cyclonedds/pull/365#pullrequestreview-871457535, but then realized that these are already fixed in this PR https://github.com/ros2/rmw_cyclonedds/pull/363. I have now rebased this branch onto the latest master and the CI should hopefully be good now. Can you please run the CI again?

eboasson:
Hi @sumanth-nirmal, I've re-run the full CI:

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/16143/)
* Linux aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/10772/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/16474/)

The Windows failure is "tf2_ros_py.pytest.missing_result" not producing a test result, which I'm pretty sure is unrelated. So I think it it is good to go.

eboasson:
@clalancette I notice that GitHub says your review was requested and I don't to go ahead and merge it if there's a possibility that you're reviewing it at the same time. As far as I am concerned, this is good to go. Please let me know (or simply press "merge" as a way of letting me know 🙂)

clalancette (approved):
Seems reasonable enough to me (though I only did a cursory look and verified that the dependencies look correct).  @eboasson Feel free to merge.

PR #369, opened 2022-01-31
[foxy] Free with the same allocator in rmw_destroy_node (#355)

jacobperron:
Backport #355 to Foxy.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1347,10 +1347,9 @@ extern "C" rmw_ret_t rmw_destroy_node(rmw_node_t * node)
   }
 
   rmw_context_t * context = node->context;
-  rcutils_allocator_t allocator = context->options.allocator;
-  allocator.deallocate(const_cast<char *>(node->name), allocator.state);
-  allocator.deallocate(const_cast<char *>(node->namespace_), allocator.state);
-  allocator.deallocate(node, allocator.state);
+  rmw_free(const_cast<char *>(node->name));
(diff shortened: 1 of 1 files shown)


eboasson (approved):
LGTM!

jacobperron:
Is it okay that we have 6 failing checks? Seems like we should disable those if we are just going to ignore them.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16052/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10707/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16377/)

eboasson:
> Is it okay that we have 6 failing checks? Seems like we should disable those if we are just going to ignore them.

I wouldn't consider it okay ... but I have never really known what it is that makes these tests work for some time, then start failing until someone tweaks "something" just so the pattern can repeat. It looks like some mismatch between versions in the CI scripting, but I am not sure. Surely there must be a way to have some level of CI integration for a stable release that doesn't require tweaking?

If I look at the Windows build it seems to be failing on installing Connext or something like that.

ivanpauno (approved):

jacobperron:
I don't have the bandwidth right now to investigate the github action failures, but following our process for other core ROS 2 packages, the FPr job is green and the results from ci.ros2.org (https://github.com/ros2/rmw_cyclonedds/pull/369#issuecomment-1027057259) look okay to me so I'm going to merge this. 

I can return later to look at the github actions; it might be better to ultimately remove them since they are redundant with ci.ros2.org or are testing unsupported platforms (e.g. 18.04 and macOS).

PR #371, opened 2022-02-03
Bump QDs to QL2

christophebedard:
And therefore requirement 5.i is fulfilled.

I also noticed that 1.ii can also be marked as done, so this results in a bump to QL2. This can't be bumped to QL1 because [`rmw_dds_common`'s QL1 requires a QL1 middleware](https://github.com/ros2/rmw_dds_common/blob/master/rmw_dds_common/QUALITY_DECLARATION.md#rmw_dds_common-quality-declaration). If `cyclonedds` is bumped from QL2 to QL1, then I think `rmw_cyclonedds_cpp` can also be bumped to QL1.

Signed-off-by: Christophe Bedard <bedard.christophe@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,4 +72,4 @@ If building ROS 2 from source ([ros2.repos](https://github.com/ros2/ros2/blob/ma
 
 ## Quality Declaration
 
-This package claims to be in the **Quality Level 4** category, see the [Quality Declaration](./rmw_cyclonedds_cpp/QUALITY_DECLARATION.md) for more details.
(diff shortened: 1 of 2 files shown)


hidmic (approved):

hidmic:
> If cyclonedds is bumped from QL2 to QL1, then I think rmw_cyclonedds_cpp can also be bumped to QL1.

That's reasonable, though perhaps 4.iii, 4.iv need a revisit before a bump to QL1.

hidmic:
Rpr job failure on Jammy is unrelated to this patch.

PR #372, opened 2022-02-07
Update CI workflow

jacobperron:
- Pin to stable versions of ros-tooling actions
- Remove overlayed version of ament_lint from workspace
  - The related PR has been merged and we should be linting against the version in Foxy anyways
- Do not test against macOS; it is no longer a supported platform for Foxy
- Pull dependencies from the testing repository

### DIFF
--- a/.github/resources/local.repos
+++ b/.github/resources/local.repos
@@ -9,8 +9,3 @@ repositories:
 
   ros2/rosidl_typesupport_fastrtps/COLCON_IGNORE: *empty_repo
   ros2/rmw_fastrtps/COLCON_IGNORE: *empty_repo
-
-  ament/ament_lint:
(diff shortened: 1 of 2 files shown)


clalancette (approved):
Windows still seems to be cranky.  Nevertheless, I'm still going to approve since this seems like an improvement to me.  Should we do something similar on master branch?

jacobperron:
> Should we do something similar on master branch?

Probably? I opened this PR in response to https://github.com/ros2/rmw_cyclonedds/pull/369#issuecomment-1030142851

jacobperron:
We decided to remove the CI workflow altogether in https://github.com/ros2/rmw_cyclonedds/pull/417

PR #373, opened 2022-02-17
Call dissociate_reader in rmw_destroy_subscription

eboasson:
A call to dissociate_writer, like it did before, is a no-op.  The result
is that the ros_discovery_info topic sample keeps growing when one
creates and destroy subscriptions in the application.


omertal88 (approved):

eboasson:
Thanks @omertal88 for reviewing it. It seems silly to run the full CI for this change, so below are the badges for a run with just `test_rmw_implementation` and `test_rclcpp` as a sanity check:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16216/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10842/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16572/)

PR #375, opened 2022-02-28
Fix a warning by making a pointer nullptr.

clalancette:
When building on Ubuntu 22.04 in Release mode, gcc complains that:

‘stact’ may be used uninitialized in this function [-Wmaybe-uninitialized] 2295

I think this is probably a false positive; in any situations
where stact was not set, we would have gotten out before this.
Regardless, just set stact to nullptr, which quiets the warning
and should have no deleterious effects.

This should fix warnings from the nightlies like the last one shown in https://ci.ros2.org/view/nightly/job/nightly_linux_release/2205/gcc/new/


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16262/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10883/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16622/)

fujitatomoya (approved):

clalancette:
Linux aarch64 warnings are from other problems that need to be fixed.  This improves the situation, so going ahead and merging this one.

PR #393, opened 2022-05-11
Add cstring include.

clalancette:
This is needed for memset() and memcpy().

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

This should fix #392

PR #398, opened 2022-06-27
[Fix] Add cstring header for memset

homalozoa:
Cherry-pick #393


clalancette:
Did any CI for this get run?  I know Humble is still pretty close to Rolling, but we generally still run CI.

ivanpauno:
I trusted the `Hpr__rmw_cyclonedds__ubuntu_jammy_amd64` job, which passed.
I usually think that's fine for really small backports like this one, but let me know if you think I should also run full CI for this kind of changes.

PR #402, opened 2022-08-15
Add rmw_get_gid_for_client impl

ihasdapie:
See: https://github.com/ros2/rmw_fastrtps/pull/631

Signed-off-by: Brian Chen <brian.chen@openrobotics.org>


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2494,6 +2494,24 @@ extern "C" rmw_ret_t rmw_get_gid_for_publisher(const rmw_publisher_t * publisher
...
+    client,
+    client->implementation_identifier,
+    eclipse_cyclonedds_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RMW_CHECK_ARGUMENT_FOR_NULL(gid, RMW_RET_INVALID_ARGUMENT);
+
+  const CddsClient * cli = static_cast<const CddsClient *>(client->data);
+  gid->implementation_identifier = eclipse_cyclonedds_identifier;
+  memset(gid->data, 0, sizeof(gid->data));
+  assert(sizeof(cli->client.id.data) <= sizeof(gid->data));
```
ivanpauno, line 2510: can you use an static_assert instead?

ivanpauno (approved):
LGTM!

I only have one minor comment

rmw_cyclonedds_cpp/src/rmw_node.cpp
jacobperron: Yes, I've updated this assert and the similar one for the publisher gid: [60c9255](https://github.com/ros2/rmw_cyclonedds/pull/402/commits/60c925587fa9d9ed151b246833ee54711ac8ae42)

ivanpauno (approved):

Issue #409, opened 2022-08-30
Include topic names in error messages


tfoote:
#### Feature description
We added better error messages when topics are invalid in #405 but the topics aren't included in the error message. It would be great to include them. 

#### Implementation considerations

The topic name can be passed into this function to be included inthe error strings. 

https://github.com/ros2/rmw_cyclonedds/pull/405/files#diff-91f31a99b38624440900850fd2df4710f2ef46c140999f3e3b8b89393e48e202R1790

And the name passed in in the 4 places it's invoked:

https://github.com/ros2/rmw_cyclonedds/pull/405/files#diff-91f31a99b38624440900850fd2df4710f2ef46c140999f3e3b8b89393e48e202R2296

https://github.com/ros2/rmw_cyclonedds/pull/405/files#diff-91f31a99b38624440900850fd2df4710f2ef46c140999f3e3b8b89393e48e202R2806

https://github.com/ros2/rmw_cyclonedds/pull/405/files#diff-91f31a99b38624440900850fd2df4710f2ef46c140999f3e3b8b89393e48e202R4638

https://github.com/ros2/rmw_cyclonedds/pull/405/files#diff-91f31a99b38624440900850fd2df4710f2ef46c140999f3e3b8b89393e48e202R4647

tfoote:
@Voldivh FYI

Voldivh:
Hey @tfoote, take a look at the PR #410 and let me know.

clalancette:
We completed this in #410, so closing this issue.

PR #410, opened 2022-08-31
Adds topic name to error msg when create_topic fails

Voldivh:
This PR addresses the issue #409. As the title states, it adds the topic name to the error message when create_topic fails.

Signed-off-by: Voldivh <eloyabmfcv@gmail.com>

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1787,23 +1787,24 @@ static dds_entity_t create_topic(dds_entity_t pp, const char * name, struct ddsi
   return tp;
 }
 
-void set_error_message_from_create_topic(dds_entity_t topic)
+void set_error_message_from_create_topic(dds_entity_t topic, std::string topic_name)
 {
   assert(topic < 0);
   if (DDS_RETCODE_BAD_PARAMETER == topic) {
-    RMW_SET_ERROR_MSG(
-      "failed to create topic because the function was given"
-      " invalid parameters");
+    std::string error_msg = "failed to create topic [" + topic_name +
+      "] because the function was given invalid parameters";
+    RMW_SET_ERROR_MSG(error_msg.c_str());
   } else if (DDS_RETCODE_INCONSISTENT_POLICY == topic) {
-    RMW_SET_ERROR_MSG(
-      "failed to create topic because it's already in use"
-      " in this context with incompatible QoS settings");
+    std::string error_msg = "failed to create topic [" + topic_name +
+      "] because it's already in use in this context with incompatible QoS settings";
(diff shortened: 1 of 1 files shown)


tfoote (approved):
I was thinking it might make sense to only have one error_msg and just set it within the scope of the if statement. Then call RMW_SET_ERROR_MSG at the end. But that will hide from the macro the line number from which each instance of the error was called. 

lgtm

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1787,23 +1787,24 @@ static dds_entity_t create_topic(dds_entity_t pp, const char * name, struct ddsi
   return tp;
 }
 
-void set_error_message_from_create_topic(dds_entity_t topic)
+void set_error_message_from_create_topic(dds_entity_t topic, std::string topic_name)
```
francocipollone, line 1790: nit: What about `const std::string& topic_name`?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1787,23 +1787,24 @@ static dds_entity_t create_topic(dds_entity_t pp, const char * name, struct ddsi
...
 {
   assert(topic < 0);
   if (DDS_RETCODE_BAD_PARAMETER == topic) {
-    RMW_SET_ERROR_MSG(
-      "failed to create topic because the function was given"
-      " invalid parameters");
+    std::string error_msg = "failed to create topic [" + topic_name +
```
francocipollone, line 1794: nit: Could it be const?

francocipollone: Here and there

francocipollone (commented):
> But that will hide from the macro the line number from which each instance of the error was called.

+1

rmw_cyclonedds_cpp/src/rmw_node.cpp
Voldivh: Done.

rmw_cyclonedds_cpp/src/rmw_node.cpp
Voldivh: Nice catch, thanks!

tfoote:
Gist: https://github.com/ros2/ros2/blob/rolling/ros2.repos

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17301/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11840/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17788/)

tfoote:
@mergify backport humble galactic foxy

PR #428, opened 2022-11-07
[rolling] Update maintainers - 2022-11-07

audrow:
This PR updates the maintainers and code owners. The new maintainers are:
* Ivan Paunovic (@ivanpauno)

This PR was made with https://github.com/audrow/update-ros2-repos.


eboasson (approved):
I'm fine with the change, but I do notice that I could be viewed as inconsistent with `package.xml`.

clalancette (changes requested):
This one didn't remove Mich from the maintainers in the package.xml.

@eboasson You are right about it being inconsistent.  Would you be OK if we added your name into the CODEOWNERS file here?

eboasson:
> This one didn't remove Mich from the maintainers in the package.xml.
> 
> @eboasson You are right about it being inconsistent. Would you be OK if we added your name into the CODEOWNERS file here?

 I'm really happy I attended ROSCon: now I know again why I do it 😁 so yes, that's certainly ok!

clalancette (approved):

PR #429, opened 2022-12-14
[rmw_cyclonedds] Improve handling of dynamic discovery

gbiggs:
This PR adds to `rmw_cyclonedds` functionality necessary to support the [improved handling of dynamic discovery](https://github.com/ros2/ros2/issues/1359).

It adds handling of the new discovery parameters passed down from `rcl`.

Current status: The desired functionality is only partially complete. Some of the entries in the [goal matrix](https://discourse.ros.org/t/proposed-changes-to-how-ros-performs-discovery-of-nodes/27640) are not working correctly.


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1030,39 +1035,65 @@ static rmw_ret_t discovery_thread_stop(rmw_dds_common::Context & common_context)
...
-      :
-      "";
+    dom.discovery_params = *discovery_params;
+
+    std::string config;
+
+    switch (discovery_params->automatic_discovery_range) {
+      case RMW_AUTOMATIC_DISCOVERY_RANGE_OFF:
+        /* Automatic discovery off: disable multicast for participant discovery. */
+        config += "<CycloneDDS><Domain><General><AllowMulticast>asm,ssm</AllowMulticast>"
```
mxgrey, line 1065: The [Cyclone docs](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/manual/options.md#cycloneddsdomaingeneralallowmulticast) have this to say about default settings for `AllowMulticast`:

> "default" maps on spdp if the network is a WiFi network, on true if it is a wired network

I assume the default of only SPDP for WiFi was chosen to minimize multicast traffic over wifi without totally killing the normal discovery behavior. That would make sense since multicast over wifi has some painful issues with slowing down transmission rates and even wasting device battery.

WiFi considerations are very important in modern robotics, so I wonder if we should also test whether the device is wired or wireless and maybe just use `false` here if it's a wireless device?

For that matter, do we actually need the `asm` and `ssm` at all if they're apparently safe to turn off for WiFi devices? The docs say that including `asm` enables receiving (but not transmitting) SPDP messages via multicast. Do we need or even want that for `OFF`? Will disabling `asm` have a negative impact on the static peer discovery?

As for `ssm`, the docs don't indicate the purpose/function of `ssm`-based traffic within CycloneDDS so it's unclear to me if that's safe for us to disable or if we really need it. But again, if the default is to turn it off when using WiFi then that makes me doubt that it's actually needed.

eboasson: > I assume the default of only SPDP for WiFi was chosen to minimize multicast traffic over wifi without totally killing the normal discovery behavior.

Indeed.

> For that matter, do we actually need the asm and ssm at all if they're apparently safe to turn off for WiFi devices? The docs say that including asm enables receiving (but not transmitting) SPDP messages via multicast. Do we need or even want that for OFF? Will disabling asm have a negative impact on the static peer discovery?

If you don't want multicast (including participant discovery), then I don't see much benefit for the ROS 2 default behaviour to anything other than `false`. Specifying `asm` with fully static discovery could still make sense in combination with deliberately mapping some traffic on multicast addresses, but I don't see that as an important scenario here.

> As for ssm, the docs don't indicate the purpose/function of ssm-based traffic within CycloneDDS so it's unclear to me if that's safe for us to disable or if we really need it. But again, if the default is to turn it off when using WiFi then that makes me doubt that it's actually needed.

You won't ever notice anything about SSM unless you explicitly use multicast addresses in the SSM block. I don't see a need to worry about it here.

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-02-16/29927/1

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1030,39 +1036,121 @@ static rmw_ret_t discovery_thread_stop(rmw_dds_common::Context & common_context)
...
+           configuration. */
+        break;
+      default:
+        /* This situation would happen if a new option is introduced for the
+           ROS_AUTOMATIC_DISCOVERY_RANGE but rmw_cyclonedds_cpp is not updated
+           to handle it. */
+        /* TODO(MXG): Displaying an integer is not very informative for users.
+           It would be better if rmw knows how to translate the integer values
+           into their associated option strings. However the knowledge of the
+           string values is currently located in rcl instead of rmw. */
```
wjwwood, line 1086: Should we address this TODO?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1030,39 +1036,121 @@ static rmw_ret_t discovery_thread_stop(rmw_dds_common::Context & common_context)
...
+        RCUTILS_LOG_WARN_NAMED(
+          "rmw_cyclonedds_cpp",
+          "check_create_domain: unsupported value for automatic_discovery_range: %i",
+          discovery_options->automatic_discovery_range);
+        /* Intentionally fall through to the LOCALHOST / DEFAULT case */
+        [[clang::fallthrough]];
+        // [[fallthrough]]; /* Uncomment this back when migrating to C++17 */
```
wjwwood, line 1093: We're on C++17, should this addressed?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1078,6 +1166,8 @@ static bool check_create_domain(dds_domainid_t did, rmw_localhost_only_t localho
       return false;
     }
 
+    RCUTILS_LOG_INFO_NAMED("rmw_cyclonedds_cpp", "Config XML is %s", config.c_str());
```
wjwwood, line 1169: This doesn't seem like a log info we should leave in, right?

rmw_cyclonedds_cpp/src/rmw_node.cpp
sloretz: Using C++17 in 8324e11

rmw_cyclonedds_cpp/src/rmw_node.cpp
sloretz: Changed to DEBUG in 8324e11

rmw_cyclonedds_cpp/src/rmw_node.cpp
sloretz: If I understand correctly, addressing this TODO would mean making an `enum -> string` conversion function at the RMW layer; however, if `rmw_cyclonedds_cpp` doesn't know the meaning of an enum value then either it's a new option, or it's an invalid value.

If it's a new option, then maybe the `rmw` library linked knows about it and can display something sensible (case 1). However, maybe the `rmw` library linked is also out of date and it doesn't know about the option so the string it displays is "unknown" (case 2). If it's an invalid value than the best a conversion function could do was say "unknown" (case 3). I think the integer value would be more valuable than "unknown" in cases 2 and 3, an inconvenient but the same amount of information in case 1.

I'd vote to remove the `TODO` instead of addressing it.

sloretz: Removed TODO in 5ed2f3b

sloretz:
@eboasson Any ideas why with automatic discovery range set to localhost, all these errors would show up on windows? The error `WSAEADDRNOTAVAIL` shows up while cyclone is trying to send a UDP packet.

```
[test_subscriber-9] 1680304639.633310 [1]        tev: slortez: WSAEADDRNOTAVAIL (or fall through from WSAEFAULT)slortez: WSAEAFNOSUPPORT (or fall through from WSAEADDRNOTAVAIL)ddsi_udp_conn_write to udp/127.0.0.1:7660 failed with retcode -3
```

The number seems to vary, but the later tests have 119 ports tried, which seems to match the maximum participant index. It's almost like on Windows there's an issue sending the discovery packets to other potential participants.

The job was built with the following commit with extra log messages in cyclone
https://ci.ros2.org/job/ci_windows/19050/testReport/junit/(root)/projectroot/test_publisher_subscriber__rclpy__rmw_cyclonedds_cpp/


https://github.com/sloretz/cyclonedds/commit/c1b549b4d0f34d5f0927b44ea069dbe96904f8f8

wjwwood (approved):

PR #431, opened 2022-12-20
Implement inconsistent topic.

clalancette:
Note that this is not hooked up inside of CycloneDDS, so this will never fire currently.

Part of https://github.com/ros2/ros2/issues/1361

Depends on https://github.com/ros2/rmw/pull/339


ivanpauno (approved):
LGTM!
It would be nice if we merge this after we have a `rmw_implementation` PR with tests and CI passing.

ivanpauno:
> LGTM!
> It would be nice if we merge this after we have a rmw_implementation PR with tests and CI passing.

or maybe the tests could live in `rcl`

eboasson (approved):
The change looks good to me, but there is the issue that I am not so sure Cyclone will raise "inconsistent topic" based on reader/writer discovery. It should (and will) do it based on *topic* discovery.

The reason I am not sure is because I do think there's a good argument that a reader/writer pair that fails to match on type name (possibly also assignability, but there other QoS settings also come into play) implies that an "inconsistent topic" notification would have been raised if topic discovery had been enabled, and that it therefore makes sense to raise it at that point if topic discovery is disabled.

I'm approving this as a Cyclone's not raising "inconsistent topic" today even when topic discovery is enabled is simply a bug, and because I believe this code will work fine when that bug is fixed.

fujitatomoya (approved):

PR #436, opened 2023-02-16
Detect incompatible types from incompatible QoS callbacks

allenh1:
@clalancette this PR attempts to detect incompatible types from the incompatible QoS callbacks that are already in place.

It's a bit of a hack, and I am not sure how to check it locally (as it seems to depend on some other changes I lack locally), but I think this approach will work.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -515,7 +515,7 @@ MAKE_DDS_EVENT_CALLBACK_FN(sample_lost, SAMPLE_LOST)
 MAKE_DDS_EVENT_CALLBACK_FN(liveliness_changed, LIVELINESS_CHANGED)
 
 /**
- * Because the inconsistent topic is not handled correctly in CycloneDDS,
+ * Because the inconsistent topic is not raised by CycloneDDS when a reader/writer fail to match because of differing type definitions
  * this callback is signalled via the incompatible qos handlers
  */
 // MAKE_DDS_EVENT_CALLBACK_FN(inconsistent_topic, INCONSISTENT_TOPIC)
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -511,11 +511,56 @@ static void dds_listener_callback(dds_entity_t entity, void * arg)
...
 MAKE_DDS_EVENT_CALLBACK_FN(liveliness_lost, LIVELINESS_LOST)
 MAKE_DDS_EVENT_CALLBACK_FN(offered_deadline_missed, OFFERED_DEADLINE_MISSED)
-MAKE_DDS_EVENT_CALLBACK_FN(requested_incompatible_qos, REQUESTED_INCOMPATIBLE_QOS)
 MAKE_DDS_EVENT_CALLBACK_FN(sample_lost, SAMPLE_LOST)
-MAKE_DDS_EVENT_CALLBACK_FN(offered_incompatible_qos, OFFERED_INCOMPATIBLE_QOS)
 MAKE_DDS_EVENT_CALLBACK_FN(liveliness_changed, LIVELINESS_CHANGED)
-MAKE_DDS_EVENT_CALLBACK_FN(inconsistent_topic, INCONSISTENT_TOPIC)
+
+/**
+ * Because the inconsistent topic is not handled correctly in CycloneDDS,
```
eboasson, line 518: I disagree with the definitive character of this statement *in this context*: I don't believe that it is sufficiently clear what "correct" handling of the "inconsistent topic" notification is to make this statement in the context of detecting the existence of multiple definitions for a topic T in the system with different types.

What I do accept without reservation is that it is a bug that Cyclone fails to raise an "inconsistent topic" notification when topic discovery is enabled while there definitions of T with differing QoS settings.

But this is too strong for the context.

```suggestion
 * Because the inconsistent topic is not raised by CycloneDDS when a reader/writer fail to match because of differing type definitions
```

eboasson (commented):
Forgot to mention: other than this one comment on a comment, it looks good to me.

rmw_cyclonedds_cpp/src/rmw_node.cpp
allenh1: @eboasson Thanks for the clarification. For what it's worth, I did not mean to knock CycloneDDS here, I was just being lazy in my comment. "Not handled correctly" was a poor way to phrase this.

asymingt:
A large PR adding key support to `rmw_cyclonedds` is about to land: https://github.com/ros2/rmw_cyclonedds/pull/550. Advise holding off until merge. Assigning @eboasson to help shepherd this PR after it merges.

PR #439, opened 2023-03-03
Revisions for changes to peer discovery changes

mxgrey:
This updates #429 to match the revised requirements matrix that was recommended [in this post](https://discourse.ros.org/t/proposed-changes-to-how-ros-performs-discovery-of-nodes/27640/25) on the discourse thread.


wjwwood:
After discussing it we decided to merge all of these into the original pr's and go from there.

PR #441, opened 2023-03-09
Update to new rmw_discovery_options API

mxgrey:
This updates the discovery options PR to match the upstream API changes that have been iterating in https://github.com/ros2/rmw/pull/349


wjwwood:
Merging with https://github.com/ros2/rmw_cyclonedds/pull/429

PR #442, opened 2023-03-21
Revert "Implement inconsistent topic."

Crola1702:
Reverts ros2/rmw_cyclonedds#431

As mentioned in https://github.com/ros2/rmw_cyclonedds/pull/431#issuecomment-1478510006. It seems this PR somehow made CycloneDDS fail on Rolling branch.

As CycloneDDS is at Tier 1 support level. I want to the PR until further investigation is done, and the issue is solved.

FYI: @clalancette @claraberendsen @Blast545

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -515,7 +515,6 @@ MAKE_DDS_EVENT_CALLBACK_FN(requested_incompatible_qos, REQUESTED_INCOMPATIBLE_QO
 MAKE_DDS_EVENT_CALLBACK_FN(sample_lost, SAMPLE_LOST)
 MAKE_DDS_EVENT_CALLBACK_FN(offered_incompatible_qos, OFFERED_INCOMPATIBLE_QOS)
 MAKE_DDS_EVENT_CALLBACK_FN(liveliness_changed, LIVELINESS_CHANGED)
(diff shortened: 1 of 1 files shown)


eboasson:
What's happening is that this PR
* adds an `RMW_EVENT_SUBSCRIPTION_INCOMPATIBLE_TYPE` event on the subscriber
* this gets mapped to `DDS_INCONSISTENT_TOPIC_STATUS`
* which gets passed to `dds_set_status_mask` in `gather_event_entities` (in `rmw_cyclonedds_cpp`) because `is_event_supported` (also `rmw_cyclonedds_cpp`) says "sure"
* and then `dds_set_status_mask` refuses the mask because this status is not supported on the reader and leaves the status mask whatever it is (everything enabled, probably)
* that error is not picked up (one of those "cannot happens"), and so everything proceeds happily
* and then `dds_wait` returns immediately because some status is set (subscription matched, or so)
* and then `rclcpp::wait_for_message` tries to take data, finding nothing is available and returns `false`
* resulting in a test failure in `test_wait_for_message.cpp:42`

https://github.com/ros2/rmw_cyclonedds/pull/436 is relevant here, in that it follows the way Cyclone reports type incompatibilities. https://github.com/eclipse-cyclonedds/cyclonedds/issues/1523#issuecomment-1359605212 also goes into some detail.

I may well have reviewed something and declared it looked good because I overlooked this detail ... If so I apologise.

clalancette:
Given that we fixed this in #444, I'm going to close this out.  Feel free to reopen if you think that was in error.

PR #444, opened 2023-03-22
Disable inconsistent topic events.

clalancette:
These can't successfully be set on the entities right now, so just skip them.

Also make sure to check for errors when setting the status mask, which should help surface this kind of problem in the future.

I'll note that we are kind of lying here; we are telling the upper rcl/rclpy/rclpp layers that we support the INCOMPATIBLE_TYPE events, but then we are failing to register for it down below.  But I think that is a reasonable compromise here.

@eboasson I'm especially interested on your take for this fix.

This replaces https://github.com/ros2/rmw_cyclonedds/pull/442 .  @Crola1702 FYI

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -4153,12 +4153,22 @@ static rmw_ret_t gather_event_entities(
       if (status_mask_map.find(dds_entity) == status_mask_map.end()) {
         status_mask_map[dds_entity] = 0;
       }
-      status_mask_map[dds_entity] |= get_status_kind_from_rmw(current_event->event_type);
+
+      uint32_t status_kind = get_status_kind_from_rmw(current_event->event_type);
+      // TODO(clalancette): This should be reenabled when Cyclone supports reporting inconsistent
+      // topic as an event
+      if (status_kind != DDS_INCONSISTENT_TOPIC_STATUS) {
+        status_mask_map[dds_entity] |= get_status_kind_from_rmw(current_event->event_type);
+      }
     }
   }
   for (auto & pair : status_mask_map) {
     // set the status condition's mask with the supported type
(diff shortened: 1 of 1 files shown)


Crola1702:
Linux CI (only CycloneDDS): [Build Status](https://ci.ros2.org/job/ci_linux/18286/)

eboasson (approved):
I'd say this is a good band-aid 🙂

What I think ought to be the real fix for Cyclone at this stage [^1] is to map the RMW `RMW_EVENT_PUBLISHER_INCOMPATIBLE_TYPE` to `DDS_OFFERED_INCOMPATIBLE_QOS_STATUS` and then map a case of a reported incompatible QoS of `DDS_TYPE_CONSISTENCY_ENFORCEMENT_QOS_POLICY` to `RMW_EVENT_PUBLISHER_INCOMPATIBLE_TYPE`. (And obviously similarly for subscribers.)

I haven't really given any thought to how easily that mapping is done. It might be somewhat tricky because the status could be overwritten by mismatch events in between the signalling of the "offered incompatible QoS" and actually reading from the event. I am certain that it is possible, because listeners are invoked synchronously in Cyclone. But in my experience there is no perfect correlation between "possible" and "easy" ...

[^1]: At this point we're still inclined towards restricting the "incompatible topic" status to topics, and if we stick with that, mapping to `DDS_INCONSISTENT_TOPIC_STATUS` is never going to work ...

clalancette:
OK, thanks for the feedback!  Here is full CI so we can get this in:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18291/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12817/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18959/)

clalancette:
The one test failures on Windows is a well-known flake.  Going ahead and merging this, thanks for the reviews!

PR #445, opened 2023-03-28
Support DDS XTypes creation

Splinter1984:
##### Description:
This changes allow us to register `ros_types` as `DDSXtypes`. 
In example this allow to use cyclonedds cli tool to access to middleware of ros2.

Example (subscribe to ros2 topic):
```bash
~$ cyclonedds subscribe rt/chatter
   0:00:01 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Entities discovered: 1

 Subscribing, CTRL-C to quit
String_(data='Hello World: 1')
String_(data='Hello World: 2')
String_(data='Hello World: 3')
```
Example (display type information):
```bash
~$ cyclonedds typeof rt/chatter
   0:00:01 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Entities discovered: 1

As defined in participant(s) 0110283f-9e45-f5f7-b391-ffe7000001c1
module std_msgs {                                                                                                                                         
    module msg {                                                                                                                                          
        module dds_ {                                                                                                                                     
            @final                                                                                                                                        
            struct String_ {                                                                                                                              
                string data;                                                                                                                              
            };                                                                                                                                            
        };                                                                                                                                                
    };                                                                                                                                                    
};
```
##### Depends on:
- [eclipse-cyclonedds/cyclonedds#1694](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1694)

### DIFF
--- a/rmw_cyclonedds_cpp/src/serdata.cpp
+++ b/rmw_cyclonedds_cpp/src/serdata.cpp
@@ -24,7 +24,7 @@
 #include "Serialization.hpp"
 #include "TypeSupport2.hpp"
 #include "bytewise.hpp"
-#include "dds/ddsi/q_radmin.h"
+#include "dds/ddsi/ddsi_radmin.h"
 #include "rmw/error_handling.h"
 #include "MessageTypeSupport.hpp"
 #include "ServiceTypeSupport.hpp"
@@ -185,7 +185,7 @@ static void serdata_rmw_free(struct ddsi_serdata * dcmn)
 static struct ddsi_serdata * serdata_rmw_from_ser(
   const struct ddsi_sertype * type,
   enum ddsi_serdata_kind kind,
-  const struct nn_rdata * fragchain, size_t size)
+  const struct ddsi_rdata * fragchain, size_t size)
 {
   try {
     auto d = std::make_unique<serdata_rmw>(type, kind);
@@ -199,7 +199,7 @@ static struct ddsi_serdata * serdata_rmw_from_ser(
       if (fragchain->maxp1 > off) {
         /* only copy if this fragment adds data */
         const unsigned char * payload =
-          NN_RMSG_PAYLOADOFF(fragchain->rmsg, NN_RDATA_PAYLOAD_OFF(fragchain));
+          DDSI_RMSG_PAYLOADOFF(fragchain->rmsg, DDSI_RDATA_PAYLOAD_OFF(fragchain));
         auto src = payload + off - fragchain->min;
         auto n_bytes = fragchain->maxp1 - off;
         memcpy(cursor, src, n_bytes);
--- a/rmw_cyclonedds_cpp/src/serdata.cpp
+++ b/rmw_cyclonedds_cpp/src/serdata.cpp
@@ -855,6 +855,31 @@ static void dynamic_type_add_member(
   }
 }
 
+static void dynamic_type_register(struct sertype_rmw * st, dds_dynamic_type_t dt, dds_entity_t dds_ppant)
+{
+  dds_typeinfo_t *type_info;
+  auto rc = dds_dynamic_type_register(&dt, &type_info);
+  if (rc != DDS_RETCODE_OK)
+    RMW_SET_ERROR_MSG("dds_dynamic_type_register failed to register type");
+  
+  dds_topic_descriptor_t *desc;
+  rc = dds_create_topic_descriptor(
+    DDS_FIND_SCOPE_GLOBAL, dds_ppant, type_info, 0, &desc);
+  if (rc != DDS_RETCODE_OK)
+    RMW_SET_ERROR_MSG("dds_create_topic_descriptor failed to create descriptor");
+
+  if (st)
+  {
+    st->type_information.data = static_cast<const unsigned char *>(ddsrt_memdup(desc->type_information.data, desc->type_information.sz));
+    st->type_information.sz = desc->type_information.sz;
+    st->type_mapping.data = static_cast<const unsigned char *>(ddsrt_memdup(desc->type_mapping.data, desc->type_mapping.sz));
+    st->type_mapping.sz = desc->type_mapping.sz;
+  }
+
+  ddsi_typeinfo_fini(type_info);
+  dds_delete_topic_descriptor(desc);
+}
+
 template<typename MembersType>
 static bool construct_dds_dynamic_type(
   dds_dynamic_type_t * dstruct, dds_entity_t dds_ppant, const MembersType * members)
@@ -927,7 +952,7 @@ static bool construct_dds_dynamic_type(
           }
           if (!member->is_array_)
           {
-            ret = dds_dynamic_type_add_member(dstruct, DDS_DYNAMIC_MEMBER(dds_dynamic_type_ref(&ddt), member->name_));
+            ret = dds_dynamic_type_add_member(dstruct, DDS_DYNAMIC_MEMBER(ddt, member->name_));
             if (ret != DDS_RETCODE_OK)
               RCUTILS_LOG_ERROR("failed to add type member: %d", DDS_DYNAMIC_STRING8);
           } else {
@@ -953,26 +978,22 @@ static bool construct_dds_dynamic_type(
             });
 
           dds_dynamic_type_set_extensibility(&ddt, DDS_DYNAMIC_TYPE_EXT_FINAL);
+
           if (!construct_dds_dynamic_type(&ddt, dds_ppant, sub_members))
           {
             RMW_SET_ERROR_MSG("construct_dds_dynamic_type error");
             return false;
           }
-          dds_typeinfo_t *type_info;
-          auto rc = dds_dynamic_type_register(&ddt, &type_info);
-          if (rc != DDS_RETCODE_OK)
-            RCUTILS_LOG_ERROR("rmw_cyclonedds_cpp", "dds_dynamic_type_register fail: %s", dds_strretcode(-rc));
 
           if (!member->is_array_)
           {
-            ret = dds_dynamic_type_add_member(dstruct, DDS_DYNAMIC_MEMBER(dds_dynamic_type_ref(&ddt), member->name_));
+            ret = dds_dynamic_type_add_member(dstruct, DDS_DYNAMIC_MEMBER(ddt, member->name_));
             if (ret != DDS_RETCODE_OK)
               RCUTILS_LOG_ERROR("failed to add dynamic type member");
           } else {
             dynamic_type_add_array(dstruct, dds_ppant, member, ddt);
           }
           
-          ddsi_typeinfo_fini(type_info);
           break;
         }
       default:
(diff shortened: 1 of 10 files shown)


rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -685,11 +749,351 @@ static std::string get_type_name(const char * type_support_identifier, void * ty
   }
 }
 
+
+extern "C" dds_dynamic_type_descriptor_t get_dynamic_type_descriptor_prim(
+  dds_dynamic_type_kind_t kind, const char *name, uint32_t num_bounds, const uint32_t *bounds, dds_dynamic_type_kind_t type)
+{
+  return{kind,name,{},{},num_bounds,bounds,DDS_DYNAMIC_TYPE_SPEC_PRIM(type),{},};
```
eboasson, line 756: I wonder if there is no better way to do this. The C example/test code uses designated initializers to set the fields that are required (and have the rest auto-initialized to 0), if that's still not supported in C++ perhaps doing:
```C++
dds_dynamic_type_descriptor t desc{};
desc.kind = kind;
...
```
works out better. I'm not quite sure, but writing it out as a struct initializer like this looks fragile.

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -685,11 +749,351 @@ static std::string get_type_name(const char * type_support_identifier, void * ty
...
+  dds_dynamic_type_t * dstruct, dds_entity_t dds_ppant, const MemberType * member, const dds_dynamic_type_kind_t type)
+{
+  dds_dynamic_type_t ddt;
+
+  if (member->array_size_) 
+  { 
+    if (!member->is_upper_bound_)
+    {
+      ddt = dds_dynamic_type_create(dds_ppant, 
+        get_dynamic_type_descriptor_prim(DDS_DYNAMIC_ARRAY, member->name_, 1, (uint32_t*)(&member->array_size_), type));
```
eboasson, line 789: I believe the actual type of `array_size_` is `size_t`, so casting it to a pointer to `uint32_t` will break things quite badly on big-endian 64-bit machines.

I've checked and `dds_dynamic_type_create` in turn calls `ddsi_dynamic_type_create_array`, which copies the data. Given that it is a single-element array, changing it to:
```
uint32_t array_size = static_cast<uint32_t>(member->array_size_);
ddt = dds_dynamic_type_create(dds_ppant, 
        get_dynamic_type_descriptor_prim(DDS_DYNAMIC_ARRAY, member->name_, 1, &array_size_, type));
```
should solve that.

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -685,11 +749,351 @@ static std::string get_type_name(const char * type_support_identifier, void * ty
...
+    ddt = dds_dynamic_type_create(dds_ppant,
+       get_dynamic_type_descriptor_prim(DDS_DYNAMIC_SEQUENCE, member->name_, 0, nullptr, type)); 
+  }
+
+  dds_return_t ret = dds_dynamic_type_add_member(dstruct, get_dynamic_member_descriptor(ddt, member->name_));
+  if (ret != DDS_RETCODE_OK)
+      RCUTILS_LOG_ERROR("failed to add type member sequence/array with prim type");
+}
+
+template<typename MemberType>
```
eboasson, line 804: I must be overlooking something obvious, because it looks like this is so close to the function above that it might as well not exist!

(But then surely the compiler would reject it. So I must be the one who is wrong here.)

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -685,11 +749,351 @@ static std::string get_type_name(const char * type_support_identifier, void * ty
...
+  if (ret != DDS_RETCODE_OK)
+    RCUTILS_LOG_ERROR("failed to add type member sequence/array with dynamic type");
+}
+
+template<typename MemberType>
+static void dynamic_type_add_member(
+    dds_dynamic_type_t * dstruct, dds_entity_t dds_ppant, const MemberType * member, const dds_dynamic_type_kind_t type)
+{
+  dds_return_t ret;
+  if (type == rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING ||
```
eboasson, line 835: If I am not mistaken, these two cases are handled by `construct_dds_dynamic_type`, but then I would prefer an `assert` or `abort` instead of returning without doing anything.

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -685,11 +749,351 @@ static std::string get_type_name(const char * type_support_identifier, void * ty
...
+    dynamic_type_add_array(dstruct, dds_ppant, member, type);
+  }
+}
+
+static void dynamic_type_register(struct sertype_rmw * st, dds_dynamic_type_t dt, dds_entity_t dds_ppant)
+{
+  dds_typeinfo_t *type_info;
+  auto rc = dds_dynamic_type_register(&dt, &type_info);
+  if (rc != DDS_RETCODE_OK)
+    RMW_SET_ERROR_MSG("dds_dynamic_type_register failed to register type");
```
eboasson, line 854: It would be better to propagate the error out here, I think. I suspect continuing after an error will crash ... (This applies in a few more places.)

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -685,11 +749,351 @@ static std::string get_type_name(const char * type_support_identifier, void * ty
...
+  
+  dds_topic_descriptor_t *desc;
+  rc = dds_create_topic_descriptor(
+    DDS_FIND_SCOPE_GLOBAL, dds_ppant, type_info, 0, &desc);
+  if (rc != DDS_RETCODE_OK)
+    RMW_SET_ERROR_MSG("dds_create_topic_descriptor failed to create descriptor");
+
+  if (st)
```
eboasson, line 862: Do we ever get here when `st == nullptr`? It seems to me there is not much reason to.

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -514,6 +516,9 @@ static void sertype_rmw_free(struct ddsi_sertype * tpcmn)
     tp->type_support.type_support_ = NULL;
   }
 
+  free((void*)tp->type_information.data);
```
eboasson, line 519: These are allocated by `ddsrt_memdup`, right? Then they should be freed with `ddsrt_free` instead of free. (It won't matter on Linux, but on Windows it gets tricky with how they deal with DLLs.)

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -685,11 +749,351 @@ static std::string get_type_name(const char * type_support_identifier, void * ty
...
+      case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_UINT64:
+        dynamic_type_add_member(dstruct, dds_ppant, member, DDS_DYNAMIC_UINT64);
+        break;
+      case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING:
+        {
+          dds_dynamic_type_t ddt;
+          if (member->is_upper_bound_)
+          {
+            ddt = dds_dynamic_type_create(dds_ppant,
+              get_dynamic_type_descriptor(DDS_DYNAMIC_STRING8, nullptr, 1, (uint32_t*)(&member->array_size_), {}));
```
eboasson, line 934: Here the same thing I mentioned above about `size_t` vs `uint32_t` applies as well.

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -685,11 +749,351 @@ static std::string get_type_name(const char * type_support_identifier, void * ty
...
+            return false;
+          }
+
+          if (!member->is_array_)
+          {
+            ret = dds_dynamic_type_add_member(dstruct, get_dynamic_member_descriptor(ddt, member->name_));
+            if (ret != DDS_RETCODE_OK)
+              RCUTILS_LOG_ERROR("failed to add dynamic type member");
+          } else {
+            dynamic_type_add_array(dstruct, dds_ppant, member, ddt);
```
eboasson, line 977: I'm surprised it checks for `is_array_` but apparently doesn't need to deal with `is_upper_bound_` or `array_size_`. Is that really true?

The `ROS_TYPE_STRING` case is again subtly different in this regard ...

I should check the `rosidl` definitions, I know, but for now, perhaps you can just double-check? ☺️

eboasson (commented):
Thank you so much for this @Splinter1984! It took forever for me to sit down and give it a proper review, but now I have. This is a lovely addition to the RMW layer and while I do have some comments, they really are about small things. I do admit that I haven't spend enough attention on all the error paths yet. Perhaps you can have a look at the comments and then we can include that aspect in a second round?

I am also quite sure that the code layout check will fail. Almost every PR seems to end up with a final round through `uncrustify` to meet the code formatting rules anway, so that's really not something to worry about. 🙂

@ivanpauno if you think I am wrong in assuming that doing a release for Cyclone and updating rolling to that new release would be something that the ROS 2 community would appreciate, then please do say so!

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -685,11 +749,351 @@ static std::string get_type_name(const char * type_support_identifier, void * ty
   }
 }
 
+
+extern "C" dds_dynamic_type_descriptor_t get_dynamic_type_descriptor_prim(
```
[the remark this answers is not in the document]
Splinter1984: I agree. `extern "C"` useless here.

rmw_cyclonedds_cpp/src/serdata.cpp
Splinter1984: Accepted your suggestions.

rmw_cyclonedds_cpp/src/serdata.cpp
Splinter1984: Apply your suggestions.

rmw_cyclonedds_cpp/src/serdata.cpp
Splinter1984: We still need two instances of the function. one for adding primitive types, one for more complex types.

rmw_cyclonedds_cpp/src/serdata.cpp
Splinter1984: I agree to continue working after errors in these functions it is not correct way.

rmw_cyclonedds_cpp/src/serdata.cpp
Splinter1984: I agree and change all the same moments to `assert`.

rmw_cyclonedds_cpp/src/serdata.cpp
Splinter1984: Yes you are right this step is not needed here.

rmw_cyclonedds_cpp/src/serdata.cpp
Splinter1984: My mistake. Change to `ddsrt_free`.

rmw_cyclonedds_cpp/src/serdata.cpp
Splinter1984: This is where I made a big mistake. I didn't check `string_upper_bound`.

rmw_cyclonedds_cpp/src/serdata.cpp
```diff
@@ -24,7 +24,9 @@
 #include "Serialization.hpp"
 #include "TypeSupport2.hpp"
 #include "bytewise.hpp"
-#include "dds/ddsi/q_radmin.h"
+#include "dds/ddsi/ddsi_radmin.h"
```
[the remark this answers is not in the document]
Splinter1984: Yes, it would be great to leave compatibility with older versions.

PR #454, opened 2023-05-15
Add timestamp to rmw_publish tracepoint

cwecht:
Depends on https://github.com/ros2/ros2_tracing/pull/74
For some background: https://github.com/ros2/ros2_tracing/issues/73

Question: Has `dds_write_ts` + `dds_time()` any drawbacks if compared with using plain `dds_write`?

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1999,8 +1999,9 @@ extern "C" rmw_ret_t rmw_publish(
     return RMW_RET_INVALID_ARGUMENT);
   auto pub = static_cast<CddsPublisher *>(publisher->data);
   assert(pub);
-  TRACETOOLS_TRACEPOINT(rmw_publish, ros_message);
-  if (dds_write(pub->enth, ros_message) >= 0) {
+  const dds_time_t tstamp = dds_time();
+  TRACETOOLS_TRACEPOINT(rmw_publish, (const void *)publisher, ros_message, tstamp);
+  if (dds_write_ts(pub->enth, ros_message, tstamp) >= 0) {
     return RMW_RET_OK;
   } else {
     RMW_SET_ERROR_MSG("failed to publish data");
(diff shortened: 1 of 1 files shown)


eboasson (approved):
LGTM

(I do note that the `ros_message` argument moves from the 2nd to the 3rd place in the `TRACEPOINT`. I assume this is intentional.)

> Question: Has dds_write_ts + dds_time() any drawbacks if compared with using plain dds_write?

None whatsoever.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2070,11 +2071,15 @@ static rmw_ret_t publish_loaned_int(
...
-    TRACETOOLS_TRACEPOINT(rmw_publish, ros_message);
+
     auto d = new serdata_rmw(cdds_publisher->sertype, ddsi_serdata_kind::SDK_DATA);
     d->iox_chunk = ros_message;
     // since we write the loaned chunk here, set the data state to raw
     shm_set_data_state(d->iox_chunk, IOX_CHUNK_CONTAINS_RAW_DATA);
+    // the following wouldn't work since dds_writecdr resets the timestamp anyways.
+    // const dds_time_t tstamp = dds_time();
+    // d->timestamp.v = tstamp;
+    TRACETOOLS_TRACEPOINT(rmw_publish, (const void *)publisher, ros_message, 0);
```
cwecht, line 2082: This is a bit unfortunate. As of now `dds_writecdr` will always overwrite the timestamp.

Will the serdata object outlive the `dds_writecdr` function call? If so, we could maybe also take the timestamp from it after the fact.

cwecht: Ok, I had a closer look into cyclonedds and it looks like the serdata object does _not_ outlive `dds_writecdr` if everything is fine.
@christophebedard I'd suggest to leave it like this for the time beeing. The only way I could image that this could be fixed is add the respective API to cyclonedds.

cwecht:
For the record: the linter warning revealed in the builds of https://github.com/ros2/rmw_connextdds/pull/120 should be fixed now.

rmw_cyclonedds_cpp/src/rmw_node.cpp
cwecht: After having a closer look at cyclones API, it turns out that there is an APi which seems to fit this usecase quite well: `dds_forwardcdr`. @eboasson and/or @ivanpauno, it would be great if you could have a look at these changes (again).

eboasson: @cwecht I'm catching up with whatever happened during my break ... I'll review as request but wanted to quickly note that this weird behaviour of `dds_writecdr` is the reason I needed to fix something ... and added `dds_forwardcdr`.

eboasson (approved):
LGTM!

PR #464, opened 2023-07-27
Clear out errors once we have handled them.

clalancette:
In particular, in the type-hash code we can handle a failure to encode or parse the type hash from the user data.  Once we've handled it, and printed out an error about it, we should clear it because we are going to continue on and something else down the line might have an error.

This will avoid errors like:

This error state is being overwritten:

 <old message>

with this new error message:

 <new message>


emersonknapp (approved):
Thanks for the patch

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19253/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13798/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19984/)

PR #468, opened 2023-09-11
Minor revamp of the CMakeLists.txt.

clalancette:
The most important bit in here is to set
CMAKE_CXX_STANDARD_REQUIRED to ON, so it builds in C++17 mode with clang.  Besides that, cleanup a comment, and get rid of extra calls to target_link_libraries that aren't required (we can just combine that all into one call).

### DIFF
--- a/rmw_cyclonedds_cpp/CMakeLists.txt
+++ b/rmw_cyclonedds_cpp/CMakeLists.txt
@@ -16,9 +16,10 @@ cmake_minimum_required(VERSION 3.5)
 
 project(rmw_cyclonedds_cpp)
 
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19492/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14012/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20211/)

eboasson (approved):
LGTM!

PR #474, opened 2023-10-18
avoid using dds common public mutex directly

iuhilnehc-ynos:
It depends on https://github.com/ros2/rmw_dds_common/pull/73

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -5152,7 +5152,7 @@ extern "C" rmw_ret_t rmw_destroy_client(rmw_node_t * node, rmw_client_t * client
       info->client.sub->gid,
       node->name, node->namespace_))
   {
-    RMW_SAFE_FWRITE_TO_STDERR("failed to publish ParticipantEntitiesInfo when destroying client");
+    RMW_SET_ERROR_MSG("failed to publish ParticipantEntitiesInfo when destroying client");
   }
 
   rmw_fini_cs(&info->client);
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1757,26 +1757,22 @@ extern "C" rmw_node_t * rmw_create_node(
   RET_ALLOC_X(node->namespace_, return nullptr);
   memcpy(const_cast<char *>(node->namespace_), namespace_, strlen(namespace_) + 1);
 
-  {
-    // Though graph_cache methods are thread safe, both cache update and publishing have to also
-    // be atomic.
-    // If not, the following race condition is possible:
-    // node1-update-get-message / node2-update-get-message / node2-publish / node1-publish
-    // In that case, the last message published is not accurate.
-    auto common = &context->impl->common;
-    std::lock_guard<std::mutex> guard(common->node_update_mutex);
-    rmw_dds_common::msg::ParticipantEntitiesInfo participant_msg =
-      common->graph_cache.add_node(common->gid, name, namespace_);
-    if (RMW_RET_OK != rmw_publish(
-        common->pub,
-        static_cast<void *>(&participant_msg),
-        nullptr))
-    {
-      // If publishing the message failed, we don't have to publish an update
-      // after removing it from the graph cache */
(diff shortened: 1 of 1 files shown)


iuhilnehc-ynos:
repos: https://gist.githubusercontent.com/iuhilnehc-ynos/a6544c0ac62eba5ef9599facc17dff4b/raw/37f31f7e29ba1b13be0fcfe65b76be9bf0173663/ros2.repos

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19673/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14180/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20397/)

eboasson (approved):
LGTM given the changes in https://github.com/ros2/rmw_dds_common/pull/73, but I would have one more look at some details there (as I wrote in a review comment).

I'll mark this approved: if https://github.com/ros2/rmw_dds_common/pull/73 changes, then this'll have to change anyway and get reviewed again; but if the decision is that the other PR is good as it is (I'm only suggesting another look, after all) then this seems ok as it is, too.

eboasson (approved):
Looks much nicer this way, thanks!

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1757,26 +1766,16 @@ extern "C" rmw_node_t * rmw_create_node(
...
-      // If publishing the message failed, we don't have to publish an update
-      // after removing it from the graph cache */
-      static_cast<void>(common->graph_cache.remove_node(common->gid, name, namespace_));
-      return nullptr;
-    }
+  // Though graph_cache methods are thread safe, both cache update and publishing have to also
+  // be atomic.
+  // If not, the following race condition is possible:
+  // node1-update-get-message / node2-update-get-message / node2-publish / node1-publish
+  // In that case, the last message published is not accurate.
```
clalancette, line 1773: I think this comment doesn't belong here anymore, right?  That is, the comment is still correct, but the handling of that race is in `rmw_dds_common` now.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1798,19 +1797,14 @@ extern "C" rmw_ret_t rmw_destroy_node(rmw_node_t * node)
...
-    rmw_dds_common::msg::ParticipantEntitiesInfo participant_msg =
-      common->graph_cache.remove_node(common->gid, node->name, node->namespace_);
-    result_ret = rmw_publish(
-      common->pub, static_cast<void *>(&participant_msg), nullptr);
-  }
+  // Though graph_cache methods are thread safe, both cache update and publishing have to also
+  // be atomic.
+  // If not, the following race condition is possible:
+  // node1-update-get-message / node2-update-get-message / node2-publish / node1-publish
+  // In that case, the last message published is not accurate.
```
clalancette, line 1804: Same here.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -5097,7 +5070,8 @@ static void rmw_fini_cs(CddsCS * cs)
   dds_delete(cs->pub->enth);
 }
 
-static rmw_ret_t destroy_client(const rmw_node_t * node, rmw_client_t * client)
+static rmw_ret_t destroy_client(
+  const rmw_node_t * node, rmw_client_t * client, bool destroy_graph = true)
```
clalancette, line 5074: Can you please describe why we need the `destroy_graph` boolean here?  It isn't totally clear to me.

clalancette (commented):
I've left just a few comments, nothing major.  I like this refactoring a lot.

rmw_cyclonedds_cpp/src/rmw_node.cpp
iuhilnehc-ynos: After using `make_scope_exit` in the `rmw_create_client`, I removed the `destroy_client` and move its original implementation code into `rmw_destroy_client since `destroy_client` is only called once. There is no boolean flag.

fujitatomoya (approved):
lgtm

clalancette (approved):
Looks good to me with green CI.

PR #482, opened 2024-02-28
Remove a bunch of unnecessary macros.

clalancette:
The main goal here is to sidestep differences between uncrustify 0.72 and 0.78.  However, I realized that a lot of these macros were either duplicating functionality we already had, or were just as long as "inlining" the code would be.  Just get rid of these macros.

In the one place that we can't do this, just mark the macro as not to be indented by uncrustify.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20384/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14794/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21133/)

eboasson (approved):
LGTM!

PR #485, opened 2024-03-13
Add tracepoint for publish/subscribe serialized message

h-suzuki-isp:
Please, refer to https://github.com/ros2/rclcpp/pull/2448 .
I have added the following trace points.

- `__rmw_publish_serialized_message()`
  - tracepoint : rmw_publish
- `_take_serialized_message()`
  - tracepoint : rmw_take

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -3509,12 +3509,6 @@ static rmw_ret_t rmw_take_ser_int(
       if (message_info) {
         message_info_from_sample_info(info, message_info);
       }
-      TRACETOOLS_TRACEPOINT(
-        rmw_take,
-        static_cast<const void *>(subscription),
-        static_cast<const void *>(serialized_message),
-        (message_info ? message_info->source_timestamp : 0LL),
-        *taken);
       // taking a serialized msg from shared memory
 #ifdef DDS_HAS_SHM
       if (d->iox_chunk != nullptr) {
@@ -3530,6 +3524,12 @@ static rmw_ret_t rmw_take_ser_int(
           serialized_message->buffer_length = size;
           ddsi_serdata_unref(d);
           *taken = true;
+          TRACETOOLS_TRACEPOINT(
+            rmw_take,
+            static_cast<const void *>(subscription),
+            static_cast<const void *>(serialized_message),
+            (message_info ? message_info->source_timestamp : 0LL),
+            *taken);
           return RMW_RET_OK;
         } else if (iox_header->shm_data_state == IOX_CHUNK_CONTAINS_RAW_DATA) {
           if (rmw_serialize(d->iox_chunk, &sub->type_supports, serialized_message) != RMW_RET_OK) {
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3553,12 +3568,24 @@ static rmw_ret_t rmw_take_ser_int(
...
       }
     }
     ddsi_serdata_unref(d);
   }
   *taken = false;
+  TRACETOOLS_TRACEPOINT(
+    rmw_take,
+    static_cast<const void *>(subscription),
+    static_cast<const void *>(serialized_message),
+    (message_info ? message_info->source_timestamp : 0LL),
```
eboasson, line 3587: If nothing was taken, `message_info->source_timestamp` is meaningless.

Wouldn't:
```suggestion
    0LL,
```
make more sense?

That is, of course, under the assumption that you really want the tracepoint to fire when successfully taking nothing, rather than firing only when actually taking something.

h-suzuki-isp: Thank you for your review, I required.

christophebedard (approved):
This looks good to me. @eboasson or @mjcarroll do you want to take a(nother) look?

h-suzuki-isp:
@eboasson @mjcarroll (Cc: @christophebedard )
I'd like to get this PR to Jazzy on time.
I would appreciate it if you could review it.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -2040,7 +2044,7 @@ extern "C" rmw_ret_t rmw_publish_serialized_message(
   }
 #endif
 
-  const bool ok = (dds_writecdr(pub->enth, d) >= 0);
+  const bool ok = (dds_forwardcdr(pub->enth, d) >= 0);
```
mjcarroll, line 2047: Just verifying this: we are switching from `writecdr` to `forwardcdr` because we are setting the timestamp above?  Is there any potential fallout from this change?

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -3506,7 +3510,6 @@ static rmw_ret_t rmw_take_ser_int(
       if (message_info) {
         message_info_from_sample_info(info, message_info);
       }
-
```
mjcarroll, line 3509: Nit: unnecessary whitespace change

mjcarroll (changes requested):
Mostly good, one nit and I would like clarification on the impact of the API change.

rmw_cyclonedds_cpp/src/rmw_node.cpp
christophebedard: Apparently not, the same thing was done in #454.

rmw_cyclonedds_cpp/src/rmw_node.cpp
h-suzuki-isp: Fixed in https://github.com/ros2/rmw_cyclonedds/pull/485/commits/2b33c0bea1e4045d5f8c86ee2184b16f297c6b99.

christophebedard:
I think we can probably skip CI since the only new change is a new blank line.

mjcarroll (approved):

PR #490, opened 2024-05-06
Update serdata.cpp

k-m-irfan:
Fixed error with reading the data from /frontvideostream topic


clalancette:
Thanks for the contribution.

Foxy is now End-of-Life, so we won't be doing any additional updates to it.  If you are still encountering this problem with Humble, Iron, or Jazzy, please feel free to reopen targeting one of those branches.

PR #498, opened 2024-06-05
Make rmw_destroy_wait_set return RMW_RET_INVALID_ARGUMENT

christophebedard:
https://github.com/ros2/rmw/pull/275 changed the `rmw_destroy_wait_set()` API documentation to indicate that it returns `RMW_RET_INVALID_ARGUMENT` if `wait_set` is `NULL`. However, this wasn't the case in practice in the implementations of `rmw_destroy_wait_set()` or in the relevant test.

Change the implementation here to return `RMW_RET_INVALID_ARGUMENT`.

Requires https://github.com/ros2/rmw_implementation/pull/234


fujitatomoya (approved):

christophebedard:
CI is over in https://github.com/ros2/rmw_implementation/pull/234#issuecomment-2150237344

clalancette (approved):

PR #501, opened 2024-06-28
Support Cyclone DDS v0.10.x and master + ROS 2 type definitions in DDS

### LINKED ISSUE #525: Extremely big latency when enabling shared memory
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - From humble binaries
- Version or commit hash:
  - Humble last
  - rclcpp
- Relevant hardware info
```bash
memory         15GiB System memory
processor      11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz 16 cores
display        GA107M [GeForce RTX 3050 Mobile]
```

#### Summary

I’ve been testing the performance of CycloneDDS with shared memory (SHM) enabled to demonstrate the performance improvements it should provide. I already encountered latencies I was not expecting using just CycloneDDS + iceoryx standalone, which I have reported in this [issue](https://github.com/eclipse-cyclonedds/cyclonedds/issues/2155). 

I've encountered two further issues only happening in rmw_cyclonedds:

1. The performance gain is almost non-existent. The SHM transport seems as performant as the COPY one, while in my CycloneDDS standalone tests, the latency for SHARED was half of the one for COPY. 
2. There is a latency explosion when increasing the number of subscribers, going well over the 200ms. In my testing machine, this happens around 18 subs. Please see the next section for the exact details of the test. 

#### Steps to reproduce issue

I used the [APEX performance test package](https://gitlab.com/ApexAI/performance_test) and [this dockerfile](https://gitlab.com/ApexAI/performance_test/-/blob/2.3.1/dockerfiles/Dockerfile.CycloneDDS?ref_type=tags) they provide (you can follow the instructions to spin up the docker directly there). 

I defined the following test:

```yaml
---
experiments:
  -
    process_configuration: INTER_PROCESS
    execution_strategy: INTER_THREAD
    sample_transport: 
      - BY_COPY
      - SHARED_MEMORY
      - LOANED_SAMPLES
    msg: Array8m
    pubs: 1
    subs: 
     - 1
     - 2
     - 4
     - 6
     - 8
     - 10
     - 12
     - 14
     - 16
     - 18
     - 20
     - 22
     - 24
     - 26
     - 28
     - 30
     - 32
    rate: 30
    reliability: BEST_EFFORT
    durability: VOLATILE
    history: KEEP_LAST
    history_depth: 5
    max_runtime: 30
    ignore_seconds: 5
```

That you can run (after installing the performance package) with:

```bash
ros2 run performance_report runner --configs definition.yaml
```

This test measures the average latency between a publisher publishing an 8MB payload (around the size of a typical pointcloud or a Full HD image) at 30Hz and an increasing number of subscribers, for the [three different available transports](https://gitlab.com/ApexAI/performance_test/-/tree/master/performance_test#:~:text=are%20currently%20implemented%3A-,Eclipse%20Cyclone%20DDS,-Eclipse%20Cyclone%20DDS) available for Cyclone: UDP(copy), SHM, and SHM with loaned samples. The QoS are the ones specified in the YAML. 

#### Expected results

A quite low latency as expected from Iceoryx mostly independent of the number of subscribers. Alternatively, if [this](https://github.com/eclipse-cyclonedds/cyclonedds/issues/2155) is indeed an issue, at least the same results that were obtained there. **For reference**:

image

#### Actual results

A negligible performance gain for less than 10 subs that continues worsening until **exploding** around 24 subs.

image

I also ran the test in one of our robots and the explosion happens even sooner, so it seems to be related to CPU capabilities (as they have the same 16GB of RAM):

image

Both the reference plot and the results plots were run in the same machine on the same conditions, just one after the other, with the same [memory pool configuration](https://gitlab.com/ApexAI/performance_test/-/blob/4d91978a1aa28d7963836f6d96be351b27d36413/dockerfiles/roudi_config.toml). At first I thought it may be running out of RAM or starting swapping, but I can confirm that doesn't seem to be the case. 

Do you guys have any ideas why this may happen and why it only seems to happen in the RMW integration of Cyclone?

Many thanks in advance!

eboasson:
This PR takes the type definition/discovery work from #445, and adds the changes needed to match the work done on Cyclone DDS's `master` branch since that PR was made. The big change here is the refactoring that moved the Iceoryx support into a plugin.

All old configs and code not using any special features should work unchanged, but the RMW layer does its own serialization and needs to support loans, and these necessitates some changes.

It also means that Iceoryx is no longer needed as build dependency, but I think it is important that the RMW layer continues to work unchanged with 0.10.x, at least until there is a tag/branch for the next release of Cyclone that incorporates all those changes.

This is a draft PR because I haven't gotten around to checking all combinations of not-loan/loan, serialized/not-serialized, simple/complex types, etc. etc. etc. I do think it worthwhile to make it easy to find for everyone else who is curious and to reduce the risk of double work, in case someone else also decides to give it a try.

Fixes #525

### DIFF
--- a/rmw_cyclonedds_cpp/CMakeLists.txt
+++ b/rmw_cyclonedds_cpp/CMakeLists.txt
@@ -34,13 +34,16 @@ find_package(tracetools REQUIRED)
 
 find_package(CycloneDDS QUIET CONFIG)
 if(CycloneDDS_FOUND)
-  # Support for shared memory in RMW depends on support for shared memory being compiled in
-  # Cyclone DDS and iceoryx_binding_c being available
-  get_target_property(_cyclonedds_has_shm CycloneDDS::ddsc SHM_SUPPORT_IS_AVAILABLE)
-  if(_cyclonedds_has_shm)
-    find_package(iceoryx_binding_c REQUIRED)
-  else()
-    message(STATUS "Cyclone DDS is NOT compiled with support for shared memory")
+  if(CycloneDDS_VERSION VERSION_LESS 0.11)
+    # Support for shared memory in RMW depends on support for shared memory being compiled
+    # in Cyclone DDS for Cyclone version 0.10.x (and older) and iceoryx_binding_c being
+    # available
+    get_target_property(_cyclonedds_has_shm CycloneDDS::ddsc SHM_SUPPORT_IS_AVAILABLE)
+    if(_cyclonedds_has_shm)
+      find_package(iceoryx_binding_c REQUIRED)
+    else()
+      message(STATUS "Cyclone DDS is NOT compiled with support for shared memory")
+    endif()
   endif()
 else()
   message(WARNING "Could not find Eclipse Cyclone DDS - skipping '${PROJECT_NAME}'")
@@ -81,6 +84,9 @@ add_library(rmw_cyclonedds_cpp
   src/Serialization.cpp
   src/TypeSupport2.cpp
   src/TypeSupport.cpp)
+
+configure_file("src/cdds_version.hpp.in" "src/cdds_version.hpp")
+target_include_directories(rmw_cyclonedds_cpp PRIVATE ${CMAKE_BINARY_DIR}/src)
 
 target_link_libraries(rmw_cyclonedds_cpp PUBLIC
   rmw::rmw)
--- a/rmw_cyclonedds_cpp/package.xml
+++ b/rmw_cyclonedds_cpp/package.xml
@@ -15,6 +15,7 @@
   <buildtool_depend>ament_cmake_ros_core</buildtool_depend>
 
   <depend>cyclonedds</depend>
+  <!-- Iceoryx not a build dependency for Cyclone versions greater than 0.10 -->
   <depend>iceoryx_binding_c</depend>
   <depend>rcutils</depend>
   <depend>rcpputils</depend>
--- a/rmw_cyclonedds_cpp/src/cdds_version.hpp.in
+++ b/rmw_cyclonedds_cpp/src/cdds_version.hpp.in
@@ -1,0 +1,22 @@
+// Copyright 2024 ZettaScale Technology
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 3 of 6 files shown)


mcp-ats:
Hey @eboasson I would love to try this on my setup running Jazzy but I have some problems like the following due to some API changes.

I would like to try it since I read in other issues related to using multiple interfaces that the newer version of CycloneDDS might fix them.

See here https://github.com/eclipse-cyclonedds/cyclonedds/issues/2024#issuecomment-2444386167

Thanks!

`serdata.cpp:859:3: error: invalid conversion from ‘std::size_t (*)(const ddsi_sertype*, const void*)’ {aka ‘long unsigned int (*)(const ddsi_sertype*, const void*)’} to ‘ddsi_sertype_get_serialized_size_t’ {aka ‘int (*)(const ddsi_sertype*, ddsi_serdata_kind, const void*, long unsigned int*, short unsigned int*)’} [-fpermissive]
  859 |   sertype_get_serialized_size,
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      |   |
      |   std::size_t (*)(const ddsi_sertype*, const void*) {aka long unsigned int (*)(const ddsi_sertype*, const void*)}
serdata.cpp:860:3: error: invalid conversion from ‘bool (*)(const ddsi_sertype*, const void*, void*, std::size_t)’ {aka ‘bool (*)(const ddsi_sertype*, const void*, void*, long unsigned int)’} to ‘ddsi_sertype_serialize_into_t’ {aka ‘bool (*)(const ddsi_sertype*, ddsi_serdata_kind, const void*, void*, long unsigned int)’} [-fpermissive]
  860 |   sertype_serialize_into
      |   ^~~~~~~~~~~~~~~~~~~~~~
      |   |
      |   bool (*)(const ddsi_sertype*, const void*, void*, std::size_t) {aka bool (*)(const ddsi_sertype*, const void*, void*, long unsigned int)}
`

jpace121:
If I was interested in helping push this along would that be a thing? There seems to be a few interesting configuration parameters exposed in 0.10 that may help with an issue I’m currently debugging.

eboasson:
Closing: superseded by #550

PR #560, opened 2026-03-05
Add thread-local MessageTypeSupport cache to rmw_deserialize

mjcarroll:
This PR adds a symmetric thread-local cache for MessageTypeSupport objects in rmw_deserialize. This avoids the expensive per-call regex_replace and stringstream overhead in the MessageTypeSupport constructor.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -12,25 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <algorithm>
-#include <atomic>
 #include <cassert>
-#include <chrono>
 #include <cstring>
-#include <functional>
-#include <iomanip>
-#include <limits>
-#include <map>
-#include <memory>
 #include <mutex>
-#include <regex>
-#include <set>
-#include <string>
-#include <tuple>
 #include <unordered_map>
 #include <unordered_set>
-#include <utility>
+#include <algorithm>
+#include <chrono>
+#include <iomanip>
+#include <map>
+#include <set>
+#include <functional>
+#include <atomic>
+#include <memory>
 #include <vector>
+#include <string>
+#include <tuple>
+#include <utility>
+#include <regex>
+#include <limits>
 
 #include "rcutils/allocator.h"
 #include "rcutils/env.h"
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -109,142 +110,152 @@ using namespace std::literals::chrono_literals;
 #define ddsi_sertype_unref(x) ddsi_sertopic_unref(x)
 #endif
 
-/* Set to > 0 for printing warnings to stderr for each messages that was taken more than this many
-   ms after writing */
+/* Set to > 0 for printing warnings to stderr for each messages that was taken
+   more than this many ms after writing */
```
biodranik, line 114: Please use the same column count/width as in the original code for formatter.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -109,142 +110,152 @@ using namespace std::literals::chrono_literals;
...
-/* Set to != 0 for periodically printing requests that have been blocked for more than 1s */
+/* Set to != 0 for periodically printing requests that have been blocked for
+ * more than 1s */
 #define REPORT_BLOCKED_REQUESTS 0
 
-#define RET_ERR_X(msg, code) do {RMW_SET_ERROR_MSG(msg); code;} while (0)
-#define RET_NULL_X(var, code) do {if (!var) {RET_ERR_X(#var " is null", code);}} while (0)
-#define RET_ALLOC_X(var, code) do {if (!var) {RET_ERR_X("failed to allocate " #var, code);} \
-} while (0)
+#define RET_ERR_X(msg, code)                                                   \
```
biodranik, line 121: There is no need to reformat everything, please either set up/use the same clang-format style as in the original code.

biodranik (changes requested):
While PR description sounds great, it is impossible to review due to unnecessary code formatting changes.

Please make a commit with only necessary/related changes using the same formatting code style.

mjcarroll:
> While PR description sounds great, it is impossible to review due to unnecessary code formatting changes.
> 
> Please make a commit with only necessary/related changes using the same formatting code style.

Yeah, don't know what happened there, here is the scoped PR: https://github.com/ros2/rmw_cyclonedds/pull/561

PR #561, opened 2026-03-05
perf(rmw_cyclonedds_cpp): Add thread-local caching to rmw_serialize and rmw_deserialize

mjcarroll:
Cache typesupport in thread local storage for repeated serialization/deserialization.  Just a prototype at the moment to start some conversation here based on results from a microbenchmark: https://github.com/mjcarroll/RosBenchmarks

Replaces #560 
Related: https://github.com/ros2/rosidl_typesupport_fastrtps/pull/142
Related: https://github.com/ros2/rmw_fastrtps/pull/866

Ref Zulip conversation: [#ROS General > Lyrical default RMW](https://openrobotics.zulipchat.com/#narrow/channel/526027-ROS-General/topic/Lyrical.20default.20RMW/with/577209575)

Generated with the help of Claude Sonnet 4.6

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -1842,67 +1842,113 @@ extern "C" rmw_ret_t rmw_get_serialized_message_size(
   return RMW_RET_UNSUPPORTED;
 }
 
-extern "C" rmw_ret_t rmw_serialize(
-  const void * ros_message,
-  const rosidl_message_type_support_t * type_support,
-  rmw_serialized_message_t * serialized_message)
-{
-  try {
-    auto writer = rmw_cyclonedds_cpp::make_cdr_writer(
-      rmw_cyclonedds_cpp::make_message_value_type(type_support));
+extern "C" rmw_ret_t
+rmw_serialize(const void *ros_message,
+              const rosidl_message_type_support_t *type_support,
+              rmw_serialized_message_t *serialized_message) {
+  // Cache the CDR writer per type_support pointer to avoid rebuilding
+  // make_message_value_type and make_cdr_writer on every call.
+  thread_local const rosidl_message_type_support_t *cached_serialize_ts =
+      nullptr;
+  thread_local std::unique_ptr<rmw_cyclonedds_cpp::BaseCDRWriter> cached_writer;
 
-    auto size = writer->get_serialized_size(ros_message);
+  try {
+    if (type_support != cached_serialize_ts) {
+      cached_writer = rmw_cyclonedds_cpp::make_cdr_writer(
+          rmw_cyclonedds_cpp::make_message_value_type(type_support));
+      cached_serialize_ts = type_support;
+    }
+
+    auto size = cached_writer->get_serialized_size(ros_message);
     rmw_ret_t ret = rmw_serialized_message_resize(serialized_message, size);
     if (RMW_RET_OK != ret) {
       rmw_reset_error();
       RMW_SET_ERROR_MSG("rmw_serialize: failed to allocate space for message");
       return ret;
     }
-    writer->serialize(serialized_message->buffer, ros_message);
+    cached_writer->serialize(serialized_message->buffer, ros_message);
     serialized_message->buffer_length = size;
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1842,67 +1842,113 @@ extern "C" rmw_ret_t rmw_get_serialized_message_size(
...
+    cached_writer->serialize(serialized_message->buffer, ros_message);
     serialized_message->buffer_length = size;
     return RMW_RET_OK;
-  } catch (std::exception & e) {
-    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING("rmw_serialize: failed to serialize: %s", e.what());
+  } catch (std::exception &e) {
+    cached_serialize_ts = nullptr; // Invalidate cache on error
+    cached_writer.reset();
+    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
+        "rmw_serialize: failed to serialize: %s", e.what());
```
biodranik, line 1876: There are still irrelevant formatting changes, it would be great to avoid them.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1842,67 +1842,113 @@ extern "C" rmw_ret_t rmw_get_serialized_message_size(
...
-  const rosidl_message_type_support_t * type_support,
-  void * ros_message)
-{
+extern "C" rmw_ret_t
+rmw_deserialize(const rmw_serialized_message_t *serialized_message,
+                const rosidl_message_type_support_t *type_support,
+                void *ros_message) {
+  // Cache the resolved typesupport and MessageTypeSupport per type_support
+  // pointer, mirroring the rmw_serialize thread-local writer cache.
+  // MessageTypeSupport's constructor runs std::regex_replace + ostringstream
```
biodranik, line 1887: Does it worth to complicate the code with caching instead of fixing the root cause [in the constructor](https://github.com/ros2/rmw_cyclonedds/blob/rolling/rmw_cyclonedds_cpp/src/MessageTypeSupport_impl.hpp) by removing std::regex and std::ostringstream? E.g. something like this:

```cpp
template<typename MembersType>
MessageTypeSupport<MembersType>::MessageTypeSupport(const MembersType * members)
{
  assert(members);
  this->members_ = members;

  std::string message_namespace(this->members_->message_namespace_);
  std::string message_name(this->members_->message_name_);

  if (!message_namespace.empty()) {
    std::string::size_type pos = 0;
    while ((pos = message_namespace.find("__", pos)) != std::string::npos) {
      message_namespace.replace(pos, 2, "::");
      pos += 2;
    }
  }

  std::string name;
  name.reserve(
    message_namespace.size() + 2 + 5 + message_name.size() + 1);  // "::" + "dds_::" + "_"

  if (!message_namespace.empty()) {
    name += message_namespace;
    name += "::";
  }
  name += "dds_::";
  name += message_name;
  name += '_';

  this->setName(name.c_str());
}
```

mjcarroll: Perhaps. I'm mostly pushing this so that @jmachowinski can take a look at it as well.  This was vibe coded in response to some micro-benchmarks that came up in Zulip, so I wouldn't read too much into the actual implementation versus the general idea behind it.

mjcarroll: Also, the TLS caching looks really good in the mcirobenchmark because you are always publishing the same message repeatedly.  In the case that you alternated the message types that you are publishing, you would constantly get cache misses, so optimizing the constructor additionally makes sense.

MoffKalast: Ah so you did in fact manage to genuinely benchmaxx the optimization @mjcarroll, I'm impressed 😂

fujitatomoya (commented):
after https://github.com/ros2/rmw_cyclonedds/pull/562, this still reduces the frequency of construction entirely.
on a thread that repeatedly serializes the same type (which is extremely common in ROS nodes with dedicated publisher threads), we can skip construction altogether.

PR #562, opened 2026-03-06
improve MessageTypeSupport performance.

fujitatomoya:
## Description

The optimization follows [biodranik's suggestion](https://github.com/ros2/rmw_cyclonedds/pull/561/changes#r2890294543) from the PR review, addressing the performance issue at its source.

Removed expensive operations:
- ❌ std::regex_replace and std::regex("__")
- ❌ std::ostringstream string concatenation

Replaced with efficient plain string operations:
- ✅ Simple find/replace loop to replace "__" with "::"
- ✅ reserve() to pre-allocate exact memory needed
- ✅ Direct string concatenation with += operator

related to https://github.com/ros2/rmw_cyclonedds/pull/561

No,

Yes, Claude Sonnet 4.6

### DIFF
--- a/rmw_cyclonedds_cpp/src/MessageTypeSupport_impl.hpp
+++ b/rmw_cyclonedds_cpp/src/MessageTypeSupport_impl.hpp
@@ -17,8 +17,6 @@
 
 #include <cassert>
 #include <memory>
-#include <regex>
-#include <sstream>
 #include <string>
 
 #include "MessageTypeSupport.hpp"
@@ -33,16 +31,31 @@ MessageTypeSupport<MembersType>::MessageTypeSupport(const MembersType * members)
   assert(members);
   this->members_ = members;
 
-  std::ostringstream ss;
   std::string message_namespace(this->members_->message_namespace_);
   std::string message_name(this->members_->message_name_);
+
   if (!message_namespace.empty()) {
     // Find and replace C namespace separator with C++, in case this is using C typesupport
-    message_namespace = std::regex_replace(message_namespace, std::regex("__"), "::");
-    ss << message_namespace << "::";
+    std::string::size_type pos = 0;
+    while ((pos = message_namespace.find("__", pos)) != std::string::npos) {
+      message_namespace.replace(pos, 2, "::");
+      pos += 2;
+    }
   }
-  ss << "dds_::" << message_name << "_";
-  this->setName(ss.str().c_str());
+
+  std::string name;
+  name.reserve(
+    message_namespace.size() + 2 + 5 + message_name.size() + 1);  // "::" + "dds_::" + "_"
+
+  if (!message_namespace.empty()) {
+    name += message_namespace;
+    name += "::";
+  }
+  name += "dds_::";
+  name += message_name;
+  name += '_';
+
+  this->setName(name.c_str());
 }
 
 }  // namespace rmw_cyclonedds_cpp
--- a/rmw_cyclonedds_cpp/src/MessageTypeSupport_impl.hpp
+++ b/rmw_cyclonedds_cpp/src/MessageTypeSupport_impl.hpp
@@ -31,31 +31,31 @@ MessageTypeSupport<MembersType>::MessageTypeSupport(const MembersType * members)
   assert(members);
   this->members_ = members;
 
-  std::string message_namespace(this->members_->message_namespace_);
   std::string message_name(this->members_->message_name_);
 
-  if (!message_namespace.empty()) {
+  std::string name;
+  if (this->members_->message_namespace_ != nullptr &&
+    this->members_->message_namespace_[0] != '\0')
+  {
+    std::string message_namespace(this->members_->message_namespace_);
     // Find and replace C namespace separator with C++, in case this is using C typesupport
     std::string::size_type pos = 0;
     while ((pos = message_namespace.find("__", pos)) != std::string::npos) {
       message_namespace.replace(pos, 2, "::");
       pos += 2;
     }
-  }
-
-  std::string name;
-  name.reserve(
-    message_namespace.size() + 2 + 5 + message_name.size() + 1);  // "::" + "dds_::" + "_"
-
-  if (!message_namespace.empty()) {
+    name.reserve(
+      message_namespace.size() + 2 + 5 + message_name.size() + 1);  // "::" + "dds_::" + "_"
     name += message_namespace;
     name += "::";
+  } else {
+    name.reserve(5 + message_name.size() + 1);  // "dds_::" + "_"
   }
   name += "dds_::";
   name += message_name;
   name += '_';
 
-  this->setName(name.c_str());
+  this->setName(name);
 }
 
 }  // namespace rmw_cyclonedds_cpp


fujitatomoya:
@biodranik CCed

biodranik (approved):
Thanks! LGTM if it works.

ahcorde:
Pulls: ros2/rmw_cyclonedds#562
Gist: https://gist.githubusercontent.com/ahcorde/b79cf81cdcba82bbe8b64dfbbfddae0e/raw/aae245497cfc359b3a20bb870967a5a8913e2461/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18372
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27893/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20892/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7748/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26866/)

fujitatomoya:
@ahcorde is this good to go?

jmachowinski:
I would like to see a performance measurement first.

fujitatomoya:
@jmachowinski do you have benchmark test? or anybody else does?

jmachowinski:
Nope, and I see this is a problem, because we have no way of figuring out, if we improve stuff, or introduce regressions. @mjcarroll recently tried to improve things, and afterwards we figured out, the benchmark was pointless.
For this specific change, one could do a quick test in compiler explore, and check if std::regex is really that slow (I expect it to be, but you never know). 
This also looks like it will not improve the real performance of a system, as this constructor will only be called a few times during initialization of the serializers etc. And this is exactly my point, I think we are optimizing the wrong things here...

rmw_cyclonedds_cpp/src/MessageTypeSupport_impl.hpp
```diff
@@ -33,16 +31,31 @@ MessageTypeSupport<MembersType>::MessageTypeSupport(const MembersType * members)
...
+    name += message_namespace;
+    name += "::";
   }
-  ss << "dds_::" << message_name << "_";
-  this->setName(ss.str().c_str());
+  name += "dds_::";
+  name += message_name;
+  name += '_';
+
+  this->setName(name.c_str());
```
jmachowinski: setName accepts a std::string
```suggestion
  this->setName(name));
```

rmw_cyclonedds_cpp/src/MessageTypeSupport_impl.hpp
```diff
@@ -33,16 +31,31 @@ MessageTypeSupport<MembersType>::MessageTypeSupport(const MembersType * members)
...
+  }
+
+  std::string name;
+  name.reserve(
+    message_namespace.size() + 2 + 5 + message_name.size() + 1);  // "::" + "dds_::" + "_"
+
+  if (!message_namespace.empty()) {
+    name += message_namespace;
+    name += "::";
   }
```
jmachowinski: The code flow is a bit odd, why not declare name first, and do everything related to the namespace in one if, and keep the 'namespace string' scope local to the if ?

rmw_cyclonedds_cpp/src/MessageTypeSupport_impl.hpp
```diff
@@ -33,16 +31,31 @@ MessageTypeSupport<MembersType>::MessageTypeSupport(const MembersType * members)
...
-  std::ostringstream ss;
   std::string message_namespace(this->members_->message_namespace_);
   std::string message_name(this->members_->message_name_);
+
   if (!message_namespace.empty()) {
     // Find and replace C namespace separator with C++, in case this is using C typesupport
-    message_namespace = std::regex_replace(message_namespace, std::regex("__"), "::");
-    ss << message_namespace << "::";
+    std::string::size_type pos = 0;
+    while ((pos = message_namespace.find("__", pos)) != std::string::npos) {
```
Timple, line 43: Perhaps comment that this is like this for performance reasons. 
People might read this snippet and try to `clean as you go` and introduce a regex :)

biodranik: Any C++ developer should understand that:
- There's a huge overhead when using std::regex
- In-place/in-memory replacement is faster than any memory allocation + memory copying

rmw_cyclonedds_cpp/src/MessageTypeSupport_impl.hpp
biodranik: Agree, the order should be
```
if (!message_namespace.empty()) {
  std::string name;
  name.reserve(...)
  // ...
}
```

biodranik (commented):
> This also looks like it will not improve the real performance of a system, as this constructor will only be called a few times during initialization of the serializers etc. And this is exactly my point, I think we are optimizing the wrong things here...

Even if this is not a performance-critical place, avoiding _unnecessary_, _avoidable_ overhead everywhere in the code would be a good example to new contributors (and LLMs) to produce a better quality, faster code.

FOSS should not be slow and ugly, right?

mjcarroll:
> And this is exactly my point, I think we are optimizing the wrong things here...

I agree that this isn't going to have a huge impact, but getting this code cleaner won't hurt.  I wouldn't encourage people to go out of their way to hunt for these things, but in this case it was uncovered when looking for other optimizations, so it seems low cost to just fix now.

rmw_cyclonedds_cpp/src/MessageTypeSupport_impl.hpp
jmachowinski: For this one there are good arguments for both code versions. If it is really only one time, the performance does not matter, and I would argue that maintainability and simplicity out weight performance. If it is in a hot code path the proposed version is clearly the winner. Anyway I am fine with both versions.

jmachowinski:
Pulls: ros2/rmw_cyclonedds#562
Gist: https://gist.githubusercontent.com/jmachowinski/5ecaa4010cc60f753995728f036dca60/raw/aae245497cfc359b3a20bb870967a5a8913e2461/ros2.repos
BUILD args:  --cmake-args -DRMW_IMPLEMENTATION=rmw_cyclonedds_cpp
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18491
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28053/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21027/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7904/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26997/)

fujitatomoya:
we can see the improvement,

```console
tomoyafujita@~/DVT/02_TEST >uname -srm
Linux 6.17.0-14-generic x86_64

tomoyafujita@~/DVT/02_TEST >grep 'model name' /proc/cpuinfo | head -1
model name      : Intel(R) Core(TM) i7-14700KF

tomoyafujita@~/DVT/02_TEST >g++ --version | head -1
g++ (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0

tomoyafujita@~/DVT/02_TEST >./bench
Correctness check passed.

=== MessageTypeSupport getName() benchmark ===
Iterations: 1000000 x 7 types = 7000000 calls each

OLD (regex + ostringstream): 3221.22 ms
NEW (find/replace + +=):     368.914 ms

OLD per call: 460.175 ns
NEW per call: 52.702 ns

Speedup: 8.73164x

```

with the following simulated benchmark program.

```c++
// bench_message_type_support.cpp
// Simulates the old vs new MessageTypeSupport constructor
// from ros2/rmw_cyclonedds PR #562 (commit 8e0ed9d)
//
// Build & run:
//   g++ -O2 -o bench bench_message_type_support.cpp && time ./bench

#include <iostream>
#include <string>
#include <regex>
#include <sstream>
#include <chrono>
#include <vector>
#include <cstring>

// Prevent compiler from optimizing away the result
static volatile size_t sink = 0;

// ============================================================
// OLD: std::regex_replace + std::ostringstream
// (MessageTypeSupport_impl.hpp before PR #562)
// ============================================================
std::string getName_old(const char* message_namespace_, const char* message_name_) {
    std::ostringstream ss;
    std::string message_namespace(message_namespace_);
    std::string message_name(message_name_);
    if (!message_namespace.empty()) {
        // Find and replace C namespace separator with C++
        message_namespace = std::regex_replace(message_namespace, std::regex("__"), "::");
        ss << message_namespace << "::";
    }
    ss << "dds_::" << message_name << "_";
    return ss.str();
}

// ============================================================
// NEW: null check + find/replace loop + reserve + direct +=
// (MessageTypeSupport_impl.hpp after PR #562, commit 8e0ed9d)
// ============================================================
std::string getName_new(const char* message_namespace_, const char* message_name_) {
    std::string message_name(message_name_);

    std::string name;
    if (message_namespace_ != nullptr && message_namespace_[0] != '\0') {
        std::string message_namespace(message_namespace_);
        // Find and replace C namespace separator with C++
        std::string::size_type pos = 0;
        while ((pos = message_namespace.find("__", pos)) != std::string::npos) {
            message_namespace.replace(pos, 2, "::");
            pos += 2;
        }
        name.reserve(
            message_namespace.size() + 2 + 5 + message_name.size() + 1);
        name += message_namespace;
        name += "::";
    } else {
        name.reserve(5 + message_name.size() + 1);
    }
    name += "dds_::";
    name += message_name;
    name += '_';
    return name;
}

int main() {
    // Realistic ROS 2 type namespaces (C typesupport uses "__" separator)
    std::vector<std::pair<const char*, const char*>> test_cases = {
        {"sensor_msgs__msg",              "PointCloud2"},
        {"geometry_msgs__msg",            "PoseStamped"},
        {"nav_msgs__msg",                 "OccupancyGrid"},
        {"tf2_msgs__msg",                 "TFMessage"},
        {"",                              "String"},          // empty namespace
        {nullptr,                         "Header"},          // null namespace
        {"very__deep__nested__ns__msg",   "Custom"},
    };

    const int ITERATIONS = 1'000'000;

    // Verify both produce same output
    for (auto& [ns, name] : test_cases) {
        const char* safe_ns = ns ? ns : "";
        std::string old_result = getName_old(safe_ns, name);
        std::string new_result = getName_new(ns, name);
        if (old_result != new_result) {
            std::cerr << "MISMATCH for ns=\"" << safe_ns << "\" name=\"" << name << "\"\n"
                      << "  old: " << old_result << "\n"
                      << "  new: " << new_result << "\n";
            return 1;
        }
    }
    std::cout << "Correctness check passed.\n\n";

    // --- Benchmark OLD ---
    auto t0 = std::chrono::high_resolution_clock::now();
    for (int i = 0; i < ITERATIONS; i++) {
        for (auto& [ns, name] : test_cases) {
            const char* safe_ns = ns ? ns : "";
            sink += getName_old(safe_ns, name).size();
        }
    }
    auto t1 = std::chrono::high_resolution_clock::now();

    // --- Benchmark NEW ---
    auto t2 = std::chrono::high_resolution_clock::now();
    for (int i = 0; i < ITERATIONS; i++) {
        for (auto& [ns, name] : test_cases) {
            sink += getName_new(ns, name).size();
        }
    }
    auto t3 = std::chrono::high_resolution_clock::now();

    double old_ms = std::chrono::duration<double, std::milli>(t1 - t0).count();
    double new_ms = std::chrono::duration<double, std::milli>(t3 - t2).count();
    long total_calls = (long)ITERATIONS * test_cases.size();

    std::cout << "=== MessageTypeSupport getName() benchmark ===\n";
    std::cout << "Iterations: " << ITERATIONS << " x " << test_cases.size()
              << " types = " << total_calls << " calls each\n\n";

    std::cout << "OLD (regex + ostringstream): " << old_ms << " ms\n";
    std::cout << "NEW (find/replace + +=):     " << new_ms << " ms\n\n";

    std::cout << "OLD per call: " << (old_ms / total_calls) * 1e6 << " ns\n";
    std::cout << "NEW per call: " << (new_ms / total_calls) * 1e6 << " ns\n\n";

    std::cout << "Speedup: " << old_ms / new_ms << "x\n";

    return 0;
}
```

i guess this is worth to take??? @mjcarroll @jmachowinski @Timple @biodranik wdyt?

mjcarroll:
Almost 10x faster, why not :)

mjcarroll (approved):
LGTM with green CI

fujitatomoya:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28069/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21039/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7918/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27010/)

jmachowinski (approved):

jmachowinski:
@fujitatomoya The CI run did not use cyclone as default DDS, but I think we are good to merge anyway.

mjcarroll:
Windows unrelated, merging.

fujitatomoya:
@Mergifyio backport kilted jazzy humble

PR #564, opened 2026-03-17
Add CPU-based fallback support for `rosidl::Buffer` type

nvcyc:
## Description

This pull request adds `rosidl::Buffer` CPU-based fallback support to `rmw_cyclonedds_cpp`.
Since CycloneDDS uses introspection typesupport instead of FastRTPS typesupport that has full support of the buffer descriptor-based endpoint-aware serialization, it gracefully falls back (by converting) to CPU-based buffer/sequence when non-CPU backends are encountered.

This pull request consists of the following key changes:

- **`TypeSupport2`** (`TypeSupport2.hpp`, `TypeSupport2.cpp`): Routes `uint8[]` fields marked `is_rosidl_buffer_` to the new `ROSIDLC_BufferSpanSequenceValueType` (C introspection) and `CppBufferSpanSequenceValueType` (C++ introspection) that convert non-CPU buffers to CPU-based types. 
- **`TypeSupport_impl`** (`TypeSupport_impl.hpp`): Extends the `uint8` deserialization path to detect `is_rosidl_buffer_` fields and deserialize directly into `rosidl::Buffer<uint8_t>` (C++ introspection) or into the `rosidl::Buffer<uint8_t>*` held by the C sequence (C introspection).

This pull request does not change existing CycloneDDS behavior. It ensures that messages with `uint8[]` fields can be correctly serialized and deserialized by CycloneDDS using CPU-backed data. Non-CPU buffer backends will trigger a CPU fallback conversion maintaining functional correctness.

Yes. Claude (claude-4.6-opus) via Cursor was used to assist with creating an initial prototype version of the changes contained in this PR.

### Additional Information

This PR is part of the broader ROS 2 native buffer feature introduced in this [post](https://discourse.openrobotics.org/t/working-prototype-of-native-buffers-accelerated-memory-transport/52399).

### DIFF
--- a/rmw_cyclonedds_cpp/src/TypeSupport2.hpp
+++ b/rmw_cyclonedds_cpp/src/TypeSupport2.hpp
@@ -553,6 +553,7 @@ class ROSIDLC_BufferSpanSequenceValueType : public SpanSequenceValueType
     size_t size;
     size_t capacity;
     bool is_rosidl_buffer;
+    bool owns_rosidl_buffer;
   };
 
 public:
--- a/rmw_cyclonedds_cpp/src/TypeSupport2.hpp
+++ b/rmw_cyclonedds_cpp/src/TypeSupport2.hpp
@@ -29,6 +29,7 @@
 #include "rosidl_typesupport_introspection_c/identifier.h"
 #include "rosidl_typesupport_introspection_c/message_introspection.h"
 #include "rosidl_typesupport_introspection_c/service_introspection.h"
+#include "rosidl_buffer/buffer.hpp"
 #include "rosidl_typesupport_introspection_cpp/field_types.hpp"
 #include "rosidl_typesupport_introspection_cpp/identifier.hpp"
 #include "rosidl_typesupport_introspection_cpp/message_introspection.hpp"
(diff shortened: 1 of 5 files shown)


eboasson:
I have good hopes #550 will make it in soonish, and that PR cleans up the (de)serialization code and so conflicts with this PR. Would it be possible to hold off on this one (or base it on top of #550)?

rmw_cyclonedds_cpp/src/TypeSupport2.hpp
```diff
@@ -513,5 +514,74 @@ auto AnyValueType::apply(UnaryFunction f)
...
+  {
+    return reinterpret_cast<const rosidl::Buffer<uint8_t> *>(ptr)->size();
+  }
+  const void * sequence_contents(const void * ptr) const override
+  {
+    auto * buf = reinterpret_cast<const rosidl::Buffer<uint8_t> *>(ptr);
+    if (buf->size() == 0) {return nullptr;}
+    if (buf->get_backend_type() == "cpu") {return buf->data();}
+    cpu_cache_ = buf->to_vector();
+    return cpu_cache_.data();
```
mjcarroll: I would need to double check, but I think this is dicey because of some potentially shared state.

mjcarroll (approved):

mjcarroll (changes requested):
.

rmw_cyclonedds_cpp/src/TypeSupport2.hpp
hidmic: Indeed. IIUC concurrent readers is not a far fetched scenario, and CPU cache assignment is not synchronized.

nvcyc: Thanks for the insight.
I'm updating it to use a `thread_local` CPU cache.
https://github.com/ros2/rmw_cyclonedds/blob/7a15cd45fcf6903852e2677b05abd09bd5d46e5d/rmw_cyclonedds_cpp/src/TypeSupport2.hpp#L536-L539

Note that this non-CPU buffer path is reached only when users intentionally publish messages with non-CPU buffers under `rmw_cyclonedds`. The existing CPU path should be safe.

nvcyc:
> I have good hopes #550 will make it in soonish, and that PR cleans up the (de)serialization code and so conflicts with this PR. Would it be possible to hold off on this one (or base it on top of #550)?

@eboasson  It looks like the new PR to track is now becoming PR #575, right?

nvcyc:
I've created another PR #576 based on #575 with the same feature changes .
We'll switch to reviewing #576 once #575 is merged.

nvcyc:
Closed as no longer needed after #575 is merged.

PR #565, opened 2026-03-17
UNSUPPORTED is not error, so should not set the error string.

fujitatomoya:
## Description

This error messages actually fail some tests, because tests expect that no error strings are set.
e.g `rcl_action` from https://ci.ros2.org/job/ci_linux/28060/#showFailuresLink

Part of https://github.com/ros2/rmw_cyclonedds/pull/550

Not really. but if the application rely on the error string in rcutils, that would be empty now.

No

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -3239,7 +3239,6 @@ extern "C" rmw_ret_t rmw_subscription_set_content_filter(
   static_cast<void>(subscription);
   static_cast<void>(options);
 
-  RMW_SET_ERROR_MSG("rmw_subscription_set_content_filter: unimplemented");
   return RMW_RET_UNSUPPORTED;
 }
 
@@ -3252,7 +3251,6 @@ extern "C" rmw_ret_t rmw_subscription_get_content_filter(
   static_cast<void>(allocator);
   static_cast<void>(options);
 
-  RMW_SET_ERROR_MSG("rmw_subscription_get_content_filter: unimplemented");
   return RMW_RET_UNSUPPORTED;
 }
 


fujitatomoya:
Pulls: ros2/rmw_cyclonedds#565
Gist: https://gist.githubusercontent.com/fujitatomoya/d93442a355a1db89b82507daccec56f0/raw/1d99d2842eb07596003202165e9e2fb2c46c7b9f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18510
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28076/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21046/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7925/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27017/)

jmachowinski:
Does a user still get an error message from the layer above using this function ? If yes I am fine with the change, if not then the end user would have no way of figuring out that this is not supported.

fujitatomoya:
if user application expects this specific message, that is behavior change with this PR. but... the API returns NOT SUPPRTED obviously, so i believe that should be no problem.

jmachowinski:
rcl/subscrition.c relies on the string being set :
https://github.com/ros2/rcl/blob/c5fe2bf0a57502462d5717b0fd2e4072e6b55c63/rcl/src/rcl/subscription.c#L515

fujitatomoya:
@jmachowinski ah... 😶‍🌫️ let me think about this tomorrow... i need sleep 😪

jmachowinski (changes requested):
This is the wrong approach, marking this PR so it does not get merged by accident.

fujitatomoya:
this error set approach has been used in else where, if we change this behavior, we need to change all the places to unset the error message strings. and as a matter of fact, upper layer actually expects the error message string to be set if the API does not return OK. this PR actually breaks the user expectation, after all i closed this PR.

Issue #577, opened 2026-04-02
Cross-RMW service interoperability: ListParameters request from rmw_cyclonedds_cpp client can be misdeserialized and trigger OOM on rmw_fastrtps_cpp server


Yu1iWoowah:
### Generated by Generative AI

GPT5.4

### Operating System:

Ubuntu 22.04 in Docker (aarch64)

### ROS version or commit hash:

humble

### 'ros2 doctor --report' output

<details><summary>ros2 doctor --report</summary>

```console
<COPY OUTPUT HERE>
```
</details>

### Steps to reproduce issue

Environment used for reproduction:
- ROS 2 Humble
- requester RMW: `rmw_cyclonedds_cpp`
- server RMW: `rmw_fastrtps_cpp`
- service: `rcl_interfaces/srv/ListParameters`
- server implementation: `rclpy`
- Python: `3.10`
- Docker memory limit in my environment: `20 GiB`

Notes:
- Run both processes on the same machine / in the same network namespace.
- Use the default ROS domain, or otherwise make sure both terminals use the same `ROS_DOMAIN_ID`.

1. Prepare a Fast DDS (`rmw_fastrtps_cpp`) parameter service server.

Create `cross_rmw_param_server.py`:

```bash
cat > cross_rmw_param_server.py <<'PY'
#!/usr/bin/env python3
import time
import rclpy
from rclpy.node import Node

rclpy.init()
node = Node("cross_rmw_param_server", start_parameter_services=True)
node.declare_parameter("foo", "bar")
node.declare_parameter("bar", 42)

try:
    end = time.monotonic() + 60.0
    while time.monotonic() < end:
        rclpy.spin_once(node, timeout_sec=0.1)
finally:
    node.destroy_node()
    rclpy.shutdown()
PY
chmod +x cross_rmw_param_server.py
```

2. In terminal A, start the server under `rmw_fastrtps_cpp`:

```bash
source /opt/ros/humble/setup.bash
export RMW_IMPLEMENTATION=rmw_fastrtps_cpp
python3 cross_rmw_param_server.py
```

3. Prepare a CycloneDDS client that sends exactly one `ListParameters` request.

Create `cross_rmw_list_client.py`:

```bash
cat > cross_rmw_list_client.py <<'PY'
#!/usr/bin/env python3
import rclpy
from rclpy.node import Node
from rcl_interfaces.srv import ListParameters

rclpy.init()
node = Node("cross_rmw_list_client")
[...]

Issue #184, opened 2020-05-12
Service interoperability with fastrtps


Stapelzeiger:
## Bug report

**Required Info:**

- Operating System:
  - macOS 10.14
  - Ubuntu 18.04 (32bit ARM)
- Installation type:
  - binary
- Version or commit hash:
  - Eloquent patch 1
- DDS implementation:
  - rmw_cyclonedds_cpp
  - rmw_fastrtps_cpp
- Client library (if applicable):
  - rclcpp & rclpy

#### Steps to reproduce issue
machine A:
```
export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp
ros2 run examples_rclcpp_minimal_client client_main
```
machine B:
```
export RMW_IMPLEMENTATION=rmw_fastrtps_cpp
ros2 run examples_rclcpp_minimal_service service_main
```

#### Expected behavior
the example works as usual:
service printing: request: 41 + 1
client printing: Result of add_two_ints: for 41 + 1 = 42

#### Actual behavior
service prints: request: 3249716421018666984 + 1 (numbers change between runs)
client doesn't return

#### Additional information
- cyclonedds<->cyclonedds service calls work
- freertps<-> freertps service calls work
- I tested on macOS locally, Ubuntu (32bit ARM) locally, and macOS to Ubuntu over wifi, results are consistent
- publish/subscribe works without issues
- rclpy versions of the examples have the same behavior

Stapelzeiger:
The issue exists also in foxy (osrf/ros2:nightly docker image).

eboasson:
@Stapelzeiger this is basically expected: see #8. There was (or I overlooked it) no clear specification how requests should be represented on the wire so did "something", which turns out to be prepending a header to the request/response content. Cyclone likely simply uses a shorter header.

So the first step to find out what it is supposed to be. One possibility seems to be that it is `rmw_requestid_t`. If so, I think making it interoperate doesn't require much more work than replacing `cdds_request_header_t` by `rmw_request_id_t` everywhere and making some local adjustments. I suppose one might then also want to put a "real" GUID in to be able to correlate it with the information stored in the graph cache, and that's available in `get_entity_gid`. So it should be pretty easy.

Stapelzeiger:
I see, I wasn't aware that this wasn't standardized.

It would be great to have different RMW either interoperate correctly or not at all (drop the message and maybe log an error).
[...]

PR #578, opened 2026-04-06
Add opt-in standard service-wire runtime support for services

Yu1iWoowah:
## Description

This PR adds opt-in standard service-wire runtime support for services in `rmw_cyclonedds` on the Humble branch.

The new path is gated behind `RMW_CYCLONEDDS_SERVICE_WIRE_MODE=standard` and keeps the existing legacy service wire mode as the default.

In standard mode, `rmw_cyclonedds`:

- uses the plain request/response payload shape for services
- parses service-correlation metadata from RTPS submessage / inline QoS metadata
- supports standard request/response send/take paths for both clients and services

This PR is intentionally scoped to **runtime support only**. It updates:

- `rmw_cyclonedds_cpp/src/serdata.hpp`
- `rmw_cyclonedds_cpp/src/serdata.cpp`
- `rmw_cyclonedds_cpp/src/rmw_node.cpp`

It does **not** include the cross-RMW launch tests used during local validation. Those tests depend on companion changes outside this repository and are better handled separately from the minimum runtime change proposed here.

This PR is only the `rmw_cyclonedds` side of the change. End-to-end cross-RMW interoperability also depends on companion CycloneDDS and Fast DDS work.

With companion local patches, CycloneDDS and Fast DDS can exchange the service-correlation metadata needed for the standard request/reply path, but `rmw_cyclonedds` on Humble still assumes the legacy wrapped service payload format by default.

Related to #184 and #577.

Yes, but only for users who explicitly opt in with `RMW_CYCLONEDDS_SERVICE_WIRE_MODE=standard`. If the variable is unset, `rmw_cyclonedds` keeps the existing legacy service wire mode by default, so current behavior remains unchanged.

For that opt-in mode, this PR enables standard service-wire interoperability for request/reply traffic. With the companion CycloneDDS and Fast DDS changes used during local validation, the following runtime matrix passes:

- `Cyclone standard -> Fast DDS dual`
- `Fast DDS dual -> Cyclone standard`
- `Cyclone legacy -> Cyclone legacy`

The default legacy behavior remains unchanged.

Yes. OpenAI Codex (GPT-5-based coding agent) was used to help with code drafting, test orchestration, and PR text preparation.

### Additional Information

What changed:

- add standard-mode request/response handling in `rmw_node.cpp`
- parse sample sequence number and related sample identity metadata in `serdata.cpp`
- extend `serdata_rmw` state in `serdata.hpp`
- use `dds_writecdr()` for freshly created standard-mode responses

CycloneDDS compile-compat handling included in this PR:

- loan/data allocator header compatibility in `rmw_node.cpp`
- internal DDSI header-name compatibility (`q_*` vs `ddsi_*`) in `serdata.cpp`
- `ddsi_serdata_ops` tail handling compatible with both the older SHM layout and the companion getter-enabled layout used during this work
- preservation of the public `releases/0.10.x` `q_*` DDSI header path
- explicit avoidance of aliasing `EntityId_t` to `ddsi_entityid_t` in the `q_protocol.h` fallback path

Validation:

- local runtime validation was performed with companion local patches for:
  - CycloneDDS core standard `related_sample_identity` emission
  - Fast DDS 2.6.x non-builtin standard `0x0083` send/receive interoperability
- the bidirectional standard-mode cases were exercised with local out-of-tree helper-based client/server runtime tests
- confirmed runtime matrix:
  - `Cyclone standard -> Fast DDS dual`: pass
  - `Fast DDS dual -> Cyclone standard`: pass
  - `Cyclone legacy -> Cyclone legacy`: pass
- mixed-mode validation also showed:
  - `Cyclone legacy -> Fast DDS dual`: fail
  - `Fast DDS dual -> Cyclone legacy`: fail

Those mixed-mode failures are expected with the current scope and are the reason this PR stays limited to **standard service-wire runtime support**.

Client-side calibration note:

- the standard-mode response mapping calibrates from the first correlated response
- before calibration, the current implementation supports only a single in-flight request

Out of scope / follow-up:

This PR does **not** attempt to make `legacy <-> standard` mixed-mode service communication interoperable.

That problem is larger than metadata compatibility alone, because legacy and standard service communication also differ in payload framing. Supporting `legacy <-> standard` robustly would likely require follow-up work in `rmw_cyclonedds`, such as:

- explicit peer wire-mode advertisement
- per-peer wire-mode selection
- dual-accept handling for both legacy wrapped payloads and standard plain payloads


asymingt:
Can we first please confirm that this issue affects the `rolling` distribution?  If so, we'll need to create a patch against the `rolling` branch and then, after merging, back-port it to humble and other distributions. We do things this way to make sure that we don't have improvements to older distributions that don't make it into newer distributions.

PR #589, opened 2026-07-10
Fix memory leak when taking service responses addressed to other clients

pavlo-yashchuk:
## Description

rmw_take_response_request drains the reply reader in a loop, deserializing each sample into the same ros_data and skipping replies that belong to other clients of the same service (GUID mismatch). The message content of the skipped iteration was never finalized, so the next deserialization overwrote sequence/string pointers, leaking the payload. 

In our deployment, three services are polling the same ros image service; each service leaked ~13-30 mb/min. With these changes, ram usage is stable even after 5h

yes, was diagnosed and patch drafted with Claude code (Fable 5)

### Additional Information

The leak needs two or more processes holding clients for the same service - single-client tests never hit it, since the caller always frees the last deserialized sample

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -79,7 +79,9 @@
 
 #include "rmw_security_common/security.hpp"
 
+#include "rosidl_runtime_c/message_initialization.h"
 #include "rosidl_runtime_c/type_hash.h"
+#include "rosidl_runtime_cpp/message_initialization.hpp"
 
 #include "rosidl_typesupport_cpp/message_type_support.hpp"
 
@@ -4737,6 +4739,29 @@ static const std::string csid_to_string(const client_service_id_t & id)
   return os.str();
 }
 
+static void refini_message(
+  const rosidl_message_type_support_t * type_supports, void * message)
+{
+  const rosidl_message_type_support_t * ts;
+  if ((ts =
+    get_message_typesupport_handle(
+      type_supports, rosidl_typesupport_introspection_cpp::typesupport_identifier)) != nullptr)
+  {
+    auto members =
+      static_cast<const rosidl_typesupport_introspection_cpp::MessageMembers *>(ts->data);
+    members->fini_function(message);
+    members->init_function(message, rosidl_runtime_cpp::MessageInitialization::ALL);
+  } else if ((ts = // NOLINT
+    get_message_typesupport_handle(
+      type_supports, rosidl_typesupport_introspection_c__identifier)) != nullptr)
+  {
+    auto members =
+      static_cast<const rosidl_typesupport_introspection_c__MessageMembers *>(ts->data);
+    members->fini_function(message);
+    members->init_function(message, ROSIDL_RUNTIME_C_MSG_INIT_ALL);
+  }
+}
+
 static rmw_ret_t rmw_take_response_request(
   CddsCS * cs, rmw_service_info_t * request_header,
   void * ros_data, bool * taken, dds_time_t * source_timestamp,
@@ -4773,6 +4798,7 @@ static rmw_ret_t rmw_take_response_request(
         *taken = true;
         return RMW_RET_OK;
       }
+      refini_message(&cs->sub->type_supports, ros_data);
     }
   }
   *taken = false;
@@ -5138,6 +5164,7 @@ static rmw_ret_t rmw_init_cs(
 
     sub_type_hash = type_supports->request_typesupport->get_type_hash_func(
       type_supports->request_typesupport);
+    sub->type_supports = *type_supports->request_typesupport;
     pub_type_hash = type_supports->response_typesupport->get_type_hash_func(
       type_supports->response_typesupport);
     subtopic_name =
@@ -5166,6 +5193,7 @@ static rmw_ret_t rmw_init_cs(
       type_supports->request_typesupport);
     sub_type_hash = type_supports->response_typesupport->get_type_hash_func(
       type_supports->response_typesupport);
+    sub->type_supports = *type_supports->response_typesupport;
     pubtopic_name =
       make_fqtopic(ROS_SERVICE_REQUESTER_PREFIX, service_name, "Request", qos_policies);
     subtopic_name = make_fqtopic(ROS_SERVICE_RESPONSE_PREFIX, service_name, "Reply", qos_policies);
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -362,6 +362,9 @@ struct CddsSubscription : CddsEntity
   rmw_gid_t gid;
   dds_entity_t rdcondh;
   rosidl_message_type_support_t type_supports;
+  // set only for client/service readers (rmw_init_cs): fini's and re-inits a
+  // taken message so the take loop can skip foreign replies without leaking
+  std::function<void(void *)> refini_message;
 #if CDDS_VERSION == CDDS_VERSION_0_10
   dds_data_allocator_t data_allocator;
 #endif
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -4737,6 +4739,29 @@ static const std::string csid_to_string(const client_service_id_t & id)
   return os.str();
 }
 
+static void refini_message(
```
fujitatomoya: if neither introspection identifier matches, refini_message does nothing and the failure mode is the original leak. (this is not supposed to happen though) can we use get_typesupport instead here, which sets RMW_SET_ERROR_MSG with both underlying errors and returns nullptr? it looks like the helper should reuse it instead of reimplementing the lookup without failure handling.

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -5138,6 +5164,7 @@ static rmw_ret_t rmw_init_cs(
 
     sub_type_hash = type_supports->request_typesupport->get_type_hash_func(
       type_supports->request_typesupport);
+    sub->type_supports = *type_supports->request_typesupport;
```
fujitatomoya: performance concern.
resolve the introspection typesupport once in rmw_init_cs (with proper error return) and store the already-resolved handle? the take path then needs no lookups, no error handling, and the "impossible" branch becomes genuinely impossible. also removes the (minor) double-lookup cost per skipped sample?

fujitatomoya (commented):
i think this is real bug, that should be addressed. a couple of comments to be resolved.

would love to hear comment from @eboasson @jmachowinski

jmachowinski:
I don't think this is the correct fix.
This points to a memory leak either in our type support functions, or in the serialization code.
This patch just clears the sample on reuse in the service case, but if one would reuse a msg in a subscriber you should also get a memory leak.

jmachowinski:
@pavlo-yashchuk Can you post the srv definition you are using, as this would point us into the direction we need to look.
E.g. Arrays, or strings etc.

pavlo-yashchuk:
sure

```
# Request
bool raw                        # true = raw Image, false = CompressedImage (JPEG)
int32 target_height             # 0 = default resolution
int32 target_width              # 0 = default resolution
bool get_rgb                    # true = color, false = grayscale
bool add_detections             # true = overlay detection bboxes
bool undistort                  # true = undistort image
bool return_time                # true = include timestamp in response
bool require_real_frame         # true = return empty if no real frame available
int32[] amr_markers             # tracked AMR marker IDs
int32[] tracked_markers         # tracked marker IDs

---

# Response
bool success
sensor_msgs/CompressedImage compressed_image  # populated when raw=false
sensor_msgs/Image image                       # populated when raw=true
```

rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -4773,6 +4798,7 @@ static rmw_ret_t rmw_take_response_request(
         *taken = true;
         return RMW_RET_OK;
       }
+      refini_message(&cs->sub->type_supports, ros_data);
```
jmachowinski: @pavlo-yashchuk is the leak also fixed, if you move this line down before the *taken = false ?

pavlo-yashchuk: No, the leak is inside the loop, not at the end. Every dds_take deserializes into the same ros_data, which overwrites the pointers from the previous payload without freeing them. So with [foreign, foreign, matching] queued, both foreign items are already leaked before we can get to a post-loop cleanup.

Also, when we find a match, we return early, so that the cleanup wouldn't even run in that case. The last sample is fine either way - the caller owns ros_data and finalizes it after the take returns. That's why it only shows up with 2 or more samples in one call, not with a single foreign reply.

jmachowinski: > No, the leak is inside the loop, not at the end.

@pavlo-yashchuk Did you test that, or is that what claude is telling your ? 


> Every dds_take deserializes into the same ros_data, which overwrites the pointers from the previous payload without freeing them.

That is the strange part, as far as I can see everything in the deserialization resizes correctly, or uses assign. And these two function should free the memory correctly, if it is already allocated.

pavlo-yashchuk: Yep, sorry. Ive measured it:
 
  Setup: ROS Jazzy, `rmw_cyclonedds_cpp` 2.2.3 / CycloneDDS 0.10.5, synthetic 256 KB                                                                                                                                                                                                                       
  responses, N clients on one service, sampling the client process RSS. Identical                                                                                                                                                                                                                          
  server/executor across runs; only `LD_LIBRARY_PATH` differs between stock and PR.                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                           
  ### Testing the reuse theory                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                           
  | clients | calls  | RSS growth     |                                                                                                                                                                                                                                                                    
  |---------|--------|----------------|                                                                                                                                                                                                                                                                    
  | 1       | 20,579 | +4.7 MB (flat) |                                                                                                                                                                                                                                                                    
  | 3       | 32,900 | +8.6 GB        |                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                           
  Flat at 1 client, leaks only with 2+ — exactly when the reply reader starts                                                                                                                                                                                                                              
  skipping foreign replies.                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                           
  ### Stock vs. fix (3 clients, 25 s)                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                           
  | build                        | calls  | RSS growth        | per call |                                                                                                                                                                                                                                 
  |------------------------------|--------|-------------------|----------|                                                                                                                                                                                                                                 
  | stock 2.2.3                  | 45,564 | +11.7 GB (linear) | ~257 KB  |                                                                                                                                                                                                                                 
  | stock + `MALLOC_ARENA_MAX=2` | 25,672 | +8.1 GB (linear)  | ~316 KB  |                                                                                                                                                                                                                                 
  | this PR                      | 78,149 | +12 MB (flat)     | ~0       |                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                           
  The arena cap doesn't reduce the per-call leak (the lower total is just fewer                                                                                                                                                                                                                            
  calls in the window — normalized, it's ~one payload/call either way, if anything                                                                                                                                                                                                                         
  slightly higher). So it's a real heap leak, not glibc fragmentation. The fix is                                                                                                                                                                                                                          
  also faster, because stock spends its time thrashing memory.                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                           
  ### Where it leaks                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                           
  `memray --native` attributes the leaked allocations to:                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                           
  ```                                                                                                                                                                                                                                                                                                      
  rmw_take_response → rmw_take_response_request → dds_take                                                                                                                                                                                                                                                 
    → serdata deserialize → rosidl_runtime_c__octet__Sequence__init                                                                                                                                                                                                                                        
  ```                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                           
  i.e. the foreign reply is deserialized and then discarded — ~1 payload leaked per                                                                                                                                                                                                                        
  call.

pavlo-yashchuk: if needed, I can provide scripts on how it was tested

jmachowinski: You tested this in jazzy. There was a major rewrite since this version. Is the memory leak still present in rolling/ lyrical ?
As your trace shows c support I guess you are using Python ?

pavlo-yashchuk: I apologise, I missed the branch 
it is fixed at the rolling/lyrical

pavlo-yashchuk:
closed due to my oversight

PR #590, opened 2026-07-12
Warn when system net.core.rmem_max is too low for CycloneDDS

Functionhx:
CycloneDDS needs sufficient kernel receive buffer space for reliable operation. The recommended minimum is 8388608 per the rmw_cyclonedds README.

Check /proc/sys/net/core/rmem_max at init time and emit a warning if the value is below the threshold. Only warns -- never modifies kernel settings.

### DIFF
--- a/rmw_cyclonedds_cpp/src/rmw_node.cpp
+++ b/rmw_cyclonedds_cpp/src/rmw_node.cpp
@@ -14,7 +14,9 @@
 
 #include <cassert>
 #include <cstring>
+#ifdef __linux__
 #include <fstream>
+#endif
 #include <mutex>
 #include <unordered_map>
 #include <unordered_set>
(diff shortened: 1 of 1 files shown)


rmw_cyclonedds_cpp/src/rmw_node.cpp
```diff
@@ -1386,6 +1387,21 @@ rmw_context_impl_s::init(rmw_init_options_t * options, size_t domain_id)
     return RMW_RET_OK;
   }
 
+  {
+    std::ifstream rmem_max_file("/proc/sys/net/core/rmem_max");
```
mjcarroll, line 1394: This is not going to work on systems that don't have `/proc/sys`, which would be Windows at least, probably WSL, and unclear on Mac.

We either need to guard to only do this on supported systems or provide alternative checks

mjcarroll (changes requested):

rmw_cyclonedds_cpp/src/rmw_node.cpp
Functionhx: You're right. I'll guard this check with `#ifdef __linux__`.

Functionhx:
Done！🚀

mjcarroll (approved):

ahcorde:
Pulls: ros2/rmw_cyclonedds#590
Gist: https://gist.githubusercontent.com/ahcorde/174c3607bb6b2eef4aecb0ab1899633a/raw/cef832ac7773e15035670c7a322636d298e76dde/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19968
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29948/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22713/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9832/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28830/)

PR #603, opened 2026-08-28
Add CPU fallback support for rosidl::Buffer

nvcyc:
## Description

  Add CPU fallback support for rosidl buffer fields to `rmw_cyclonedds_cpp`.

  CycloneDDS uses rosidl introspection to serialize messages. Previously, serialization accessed buffer fields through their CPU-only sequence callbacks. Publishing a message containing a non-CPU-backed
  `rosidl::Buffer<uint8_t>` therefore failed instead of falling back to the standard DDS wire representation.

  This change:

  - Detects rosidl buffer fields through introspection metadata.
  - Supports both introspection C++ and introspection C message representations.
  - Keeps the existing direct serialization path for CPU-backed buffers.
  - Converts non-CPU buffers to a temporary CPU-owned `std::vector<uint8_t>` using `to_vector()` during serialization.
  - Does not mutate the message or replace its original backend while publishing.
  - Deserializes received buffer fields into owned CPU storage.
  - Replaces a non-CPU-backed destination buffer with CPU storage when deserializing into a reused message.
  - Reports serialization failure when backend-to-CPU conversion fails, instead of emitting a partial or malformed sample.
  - Adds the required dependency on `rosidl_buffer`.

  This implements CPU fallback only. It does not add backend metadata discovery or direct transport of backend-specific buffer representations.

  Yes.

  Applications using `rmw_cyclonedds_cpp` can now publish messages containing non-CPU-backed rosidl buffers. The buffer is converted to CPU memory for standard DDS serialization, and subscribers receive a
  CPU-backed buffer with the same contents.

  CPU-backed messages continue to use the direct serialization path without an additional fallback copy.

  Yes. OpenAI Codex (GPT-5) was used to draft the changes in this pull request.

  ### Additional Information

  The implementation relies on the existing rosidl buffer API and introspection metadata, including `rosidl::Buffer<uint8_t>` and the `is_rosidl_buffer` field marker.

### DIFF
--- a/rmw_cyclonedds_cpp/src/Serialization.cpp
+++ b/rmw_cyclonedds_cpp/src/Serialization.cpp
@@ -511,8 +511,15 @@ protected:
   {
     size_t count = value_type.sequence_size(src);
     serialize_u32(dst, count);
+    std::vector<uint8_t> cpu_storage;
+    const bool data_required = !dst.ignores_data();
+    const void * contents =
+      value_type.serialization_contents(src, cpu_storage, data_required);
+    if (data_required && contents == cpu_storage.data() && cpu_storage.size() != count) {
+      throw std::runtime_error("rosidl buffer size changed during CPU fallback conversion");
+    }
     serialize_many(
-      dst, value_type.sequence_contents(src), count, value_type.element_value_type(), what);
+      dst, contents, count, value_type.element_value_type(), what);
   }
 
   template<typename WriteCursor>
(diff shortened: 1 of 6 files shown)


rmw_cyclonedds_cpp/src/Serialization.cpp
```diff
@@ -511,8 +511,15 @@ class CDRWriter final : public BaseCDRWriter
   {
     size_t count = value_type.sequence_size(src);
     serialize_u32(dst, count);
+    std::vector<uint8_t> cpu_storage;
+    const bool data_required = !dst.ignores_data();
+    const void * contents =
```
eboasson, line 516: If `ignores_data()`, then `contents` can now become `null` even for non-empty buffers, that triggers undefined behaviour in `serialize_many`.

eboasson (commented):
LGTM overall, but I have one real comment.

